PR #1, opened 2021-07-27
implement --help

erh:


flag.go
```diff
@@ -21,10 +22,15 @@ func ParseFlags(args []string, into interface{}) error {
 	cmdLine.SetOutput(&buf)
 
 	if err := extractFlags(cmdLine, into); err != nil {
-		return err
+		return fmt.Errorf("error extracing flags: %w", err)
 	}
 	if err := cmdLine.Parse(args[1:]); err != nil {
-		return err
+		if errors.Is(err, flag.ErrHelp) {
+			cmdLine.SetOutput(os.Stderr)
```
edaniels, line 29: Based on how this is used, we want to have the output be in buf (see above). Can you keep PrintDefaults but remove this line and return an error similar to line 39?

erh: ah yes, even easier than that.

edaniels (approved):
LGTM!

PR #5, opened 2021-12-08
merged egoutil into viam utils

erh:


web/api.go
```diff
@@ -0,0 +1,89 @@
...
+type APIMiddleware struct {
+	Handler APIHandler
+}
+
+func handleAPIError(w http.ResponseWriter, err error, extra interface{}) bool {
+	if err == nil {
+		return false
+	}
+
+	log.Printf("api error: %s %s\n", err, extra)
```
edaniels, line 33: might as well use utils.Logger.Debugw

edaniels (approved):
LGTM. Would be nice to replace `log` with `utils.Logger` for output consistency

web/api.go
erh: done

PR #6, opened 2021-12-10
Add token callback for testing

njooma:
When front-end testing, we cannot/should not programmatically type in username/passwords or click through SSO flows. This PR adds a way for the frontend to pass tokens received by Auth0 to the server for verification and authentication. It is currently only enabled in test mode.


web/auth0.go
```diff
@@ -170,6 +178,99 @@ func (h *callbackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "No id_token field in oauth2 token.", http.StatusInternalServerError)
+		return
+	}
+
+	p, err := h.state.newAuthProvider(ctx)
+	if HandleError(w, err) {
+		return
+	}
+
+	idToken, err := p.Verifier(h.state.authOIConfig).Verify(ctx, rawIDToken)
```
edaniels, line 241: can we refactor the two methods to use this common code?

edaniels (changes requested):
Just a request to reuse most of the duplicated code here. Otherwise LGTM

edaniels (approved):

PR #16, opened 2021-12-27
Add AppImage detection and workaround

Otterverse:
The AppImage runtime does a lot of "magic" to hook execve() (and related) calls, and detects if the target is inside or outside of the appimage, and swaps out the libc and ld-linux interpreter. The problem is that Go doesn't use libc but instead codes syscalls directly in assembly. So it can't be hooked by the normal AppImage runtime tricks, and on systems where the appimage libc/ld-linux are in use, execing external binaries then fails.

This adds a short block that detects the APPIMAGE environment variable, and if found, basically runs processes through bash (included in the appimage) instead of directly, and thus the real execve() can be properly hooked.

If there's a better way to do this, I'm all ears, as this feels very hackish... but so does the rest of the AppImage "magic."

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	cloud.google.com/go/secretmanager v1.0.0
 	cloud.google.com/go/storage v1.10.0
 	github.com/Masterminds/sprig v2.22.0+incompatible
+	github.com/alessio/shellescape v1.4.1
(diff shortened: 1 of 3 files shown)


Otterverse:
Nevermind, moving this to be directly in RDK.

Issue #12, opened 2021-12-14
Use auth interceptors in WebRTC


edaniels:
Right now WebRTC assumes that authentication to the signaler implies authentication to the intended recipient. This is true in Viam's usage of it, but it should not be a universal assumption. Instead, once a WebRTC session is established, authentication should happen with credentials specific to the intended host. This means a new DialOption will be required to distinguish the two credentials. Supporting the case where they are both the same should still work and maybe there's a simple enough default to keep the current assumption.

PR #22, opened 2022-01-14
rpc: add ExternalAuthService delegated authentication service

### LINKED ISSUE #18: Direct gRPC connections to local addresses should verify/authenticate the host
Local addresses resolve to private IP address which in certain scenarios can have one connect to the wrong robot. This is a minor security issue at the time of writing and we should implement a way to have a host authenticate itself to a client. We may want to accomplish this with certificates issued by cloud.

This may also warrant a connection string scheme that allows an insecure setting to ignore this verification.

edaniels:
Closes #18 

This new `ExternalAuthService` provides the way for us to be able to do authentication and authorization for local clients in the cloud while having the client trust the cloud using PKI. The new method is `AuthenticateTo` which will the direct and webrtc dialers to use it with the external auth address options. This means an `Authenticate` call happens against an external address and then `AuthenticateTo` which will use the JWT response on all requests to the actual address. For viam's sake, AuthenticateTo is only used in direct gRPC connections on local robot UI but not yet used fully until https://github.com/viamrobotics/goutils/issues/12 is addressed.

Other than that, there's a lot of fixes going on here with external authentication that were discovered while working on https://github.com/viamrobotics/app/pull/58.

Other PRs for this task:
https://github.com/viamrobotics/app/pull/58
https://github.com/viamrobotics/rdk/pull/446


edaniels:
self review done

erh (approved):
I think I get this, but this really needs something in a rpc/README.md describing how auth works. This style is good, but so much more complicated.

edaniels:
yep, needs a doc. I'll write one later similar to the one in app's godoc.

rpc/server_options.go
```diff
@@ -143,3 +157,25 @@ func WithAuthHandler(forType CredentialsType, handler AuthHandler) ServerOption
...
 }
+
+// WithAuthenticateToHandler returns a ServerOption which adds an authentication
+// handler designed to allow the caller to authenticate itself to some other entity.
+// This is useful when externally authenticating as one entity for the purpose of
+// getting access to another entity. Only one handler can exist and the forType
+// parameter will be the type associated with the JWT made for the authenticated to entity.
+// This can technically be used internal to the same server to "assume" the identity of
+// another entity but is not intended for such usage.
+func WithAuthenticateToHandler(forType CredentialsType, handler AuthenticateToHandler) ServerOption {
```
edaniels, line 168: made it simpler for users to get AuthenticateTo behavior and not mess it up as easily :). Note this is almost like an AWS assume role now but not designed for it (yet?)

njooma (approved):
I understand it more since talking about it in person

PR #23, opened 2022-01-20
Updating name of buf module

njooma:
So we can publish this to the buf repository


buf.yaml
```diff
@@ -24,4 +24,3 @@ breaking:
     - FILE
 deps:
   - buf.build/googleapis/googleapis:c33c435046f2f4db5e8d6db52bd6c662f50f60d8
-
```
edaniels, line 27: new line

edaniels (approved):
LGTM

PR #34, opened 2022-02-15
Change workflow runner image for buf

njooma:
Per our conversation @Otterverse, this PR updates the workflow runner to use the canon image


Otterverse (approved):
I think this is one of those actions that has to be pushed to see if it works. So, LGTM and then we'll find out. (Maybe wait to merge until tomorrow so troubleshooting time is there if needed.)

edaniels (approved):
LGTM

PR #38, opened 2022-04-21
RSDK-319 - refactor to make AtomicStore more accessible

stuqdog:
Makes AtomicStore more easily accessible to allow for safer cloud caching of robot configs in RSDK-305.

Since this is an extremely trivial change, `make build lint` should be sufficient testing.


edaniels (approved):
LGTM!

cheukt (approved):

PR #40, opened 2022-05-10
Fixing workflow

npmenard:
Repo dispatch syntax is not what I expected therefore dispatch action are only applied to python-sdk and not rust-sdk

See [here](https://github.com/marketplace/actions/repository-dispatch) for the proper syntax


njooma (approved):
Ohh that is weird syntax

PR #41, opened 2022-07-14
RSDK-246, RSDK-243 and bump to Go 1.18

Otterverse:
The workflow is the main thing here. I believe these changes will still make it behave the same as before, but as you know with workflows, hard to tell until actual merge.


edaniels (approved):
LGTM. May want to test this as a non-viam org member github PR

PR #46, opened 2022-08-02
add ContextCloser

alexis-wei:
related to [RSDK-503] to add closer than we can compare with

[RSDK-503]: https://viam.atlassian.net/browse/RSDK-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/io.go
+++ b/io.go
@@ -5,15 +5,18 @@ import (
 	"io"
 )
 
+// type ContextCloser has a Close method with context.
+type ContextCloser interface {
+	Close(ctx context.Context) error
+}
+
 // TryClose attempts to close the target if it implements
 // the right interface.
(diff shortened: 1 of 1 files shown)


io.go
```diff
@@ -5,15 +5,18 @@ import (
 	"io"
 )
 
+// type ContextCloser has a Close method with context.
```
edaniels, line 8: just `ContextCloser is a Closer with a context argument`

edaniels (approved):

io.go
alexis-wei: okie changed

PR #49, opened 2022-08-23
Add helpers to create metric/span monitoring using OpenCensus

AdamMagaluk:


perf/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"go.opencensus.io/stats/view"
+	"go.uber.org/multierr"
+)
+
+// RegisterApplicationViews registers all the default views we may need for the application. gRPC, MongoDb, http, etc...
```
edaniels, line 11: [supernit] MongoDb -> MongoDB; http -> HTTP

perf/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+// the exported data to the console.
+//
+// The intent is help new users familiarize themselves with the
+// capabilities of opencensus.
+//
+// This should NOT be used for production workloads.
+type PrintExporter struct{}
+
+// ExportView logs the view data.
+func (e *PrintExporter) ExportView(vd *view.Data) {
```
edaniels, line 38: can you add somewhere in a readme or doc.go how to use all this stuff? either via GCP / a local server/viewer/printer?

perf/stackdriver.go
```diff
@@ -0,0 +1,177 @@
...
+// NewCloudExporter creates a new Stackdriver (Cloud Monitoring) OpenCensus exporter with all options setup.
+func NewCloudExporter(ctx context.Context, logger golog.Logger) (Exporter, error) {
+	sdOpts := stackdriver.Options{
+		Context: ctx,
+		OnError: func(err error) {
+			logger.Errorw("opencensus stackdriver error", "error", err)
+		},
+		// ReportingInterval sets the frequency of reporting metrics to stackdriver backend.
+		ReportingInterval: 60 * time.Second,
+		MetricPrefix:      path.Join("app.viam.com", "opencensus"),
```
edaniels, line 39: this is a public library so we can't use this value here. this should be usable by anyone with minimal work as a utility library on top of rpc.

edaniels (changes requested):

perf/metrics.go
AdamMagaluk: I added a few examples to exporters_test.go

perf/stackdriver.go
AdamMagaluk: Ah good point. There's now a options struct passed in with it and the logger/context.

edaniels (approved):

PR #50, opened 2022-09-07
[DATA-298] Support getting log lines from managed processes

tessavitabile:
There are 3 PRs necessary to reserve the port in the slam process:

https://github.com/viamrobotics/slam/pull/29: orbslam reserves and logs the port
https://github.com/viamrobotics/goutils/pull/50 (this PR): add support for extracting log lines from managed processes
https://github.com/viamrobotics/rdk/pull/1269: slam service reads port from orbslam logs

https://github.com/viamrobotics/rdk/pull/1269 must be merged after https://github.com/viamrobotics/slam/pull/29 and https://github.com/viamrobotics/goutils/pull/50.

### DIFF
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -35,11 +35,6 @@ type ManagedProcess interface {
 	// Stop signals and waits for the process to stop. An error is returned if
 	// there's any system level issue stopping the process.
 	Stop() error
-
-	// Returns the oldest buffered log line. Up to 1000 log lines are buffered, with the
-	// oldest evicted first. Only supports a single reader (if there are multiple callers
-	// of GetLogLine(), log lines will be partitioned among them).
-	GetLogLine(ctx context.Context) (string, error)
 }
 
 // NewManagedProcess returns a new, unstarted, from the given configuration.
@@ -55,7 +50,7 @@ func NewManagedProcess(config ProcessConfig, logger golog.Logger) ManagedProcess
 		managingCh: make(chan struct{}),
 		killCh:     make(chan struct{}),
 		logger:     logger,
-		logBuffer:  make(chan string, 1010),
+		logWriter:  config.LogWriter,
 	}
 }
 
@@ -76,7 +71,7 @@ type managedProcess struct {
 	lastWaitErr error
 
 	logger    golog.Logger
-	logBuffer chan string
+	logWriter *io.Writer
 }
 
 func (p *managedProcess) ID() string {
@@ -105,11 +100,17 @@ func (p *managedProcess) Start(ctx context.Context) error {
 		cmd := exec.CommandContext(ctx, p.name, p.args...)
 		cmd.Dir = p.cwd
 		var runErr error
-		if p.shouldLog {
+		if p.shouldLog || p.logWriter != nil {
 			out, err := cmd.CombinedOutput()
 			if len(out) > 0 {
-				p.logger.Debugw("process output", "name", p.name, "output", string(out))
-				p.logBuffer <- string(out)
+				if p.shouldLog {
+					p.logger.Debugw("process output", "name", p.name, "output", string(out))
+				}
+				if p.logWriter != nil {
+					if _, err := (*p.logWriter).Write(out); err != nil && err != io.ErrClosedPipe {
+						return errors.Wrapf(err, "error writing process output to log writer")
+					}
+				}
 			}
 			if err != nil {
 				runErr = err
@@ -130,7 +131,7 @@ func (p *managedProcess) Start(ctx context.Context) error {
 	cmd.Dir = p.cwd
 
 	var stdOut, stdErr io.ReadCloser
-	if p.shouldLog {
+	if p.shouldLog || p.logWriter != nil {
 		var err error
 		stdOut, err = cmd.StdoutPipe()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pexec/managed_process.go
```diff
@@ -35,6 +35,11 @@ type ManagedProcess interface {
 	// Stop signals and waits for the process to stop. An error is returned if
 	// there's any system level issue stopping the process.
 	Stop() error
+
+	// Returns the oldest buffered log line. Up to 1000 log lines are buffered, with the
```
tessavitabile, line 39: Should I make the buffer size configurable?

pexec/managed_process.go
```diff
@@ -325,3 +337,21 @@ func (p *managedProcess) Stop() error {
...
+		return "", errors.Wrapf(err, "error while getting log line from managed process")
+	}
+	select {
+	case line := <-p.logBuffer:
+		return line, nil
+	case <-ctx.Done():
+		if err := ctx.Err(); err != nil {
+			return "", errors.Wrapf(err, "error while getting log line from managed process")
+		}
+		return "", errors.New("context canceled while getting log line from managed process")
```
tessavitabile, line 355: I'm not sure it's possible to hit this statement.

pexec/managed_process.go
edaniels: what about having ProcessConfig have a io.Writer for StdOut + StdErr that you manage? that gives you more options. This current method feels a little forced into to support the SLAM use case.

edaniels (changes requested):
Want to see if we can do something other than GetLogLine.

pexec/managed_process.go
tessavitabile: I thought everything in ProcessConfig has to be serializable as json.

tessavitabile: I could add more arguments to NewManagedProcess() to pass in a io.Writer, but that seemed invasive.

edaniels: nope doesn't have to serialize necessarily. can mark those fields as `json:"-"`

pexec/managed_process.go
```diff
@@ -98,10 +100,17 @@ func (p *managedProcess) Start(ctx context.Context) error {
...
+		if p.shouldLog || p.logWriter != nil {
 			out, err := cmd.CombinedOutput()
 			if len(out) > 0 {
-				p.logger.Debugw("process output", "name", p.name, "output", string(out))
+				if p.shouldLog {
+					p.logger.Debugw("process output", "name", p.name, "output", string(out))
+				}
+				if p.logWriter != nil {
+					if _, err := (*p.logWriter).Write(out); err != nil && !errors.Is(err, io.ErrClosedPipe) {
+						return errors.Wrapf(err, "error writing process output to log writer")
```
edaniels, line 111: dont think this should error. just an error log

pexec/managed_process.go
```diff
@@ -187,7 +197,24 @@ func (p *managedProcess) manage(stdOut, stdErr io.ReadCloser) {
...
 					return
 				}
-				p.logger.Debugw("output", "name", name, "data", string(line))
+				if p.shouldLog {
+					p.logger.Debugw("output", "name", name, "data", string(line))
+				}
+				if p.logWriter != nil && !logWriterError {
+					_, err := (*p.logWriter).Write(line)
+					if err == nil {
+						_, err = (*p.logWriter).Write([]byte("\n"))
```
edaniels, line 206: add continue here so you can remove err != nil and dedent block

pexec/managed_process.go
```diff
@@ -187,7 +197,24 @@ func (p *managedProcess) manage(stdOut, stdErr io.ReadCloser) {
...
+					}
+					if err != nil {
+						if !errors.Is(err, io.ErrClosedPipe) {
+							p.logger.Errorw("error writing process output to log writer", "name", name, "error", err)
+						}
+						if !p.shouldLog {
+							return
```
edaniels, line 213: this returning doesn't seem right and like a bug

pexec/process.go
```diff
@@ -5,16 +5,21 @@
...
-	CWD     string   `json:"cwd"`
-	OneShot bool     `json:"one_shot"`
-	Log     bool     `json:"log"`
+	ID        string     `json:"id"`
+	Name      string     `json:"name"`
+	Args      []string   `json:"args"`
+	CWD       string     `json:"cwd"`
+	OneShot   bool       `json:"one_shot"`
+	Log       bool       `json:"log"`
+	LogWriter *io.Writer `json:"-"`
```
edaniels, line 22: you don't need a pointer

edaniels (changes requested):

pexec/managed_process.go
tessavitabile: I don't think I should add continue here, since I want to check if either write had an error.

pexec/process.go
tessavitabile: How else can I make this optional?

pexec/managed_process.go
edaniels: oh I see.

so a io.Writer writes should be a DEBUG and are very exceptional

pexec/managed_process.go
edaniels: that's fine. it's better than stopping the process. I think that needs to be an option if an error logging should quit. in practice failing to write to an io.Writer should never happen

tessavitabile: Returning here won't stop the managed process, just the goroutine responsible for logging. And I do expect to fail to write to the io.Writer for slam, since slam closes the reader/writer pair, once it retrieves the port.

edaniels: oh duh! sure, great idea then

pexec/process.go
edaniels: an interface is still able to be nil, so you check that

edaniels (approved):
LGTM :)

JohnN193 (approved):
LGTM!

PR #53, opened 2022-09-14
APP-617 add license prop to package.json

DTCurrie:
Updating/adding license values in the project's package.json files to match the license.


stevebriskin (approved):
version change is safe?

DTCurrie:
> version change is safe?

I used the same version we used in `PRIME`. The version change in `rpc/js/package.json` was for the `MIT` license it was using previously.

PR #56, opened 2022-09-17
RSDK-559 Propagate cancels through WebRTC

cheukt:
tested with a local robot and a remote using grpcurl and a cancel that triggers after a second. Also tested using the UI - cancelling an operation on the UI works.

Added a few tests as well

follows https://github.com/grpc/grpc/blob/3ec0e29bcf48e5774a1b124f0edf581b7ca40055/doc/PROTOCOL-HTTP2.md#errors by sending a RST_STREAM signal to the server on cancellations, and the server also sends an error back


stuqdog (approved):
lgtm

PR #63, opened 2022-10-04
[RSDK-731] Handle wrtc noisy client

Kschappacher:
wrtc client is currently really noisy and we want the client logs to be set to debug if no error is found


edaniels (approved):
LGTM assuming you tested it

PR #64, opened 2022-10-04
Support better error logging in the HTTP service.

AdamMagaluk:
I found cases today where we are panicing in the app but the logs go unnoticed because the logs are split between stackdriver lines.

This includes a panic capure to the template handler and api handler.

It includes basic error logging for cases where we return a 5xx to the user.

### DIFF
--- a/web/api.go
+++ b/web/api.go
@@ -26,6 +26,20 @@ type APIHandler interface {
 type APIMiddleware struct {
 	Handler APIHandler
 	Logger  golog.Logger
+
+	// Recover from panics with a proper error logs.
+	PanicCapture
+}
+
+// NewAPIMiddleware returns a configured APIMiddleware with a panic capture configured.
+func NewAPIMiddleware(h APIHandler, logger golog.Logger) *APIMiddleware {
+	return &APIMiddleware{
+		Handler: h,
+		Logger:  logger,
+		PanicCapture: PanicCapture{
+			Logger: logger,
+		},
+	}
 }
 
 func handleAPIError(w http.ResponseWriter, err error, logger golog.Logger, extra interface{}) bool {
@@ -45,6 +59,9 @@ func handleAPIError(w http.ResponseWriter, err error, logger golog.Logger, extra
 		temp := fmt.Sprintf("err not able to be converted to json (%s) (%s)", data, err)
 		w.WriteHeader(http.StatusInternalServerError)
 		_, err = w.Write([]byte(temp))
+
+		logger.Errorf("Internal API Issue: %s", err)
+
 		if err != nil {
 			// hack for linter
 			return true
@@ -52,11 +69,17 @@ func handleAPIError(w http.ResponseWriter, err error, logger golog.Logger, extra
 	} else {
 		w.Header().Set("Content-Type", "application/json")
 		var er ErrorResponse
+		statusCode := http.StatusInternalServerError
 		if errors.As(err, &er) {
-			w.WriteHeader(er.Status())
-		} else {
-			w.WriteHeader(http.StatusInternalServerError)
+			statusCode = er.Status()
 		}
+
+		if statusCode >= 500 {
+			logger.Errorf("Internal API Issue: %s", err)
+		}
+
+		w.WriteHeader(statusCode)
+
 		_, err = w.Write(js)
 		if err != nil {
 			// hack for linter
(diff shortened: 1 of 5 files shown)


web/api.go
```diff
@@ -45,18 +59,27 @@ func handleAPIError(w http.ResponseWriter, err error, logger golog.Logger, extra
 		temp := fmt.Sprintf("err not able to be converted to json (%s) (%s)", data, err)
 		w.WriteHeader(http.StatusInternalServerError)
 		_, err = w.Write([]byte(temp))
+
+		logger.Errorf("Internal API Issue: %s", err)
```
edaniels, line 63: can you use Errorw on both instead?

web/panic_capture.go
```diff
@@ -0,0 +1,30 @@
...
+	err := recover()
+	if err == nil {
+		return
+	}
+
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusInternalServerError)
+	_, err = w.Write([]byte("internal server error"))
+	if err != nil {
+		p.Logger.Warnf("failed to write to response: %s", err)
```
edaniels, line 26: same Levelw comments

web/panic_capture.go
```diff
@@ -0,0 +1,30 @@
...
+// Recover captures and prints the error if recover() has an error.
+func (p *PanicCapture) Recover(w http.ResponseWriter, r *http.Request) {
+	err := recover()
+	if err == nil {
+		return
+	}
+
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusInternalServerError)
+	_, err = w.Write([]byte("internal server error"))
```
edaniels, line 24: if recover has an error, we are losing it at this point? I assume you want to log it on the last line but now it's the .Write err instead

web/panic_capture_test.go
```diff
@@ -0,0 +1,31 @@
...
+		panic(err)
+	}
+
+	w := httptest.NewRecorder()
+	handlerWithError(errors.New("some error"), w)
+
+	test.That(t, w.Code, test.ShouldEqual, http.StatusInternalServerError)
+	test.That(t, w.Body.String(), test.ShouldEqual, "internal server error")
+	test.That(t, observedLogs.All(), test.ShouldHaveLength, 1)
+	test.That(t, observedLogs.All()[0].Message, test.ShouldContainSubstring, "Unhandled error: some error")
```
edaniels, line 30: did this pass given my comment?

web/panic_capture.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// PanicCapture allows recovery during a request handler from panics. It prints a
+// formatted log to the underlying logger.
+type PanicCapture struct {
+	Logger golog.Logger
+}
+
+// Recover captures and prints the error if recover() has an error.
+func (p *PanicCapture) Recover(w http.ResponseWriter, r *http.Request) {
```
edaniels, line 16: optional: our middleware can still be catching panics and we could repanick here unless you think it's right to just squash it here

edaniels (changes requested):

web/panic_capture.go
AdamMagaluk: Ah yeah that was a last minute add from lint :)

web/api.go
AdamMagaluk: Why do you prefer Errorw()? One downside i see a lot in our prod logs is that most errors only show in Stackdriver by default with the generic error like "Internal API Issue". This makes quickly assessing the actual error require an extra step of drilling into the error field.

With this the message includes both the generic error text and underlying error.

I'm also not opposed to `logger.Errorsw(fmt.Sprintf("Internal API Issue: %s", err.Error()), "error", err)` but feels a bit heavy handed.

edaniels: oh okay i assumed gcp would ultimately ingest structured logs

web/panic_capture.go
AdamMagaluk: It's an option. The http server panic recovering logic would kick in a fire off another log. I don't see much value in it.

web/api.go
edaniels: when i used splunk this just worked out of the box

web/panic_capture_test.go
AdamMagaluk: No, I tested before i fixed the lints. +1 to get presubmit checks here: https://viam.atlassian.net/browse/SRE-2

web/panic_capture.go
edaniels: okay cool. as is is fine then!

web/panic_capture_test.go
edaniels: ah, yes

edaniels (approved):
LGTM

PR #65, opened 2022-10-04
Remove development metrics logger

AdamMagaluk:
In dev mode we were printing the metrics recorded ever 5 seconds. This was a lot and not useful. Getting rid of it. Trace logging stays since that is our only local HTTP request(ish) logging in the app locally.


edaniels (approved):
great

PR #66, opened 2022-10-05
Support for protojson encoding in API Middleware and Template Middleware

AdamMagaluk:
This changes includes a custom json marshaler that wraps the normal json.Marshal and protojson.Marshal. It will determine if the passed in data is a proto.Message or slice of proto.Messages and if so use the protojson.Marshal instead of the json.Marshal. Otherwise it will fallback to json.Marshal()

It also includes a new template function called `protoJson` which uses the custom marshaler. We can use this now in our templates when we pass data to the window that should be serialized with the protjson.

### DIFF
--- a/web/protojson/marshaler.go
+++ b/web/protojson/marshaler.go
@@ -1,0 +1,153 @@
+// Package protojson provides helpers to marshal proto.Message to json
+package protojson
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"reflect"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+var (
+	protoMessageType = reflect.TypeOf((*proto.Message)(nil)).Elem()
+
+	defaultMarshaler = Marshaler{DefaultMarshalingOptions()}
+)
+
+// Marshaler used to marshal data to json repecting proto.Message types.
+type Marshaler struct {
+	Opts MarshalingOptions
+}
+
+// Marshal encodes the interface to json. It identifies if a proto.Message or an slice/array of
+// proto.Message are passed in and for those types uses the protobuf protojson marshaler. Otherwise
+// uses the standard json.Marshal().
+//
+// Note beware of its limitations:
+//   - Does not support proto.Message within a map.
+//   - Does not support proto.Message embedded within another struct.
+func (m *Marshaler) Marshal(data interface{}) ([]byte, error) {
+	protoMarshaler := m.Opts.JSONOptions
+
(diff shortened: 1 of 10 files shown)


web/protojson/options.go
```diff
@@ -0,0 +1,36 @@
+package protojson
+
+import "google.golang.org/protobuf/encoding/protojson"
+
+// MarshalingOptions has options for how web http middleware marshals data to the clients.
+type MarshalingOptions struct {
+	// Enable protojson encoding in API responses and any json encoding.
+	EnableProtoJSON bool
+
+	// UseProtoNames uses proto field name instead of lowerCamelCase name in JSON
```
edaniels, line 10: does it make sense to just embed or nest protojson.MarshalOptions here?

web/protojson/marshaler.go
```diff
@@ -0,0 +1,154 @@
...
+type Marshaler struct {
+	Opts MarshalingOptions
+}
+
+// Marshal encodes the interface to json. It identifies if a proto.Message or an slice/array of
+// proto.Message are passed in and for those types uses the protobuf protojson marshaler. Otherwise
+// uses the standard json.Marshal().
+//
+// Note beware of its limitations:
+//   - Does not support proto.Message within a map..
```
edaniels, line 30: .. -> ...?

web/protojson/marshaler.go
```diff
@@ -0,0 +1,154 @@
...
+//   - Does not support proto.Message embedded within another struct
+func MarshalToInterface(data interface{}) (interface{}, error) {
+	return defaultMarshaler.MarshalToInterface(data)
+}
+
+func isSliceOfProtoMessage(input interface{}) bool {
+	t := reflect.TypeOf(input)
+	if t.Kind() == reflect.Array || t.Kind() == reflect.Slice {
+		ok := t.Elem().Implements(protoMessageType)
```
edaniels, line 93: just return instead of assign

edaniels (approved):
nice.

web/protojson/options.go
AdamMagaluk: Yeah changed the structure a bit. A bit simpler now

web/protojson/options.go
```diff
@@ -0,0 +1,28 @@
...
+// JSONOptions is the underlying protojson marshal options.
+type JSONOptions = protojson.MarshalOptions
+
+// MarshalingOptions has options for how web http middleware marshals data to the clients.
+type MarshalingOptions struct {
+	// Enable protojson encoding in API responses and any json encoding.
+	EnableProtoJSON bool
+
+	// Options used for protojson marshaling
+	JSONOptions
```
edaniels, line 14: nice

PR #67, opened 2022-10-05
Move select protoutils functions from rdk to goutils

DTCurrie:
There are certain `protoutils` functions that lived in `rdk` that we want to use in `app` as well. This PR introduces those functions to `goutils` so they can be shared between applications.


AdamMagaluk (approved):
LGTM

PR #68, opened 2022-10-05
Fix Typescript issues for downstream repos

micheal-parks:
Since RDK is directly using TS files here, and since we've recently increased TS strictness on RDK, I needed to sync up TS rules here. What follows is just cleaning up issues resulting from the added rules.

Most common problems:
- type imports must be explicit
- `unknown` is more type safe than `any`
- Other misc invalid types that were fixed

### DIFF
--- a/rpc/js/package.json
+++ b/rpc/js/package.json
@@ -1,25 +1,27 @@
 {
   "name": "@viamrobotics/rpc",
-  "version": "0.1.22",
+  "version": "0.1.23",
   "license": "Apache-2.0",
   "dependencies": {
     "@improbable-eng/grpc-web": "^0.13.0",
-    "google-protobuf": "^3.14.0",
-    "@types/webrtc": "^0.0.31"
+    "google-protobuf": "^3.14.0"
   },
   "devDependencies": {
-    "set-value": ">=4.0.1",
     "@types/google-protobuf": "^3.7.4",
-    "ts-loader": "^8.0.14",
+    "set-value": ">=4.0.1",
+    "ts-loader": "^9.4.1",
     "ts-protoc-gen": "^0.14.0",
-    "typescript": "^4.1.3",
-    "webpack": "^4.43.0",
-    "webpack-cli": "^4.7.2"
+    "typescript": "^4.8.4",
+    "vite": "^3.1.4",
+    "webpack": "^5.74.0",
+    "webpack-cli": "^4.10.0"
   },
   "main": "dist/main.js",
   "types": "src/index.d.ts",
   "scripts": {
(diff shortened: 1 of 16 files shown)


rpc/js/package.json
```diff
@@ -1,25 +1,27 @@
 {
   "name": "@viamrobotics/rpc",
-  "version": "0.1.22",
+  "version": "0.1.23",
   "license": "Apache-2.0",
   "dependencies": {
     "@improbable-eng/grpc-web": "^0.13.0",
-    "google-protobuf": "^3.14.0",
-    "@types/webrtc": "^0.0.31"
```
micheal-parks, line 8: This should already be included in Typescript's DOM library. I believe this package was meant for bleeding edge stuff that TS refused to add due to it not being browser-standard yet.

rpc/js/src/peer.ts
```diff
@@ -32,8 +32,7 @@ export async function newPeerConnectionForClient(disableTrickle: boolean, rtcCon
 		ordered: true
 	});
 	negotiationChannel.binaryType = "arraybuffer";
-	
-	let makingOffer = false;
```
micheal-parks, line 36: This was unused.

rpc/js/src/dial.ts
```diff
@@ -352,12 +353,12 @@ export async function dialWebRTC(signalingAddress: string, host: string, opts?:
 	});
 
 	let clientEndResolve: () => void;
-	let clientEndReject: (reason?: any) => void;
+	let clientEndReject: (reason?: unknown) => void;
 	let clientEnd = new Promise<void>((resolve, reject) => {
 		clientEndResolve = resolve;
 		clientEndReject = reject;
 	});
-	client.onEnd((status: grpc.Code, statusMessage: string, trailers: grpc.Metadata) => {
```
micheal-parks, line 360: Unused.

rpc/js/src/BaseChannel.ts
```diff
@@ -1,13 +1,13 @@
...
+import type { ProtobufMessage } from "@improbable-eng/grpc-web/dist/typings/message";
 
 export class BaseChannel {
-	public readonly ready: Promise<any>;
+	public readonly ready: Promise<unknown>;
 
 	private readonly peerConn: RTCPeerConnection;
 	private readonly dataChannel: RTCDataChannel;
-	private pResolve: (value: any) => void;
+	private pResolve: ((value: unknown) => void) | undefined;
```
micheal-parks, line 8: pResolve is not initialized in the constructor, so must be marked as possibly undefined.

DTCurrie (approved):
✅

PR #72, opened 2022-10-14
RSDK-578 Add client interceptors

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-578

Required for https://github.com/viamrobotics/rdk/pull/1501

### DIFF
--- a/rpc/dial_options.go
+++ b/rpc/dial_options.go
@@ -3,6 +3,7 @@ package rpc
 import (
 	"crypto/tls"
 
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/stats"
 )
 
@@ -52,6 +53,10 @@ type dialOptions struct {
 
 	// stats monitoring on the connections.
 	statsHandler stats.Handler
+
+	// interceptors
+	unaryInterceptor  grpc.UnaryClientInterceptor
+	streamInterceptor grpc.StreamClientInterceptor
 }
 
 // DialMulticastDNSOptions dictate any special settings to apply while dialing via mDNS.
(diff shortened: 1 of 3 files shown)


rpc/dial_options.go
```diff
@@ -52,6 +53,10 @@ type dialOptions struct {
 
 	// stats monitoring on the connections.
 	statsHandler stats.Handler
+
+	// interceptors
+	unaryInterceptor  grpc.UnaryClientInterceptor
```
cheukt, line 58: can we make these lists so we can add more than one interceptor?

maximpertsov: it seems like we can do that by passing in multiple dial options with interceptors and then chaining those interceptors. this is the pattern we have for server options

cheukt (approved):
LGTM

edaniels (approved):
LGTM if you add tests :)

rpc/dialer_test.go
```diff
@@ -357,3 +358,146 @@ func TestWithDialStatsHandler(t *testing.T) {
...
+
+	test.That(
+		t,
+		interceptedMethods,
+		test.ShouldResemble,
+		[]string{
+			"/proto.rpc.webrtc.v1.SignalingService/OptionalWebRTCConfig",
+			"/proto.rpc.v1.AuthService/Authenticate",
+			"/proto.rpc.examples.echo.v1.EchoService/Echo",
+			"/proto.rpc.v1.AuthService/Authenticate",
```
maximpertsov, line 431: we probably should exempt auth and signaling method like we do for server-side interceptors

maximpertsov: @edaniels any thoughts on this?

PR #76, opened 2022-10-17
APP-822 - Add handler wrapping rc/cors module

AdamMagaluk:


edaniels (commented):
Can we just use https://github.com/rs/cors in app?

AdamMagaluk:
We can

edaniels (approved):

PR #77, opened 2022-10-17
Match our grpc-gateway json settings to our legacy api.

AdamMagaluk:
Ensure we use proto field names instead of json names.


edaniels (approved):
l g t m

PR #78, opened 2022-10-18
APP-823 custom claims in rpc AuthHandler

AdamMagaluk:
To support other use-cases where the JWT claims may contain other metadata about the entity or require validation outside the standard JWT claims we expose a way to get all claims in the auth handler.

### DIFF
--- a/rpc/context.go
+++ b/rpc/context.go
@@ -15,6 +15,7 @@ const (
 	ctxKeyPeerConnection
 	ctxKeyAuthMetadata
 	ctxKeyAuthEntity
+	ctxKeyAuthClaims // all jwt claims
 )
 
 // contextWithHost attaches a host name to the given context.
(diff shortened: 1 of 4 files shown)


rpc/context.go
```diff
@@ -73,6 +74,20 @@ func ContextAuthMetadata(ctx context.Context) map[string]string {
 	return authMD.(map[string]string)
 }
 
+// contextWithAllAuthClaims attaches authentication jwt claims to the given context.
+func contextWithAllAuthClaims(ctx context.Context, claims Claims) context.Context {
+	return context.WithValue(ctx, ctxKeyAuthClaims, claims)
+}
+
+// ContextAllAuthClaims returns authentication jwt claims.
+func ContextAllAuthClaims(ctx context.Context) Claims {
```
edaniels, line 83: [nit] I think ContextAuthClaims is cleaner

edaniels (approved):
lgtm!

rpc/context.go
AdamMagaluk: Works for me. Updated

PR #79, opened 2022-10-19
js: Support dial option with a client passed access token

AdamMagaluk:


rpc/js/src/dial.ts
```diff
@@ -425,3 +450,42 @@ function iceCandidateToProto(i: RTCIceCandidateInit): ICECandidate {
...
+
+		if (opts.externalAuthAddress) {
+			throw new Error("cannot set externalAuthAddress with accessToken");
+		}
+
+		if (opts.externalAuthToEntity) {
+			throw new Error("cannot set externalAuthToEntity with accessToken");
+		}
+
+		if (opts.webrtcOptions) {
```
edaniels, line 476: so i think the one change to make here is that you can set a signalingAccessToken as well

edaniels (changes requested):

rpc/js/src/dial.ts
AdamMagaluk: Im not sure i get the case for separate access token, at least for how things stand today.

If your using signaling/webrtc the access token is only relevant for the signaling server. The actual peer connection doesn't need header anyways. 

If you're direct connecting you're never dealing with a signaling server.

edaniels: it's for the external auth concept which is a part of the rpc utils as it pertains to this being its own open source thing that has more options than we expose in rdk.

the case is your signaling server (external) uses its own auth that is separate from your peer such that you'd send two separate sets of credentials and never comingle the two

AdamMagaluk: Updated PR to include that.

rpc/js/src/dial.ts
```diff
@@ -59,43 +90,51 @@ async function makeAuthenticatedTransportFactory(address: string, defaultFactory
...
-				onMessage: (message: AuthenticateResponse) => {
-					thisAccessToken = message.getAccessToken();
-				},
-				onEnd: (code: grpc.Code, msg: string | undefined, _trailers: grpc.Metadata) => {
-					if (code == grpc.Code.OK) {
-						pResolve(md);
-					} else {
-						pReject(msg);
+
+			if (opts?.webrtcOptions?.signalingAccessToken != "") {
```
edaniels, line 98: this is almost right but in makeAuthenticatedTransportFactory, we don't want to know about webrtcOptions, just access tokens. That's why you see field reassignment in dialWebrtc. As far as both direct dialing and webrtc signaling are concerned, the authentication process is the same. If you are dialing direct and have a token, we use that token instead of the creds, but still call `AuthenticateTo` in the event external addr is provided.

For signaling, the same logic applies, we will directly pass the access token but if the external address is set, we will pass the token to that external address and call `AuthenticateTo`.

Overall, that means this function should never check a `opts.webrtcOptions`

rpc/js/src/dial.ts
```diff
@@ -183,17 +224,19 @@ export async function dialWebRTC(signalingAddress: string, host: string, opts?:
...
+		if (!opts.accessToken) {
+			optsCopy.authEntity = opts?.webrtcOptions?.signalingAuthEntity
+			if (!optsCopy.authEntity) {
+				if (optsCopy.externalAuthAddress) {
+					optsCopy.authEntity = opts.externalAuthAddress?.replace(/^(.*:\/\/)/, '');
+				} else {
+					optsCopy.authEntity = signalingAddress.replace(/^(.*:\/\/)/, '');
+				}
 			}
+			optsCopy.credentials = opts?.webrtcOptions?.signalingCredentials
```
edaniels, line 236: this is where we'd pass through a signaling access token, if set, for the sake of dialing the signaler.

edaniels (changes requested):

rpc/js/src/dial.ts
AdamMagaluk: Updated to pass accessToken to the AuthenticateTo() request if the external entity and other option are set.

edaniels (approved):
LGTM!

PR #80, opened 2022-10-21
APP-70 Add JSON tags to UserInfo

DTCurrie:
Adds JSON tags to the `UserInfo` struct to match standard casing when used in `app`.

### DIFF
--- a/web/user.go
+++ b/web/user.go
@@ -9,8 +9,8 @@ import (
 
 // UserInfo basic info about a user from a session.
 type UserInfo struct {
-	LoggedIn   bool
-	Properties map[string]interface{}
+	LoggedIn   bool                   `json:"loggedIn"`
+	Properties map[string]interface{} `json:"properties"`
 }
 
 // GetEmail the email.


web/user.go
```diff
@@ -9,8 +9,8 @@ import (
 
 // UserInfo basic info about a user from a session.
 type UserInfo struct {
-	LoggedIn   bool
-	Properties map[string]interface{}
+	LoggedIn   bool                   `json:"loggedIn"`
```
AdamMagaluk, line 12: Why is this needed?

DTCurrie: Trying to keep our casing consistent in `app`. Currently we put a lot of info on the window, some don't use any tags so we get PascalCase, others do have a tag but use snake_case, and others already have tags for camelCase. Since this is specifically for JSON and the standard casing for JSON is camelCase, I am trying to standardize our usage to that.

It's also causing a bit of a problem with our typing in `app`. Here is the file where we maintain most of the types for API interactions: https://github.com/viamrobotics/app/blob/main/frontend/src/types.ts

The casing usage is really inconsistent and has caused a lot of ugliness to maintain.

AdamMagaluk: I see. Ok lgtm. I'm more concerned about this being sent to the frontend. It should be in our API and documented in proto. Im worried we're going to add details here we don't expect a client to see.

AdamMagaluk (approved):

PR #99, opened 2022-10-25
RSDK-150 Add client interceptors for webrtc connections

maximpertsov:
Required for https://viam.atlassian.net/browse/RSDK-150

* Adds client interceptors for webrtc connections
* Chains together all interceptors provided in dial options - right now we are only keeping the last client / stream interceptor

### DIFF
--- a/rpc/dial_options.go
+++ b/rpc/dial_options.go
@@ -3,6 +3,7 @@ package rpc
 import (
 	"crypto/tls"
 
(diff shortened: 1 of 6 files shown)


cheukt (commented):
can we also test that the chaining is going as expected? otherwise this looks great

cheukt (changes requested):
see comment about streaming

cheukt (approved):
LGTM

PR #100, opened 2022-11-02
Migrate jwks/aweb-oauth handler to goutils from app

AdamMagaluk:
Copies our web-oauth code from app.


rpc/oauth/web_oauth.go
```diff
@@ -0,0 +1,144 @@
...
+// system instead of using Authenticate() to create the access token.
+func (a *webOAuthHandler) Authenticate(ctx context.Context, entity, payload string) (map[string]string, error) {
+	return nil, status.Error(codes.Unimplemented, "not implemented with webauth")
+}
+
+// VerifyEntity verifies that this handler is allowed to authenticate the given entity.
+// The handler can optionally return opaque info about the entity that will be bound to the
+// context accessible via ContextAuthEntity.
+func (a *webOAuthHandler) VerifyEntity(ctx context.Context, entity string) (interface{}, error) {
+	return a.EntityVerifier(ctx, entity)
```
edaniels, line 96: if this is nil, should we error?

edaniels (approved):
LGTM but 1 q from port

rpc/oauth/web_oauth.go
AdamMagaluk: Yeah good call. Throwing an internal error and added tests.

PR #101, opened 2022-11-03
Changes to support Modular Resources (Includes RSDK-472 )

Otterverse:
Two changes needed to support modular resources. (Note, this PR should probably not be merged until the big mod resources PR in RDK is ready to merge, in case other changes are needed here.)

1. This adds support to make a direct connection attempt the primary for unix sockets (and an option to do so for other possible use cases in the future.)

2. It adds Segpgid to ManagedProcess configuration. This prevents background processes spawned from getting foreground signals. For example, without this, hitting CTRL+C on a running viam-server also directly sent that SIGINT to the module. Now it will only be seen by the parent viam-server process, which can then handle stopping the background process/module in a clean way.

3. It switches the pexec final SIGKILL to signal the entire progress group (set above) so that children/grandchildren aren't left running and orphaned.

### DIFF
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -94,10 +94,17 @@ func dial(
 	tryLocal bool,
 ) (ClientConn, bool, error) {
 	var isJustDomain bool
-	if strings.ContainsRune(address, ':') {
-		isJustDomain = false
+	if strings.HasPrefix(address, "unix://") {
+		dOpts.mdnsOptions.Disable = true
+		dOpts.webrtcOpts.Disable = true
+		dOpts.insecure = true
+		dOpts.disableDirect = false
 	} else {
-		isJustDomain = net.ParseIP(address) == nil
+		if strings.ContainsRune(address, ':') {
+			isJustDomain = false
+		} else {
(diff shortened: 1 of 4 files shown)


edaniels (approved):

rpc/dial.go
```diff
@@ -94,10 +94,17 @@ func dial(
 	tryLocal bool,
 ) (ClientConn, bool, error) {
 	var isJustDomain bool
-	if strings.ContainsRune(address, ':') {
-		isJustDomain = false
+	if strings.HasPrefix(address, "unix://") {
```
edaniels, line 97: Please add a test for this

rpc/dial_options.go
```diff
@@ -254,3 +254,12 @@ func WithStreamClientInterceptor(interceptor grpc.StreamClientInterceptor) DialO
 		}
 	})
 }
+
+// WithForceDirectGRPC forces direct dialing first.
+func WithForceDirectGRPC() DialOption {
```
edaniels, line 259: Test for this

edaniels (approved):
LGTM but needs tests int this repo before merge

PR #102, opened 2022-11-04
Add goleak ignore lines for net/http

Otterverse:
Found the leak I was discussing with @edaniels earlier today. In RDK, the artifact system opens a grpc connection to google during VerifyMainTest() but the cache.Close() can only attempt net/http.(*Transport).CloseIdleConnections(). This doesn't interrupt in-progress connections, or yet-to-complete dial attempts, which need a full second or so at times to establish before they can be closed and become idle (and get closed for real.) As such, very short tests can leak two (related) routines while the Transport still has new connections dialing.


edaniels:
Unfortunately those new lines are very general and show up in leaks that can be fixed. I'll also take a look and see what we can do

edaniels:
It's this https://github.com/googleapis/google-cloud-go/issues/5430. gonna try to workaround

edaniels:
haha @AdamMagaluk  found it here too https://github.com/viamrobotics/rdk/pull/1140. I knew it felt familiar

PR #109, opened 2022-11-18
Add WaitForAssertionWithSleep testing util

stevebriskin:
Some time-sensitive tests need to tweak the sleep value. The default 50ms value is too high.

### DIFF
--- a/testutils/wait.go
+++ b/testutils/wait.go
@@ -11,9 +11,15 @@ import (
 // That is, if waiting with respect time is critical, it's okay to use this.
 func WaitForAssertion(t *testing.T, assertion func(tb testing.TB)) {
 	t.Helper()
+	WaitForAssertionWithSleep(t, 50*time.Millisecond, 100, assertion)
+}
+
+// WaitForAssertionWithSleep is similar to WaitForAssertion but takes a wait
+// time and maximum number of attempts as parameters.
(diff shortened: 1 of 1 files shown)


testutils/wait.go
```diff
@@ -10,18 +10,24 @@ import (
 // test code be able to be signaled that the assertion is ready to be checked.
 // That is, if waiting with respect time is critical, it's okay to use this.
 func WaitForAssertion(t *testing.T, assertion func(tb testing.TB)) {
+	t.Helper()
+	WaitForAssertionWithSleep(t, 50*time.Millisecond, 100, assertion)
```
edaniels, line 14: Can you elaborate on how this helps with tests? Was it causing too much CPU usage? Otherwise a test ideally wouldn't rely on how often this checks something because if it does, then this acts as a crutch for flakiness which it's designed to avoid.

stevebriskin: It is specifically for this test, which checks for expected polling frequency -- https://github.com/viamrobotics/rdk/blob/main/components/input/gpio/gpio_test.go#L448-L510. The default 50ms sleep too high for this test.

Generally, since these functions are meant to be used as a last resort where timing is important I think the user needs to have more control over check intervals and timeout.

edaniels: Okay fair

edaniels (approved):

PR #110, opened 2022-11-19
rpc: generate/expose unique auth identifier to grpc handlers

edaniels:


rpc/auth.go
```diff
@@ -48,15 +48,18 @@ type Claims interface {
...
+	// ID returns the identity (typically "jti") associated with the claims.
+	ID() string
+
 	// Entity must return the "entity" making the request to the rpc system from the jwt claims
 	// presented. Return error if entity is missing. Should not preform any entity checks.
 	Entity() (string, error)
 
-	// GetCredentialsType returns the rpc CredentialsType based on the jwt claims.
-	GetCredentialsType() CredentialsType
+	// CredentialsType returns the rpc CredentialsType based on the jwt claims.
```
edaniels, line 58: some goism stuff

rpc/context.go
```diff
@@ -60,26 +60,14 @@ func ContextPeerConnection(ctx context.Context) (*webrtc.PeerConnection, bool) {
 	return pc.(*webrtc.PeerConnection), true
 }
 
-// contextWithAuthMetadata attaches authentication metadata to the given context.
-func contextWithAuthMetadata(ctx context.Context, authMD map[string]string) context.Context {
-	return context.WithValue(ctx, ctxKeyAuthMetadata, authMD)
-}
-
-// ContextAuthMetadata returns authentication metadata. It may be nil if the value was never set.
-func ContextAuthMetadata(ctx context.Context) map[string]string {
```
edaniels, line 69: refactored this into claims. also ultimately decided not to pass claims to handlers because it's harder to signal to the handlers if the claims are confidential or not

rpc/server_auth.go
```diff
@@ -210,14 +217,22 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (interface{}, error) {
...
-			return tlsAuthEntity, nil
+			// mTLS based authentication contexts do not really have a sense of a unique identifier
+			// when considering multiple clients using the certificate. We deem this okay but it does
+			// mean that if the identifier is used to bind to the concept of a unique session, it is
+			// not sufficient without another piece of information (like an address and port).
+			// Furthermore, if TLS certificate verification is disabled, this trust is lost.
+			// Our best chance at uniqueness with a compliant CA is to use the issuer DN (Distinguished Name)
+			// along with the serial number; compliancy hinges on issuing unique serial numbers and if this
+			// is an intermediate CA, their parent issuing unique DNs.
+			return verifiedCert.Issuer.String() + ":" + verifiedCert.SerialNumber.String(), tlsAuthEntity, nil
```
edaniels, line 231: my research tells me this is safe to use as a unique identifier. unfortunately the TLS session/unique ID is not safe and the security considerations are too complex for me to understand if they are safe in our cases, so I did not use it

rpc/server_auth.go
```diff
@@ -272,29 +287,31 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (interface{}, error) {
...
 	}
 
-	// Pass the raw claims to the Context.
-	ctx = contextWithAuthClaims(ctx, claims)
-
-	// Pass the auth metadata to the context.
-	if claims.GetAuthMetadata() != nil {
-		ctx = contextWithAuthMetadata(ctx, claims.GetAuthMetadata())
+	claimsID := claims.ID()
+	if claimsID == "" {
```
edaniels, line 305: darn was abut about to say this isn't a breaking change for rdk receiving tokens but realized it is for app receiving tokens. gonna make it opt in

edaniels: once this is out for a while, we can start checking it

rpc/wrtc_server_channel.go
```diff
@@ -119,6 +130,12 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
 		}
 		handlerCtx = contextWithPeerConnection(handlerCtx, ch.peerConn)
 
+		// TODO(GOUT-11): Handle auth; right now we assume
+		// successful auth to the signaler implies that auth should be allowed here, which is not 100%
+		// true.
+		handlerCtx = ContextWithAuthUniqueID(handlerCtx, ch.uid)
```
edaniels, line 136: this was missing for webrtc handlers expecting similar info that grpc ones get

rpc/oauth/web_oauth.go
```diff
@@ -128,14 +129,15 @@ func SignWebAuthAccessToken(key *rsa.PrivateKey, entity, aud, iss, keyID string)
 		},
 		Claims: rpc.JWTClaims{
 			RegisteredClaims: jwt.RegisteredClaims{
+				ID:       uuid.NewString(),
```
edaniels, line 132: @AdamMagaluk does auth0 provide the jti? If not, any suggestions on how to work around us wanting unique identification of JWTs?

rpc/auth.go
AdamMagaluk: 👍

rpc/context.go
```diff
@@ -103,11 +91,38 @@ func contextAuthEntity(ctx context.Context) (interface{}, error) {
...
 func MustContextAuthEntity(ctx context.Context) interface{} {
 	authEntity, err := contextAuthEntity(ctx)
 	if err != nil {
 		panic(err)
 	}
 	return authEntity
 }
+
+// ContextWithAuthUniqueID attaches a unique ID for an authenticated entity to the given context.
+func ContextWithAuthUniqueID(ctx context.Context, authUniqueID interface{}) context.Context {
```
AdamMagaluk, line 105: Why cannot this be a string? We assume it's a string on reads?

rpc/context.go
edaniels: They can be when in use with JOSE which we wouldn't be well aware of but I can make a note of the condition

rpc/context.go
edaniels: Because I like to copy and paste

rpc/server_auth.go
```diff
@@ -36,11 +37,19 @@ const (
...
-	CredentialsType CredentialsType   `json:"rpc_creds_type,omitempty"`
-	AuthMetadata    map[string]string `json:"rpc_auth_md,omitempty"`
+	AuthCredentialsType CredentialsType   `json:"rpc_creds_type,omitempty"`
+	AuthMetadata        map[string]string `json:"rpc_auth_md,omitempty"`
 }
 
-// Entity entity from the claims Audience.
+// UID returns the unique ID from the claims ID.
+func (c JWTClaims) UID() (string, error) {
+	if c.RegisteredClaims.ID == "" {
```
AdamMagaluk, line 46: This will be problematic for Auth0. They do not set jti and explicitly do not allow it from my understanding. They recommend against any jwt id and blacklisting and instead push for short expiration fro access tokens with refresh or silent login in the browser.

I haven't tried it but see: https://community.auth0.com/t/jti-field-doesnt-work/44364

That said the web auth handler can override this and use something else if needed.

rpc/oauth/web_oauth.go
AdamMagaluk: See my other comment.

AdamMagaluk:
Your PR description: "... we need to expose a unique identifier with an authentication context.".

Why does the "control" session need to be tied to the authorization system?

rpc/server_auth.go
```diff
@@ -141,11 +148,14 @@ func (ss *simpleServer) authUnaryInterceptor(
 	handler grpc.UnaryHandler,
 ) (interface{}, error) {
 	if !ss.exemptMethods[info.FullMethod] {
-		authEntity, err := ss.ensureAuthed(ctx)
+		authSubject, authEntity, err := ss.ensureAuthed(ctx)
 		if err != nil {
 			return nil, err
 		}
 		ctx = ContextWithAuthEntity(ctx, authEntity)
+		if authSubject != "" {
```
AdamMagaluk, line 156: Is there a case where this is empty?

rpc/server_auth.go
```diff
@@ -272,29 +293,32 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (interface{}, error) {
...
+		return "", nil, err
 	}
 
-	// Pass the raw claims to the Context.
-	ctx = contextWithAuthClaims(ctx, claims)
-
-	// Pass the auth metadata to the context.
-	if claims.GetAuthMetadata() != nil {
-		ctx = contextWithAuthMetadata(ctx, claims.GetAuthMetadata())
+	// TODO(RSDK-886): start requiring this; until then, log it for observation.
```
AdamMagaluk, line 310: // TODO(RSDK-886): start requiring this; until then, log it for observation.

You're returning an error here?

rpc/server_auth.go
AdamMagaluk: Did you look at Thumbprint as an option?

rpc/server_auth.go
AdamMagaluk: Resolved. This should be Subject as we talked about.

rpc/server_auth.go
edaniels: nope. removed

rpc/server_auth.go
edaniels: yeah removing. old

rpc/wrtc_server.go
```diff
@@ -167,8 +167,12 @@ func (srv *webrtcServer) handler(path string) (handlerFunc, bool) {
 
 // NewChannel binds the given data channel to be serviced as the server end of a gRPC
 // connection.
-func (srv *webrtcServer) NewChannel(peerConn *webrtc.PeerConnection, dataChannel *webrtc.DataChannel) *webrtcServerChannel {
-	serverCh := newWebRTCServerChannel(srv, peerConn, dataChannel, srv.logger)
+func (srv *webrtcServer) NewChannel(
+	peerConn *webrtc.PeerConnection,
+	dataChannel *webrtc.DataChannel,
+	forHosts []string,
```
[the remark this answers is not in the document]
edaniels: changed to authAudience, see `webrtcServerChannel` comment

rpc/server_auth.go
edaniels: I did but I opted for this as actual uniqueness over a thumbprint which is still less unique. Do you have a preference?

AdamMagaluk: I don't have a strong preference. My understanding was that thumbprint will at least always be there and outside of sha-1 collisions should be quite unique.

AdamMagaluk (approved):
LGTM

PR #116, opened 2022-12-13
Improve Managed Process Stop() signaling/waiting

Otterverse:
Half of these changes are in the modular resources PR already, but as I hit more issues, it felt cleaner to make this a new PR and make things configurable.

This PR makes both the stop signal and the timeout configurable, and stops in three steps. To start, the child process is signaled directly. After 1/3 of the timeout has passed, it then signals the entire process group (process and its children) in case the parent isn't managing its own child processes. Lastly, after the remaining 2/3 of the timeout interval, all processes in the process group are killed.

The one potentially "breaking" change is using SIGTERM instead of SIGINT by default. To be clear, I don't think this will actually break anything, and will actually improve/fix some uses.

Reasoning: Under linux and macos, SIGTERM is the correct signal for requesting a program to stop, whereas SIGINT is meant for keyboard interrupts (CTRL+C), and doesn't always indicate that a program should exit (for example, python triggers a traceback on SIGINT, vs ending normally on SIGTERM.) The "kill" command and process managers like systemd also signal with SIGTERM, so many programs only ever expect that.

### DIFF
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -38,13 +38,13 @@ type ManagedProcess interface {
 func NewManagedProcess(config ProcessConfig, logger golog.Logger) ManagedProcess {
 	logger = logger.Named(fmt.Sprintf("process.%s_%s", config.ID, config.Name))
 
-	if err := config.Validate("new managed process"); err != nil {
-		logger.Warn(err)
+	if config.StopSignal == 0 {
+		config.StopSignal = syscall.SIGTERM
 	}
-	// SMURF: Or better to do this then the log/warn above?
-	// if config.StopSignal == nil {
-	//	config.StopSignal = syscall.SIGTERM
-	//}
+
+	if config.StopTimeout == 0 {
+		config.StopTimeout = defaultStopTimeout
+	}
 
 	return &managedProcess{
 		id:               config.ID,
@@ -76,7 +76,7 @@ type managedProcess struct {
 	stopped          bool
 	managingCh       chan struct{}
 	killCh           chan struct{}
-	stopSig          os.Signal
+	stopSig          syscall.Signal
 	stopWaitInterval time.Duration
 	lastWaitErr      error
 
@@ -333,7 +333,7 @@ func (p *managedProcess) Stop() error {
 	select {
 	case <-timer.C:
 		p.logger.Infof("stopping entire process group %d with signal %s", p.cmd.Process.Pid, p.stopSig)
-		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig.(syscall.Signal)); err != nil && !errors.Is(err, os.ErrProcessDone) {
+		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig); err != nil && !errors.Is(err, os.ErrProcessDone) {
 			return errors.Wrapf(err, "error signaling process group %d with signal %s", p.cmd.Process.Pid, p.stopSig)
 		}
 	case <-p.managingCh:
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -19,8 +19,8 @@ import (
 
 var errAlreadyStopped = errors.New("already stopped")
 
-// waitInterrupt is how long to wait after interrupting to move onto killing.
-const waitInterrupt = 3 * time.Second
+// defaultStopTimeout is how long to wait in seconds (all stages) between first signaling and finally killing.
+const defaultStopTimeout = 10
 
 // A ManagedProcess controls the lifecycle of a single system process. Based on
 // its configuration, it will ensure the process is revived if it every unexpectedly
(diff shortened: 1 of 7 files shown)


pexec/managed_process.go
```diff
@@ -40,17 +40,46 @@ type ManagedProcess interface {
 // NewManagedProcess returns a new, unstarted, from the given configuration.
 func NewManagedProcess(config ProcessConfig, logger golog.Logger) ManagedProcess {
 	logger = logger.Named(fmt.Sprintf("process.%s_%s", config.ID, config.Name))
+
+	var stopSig syscall.Signal
```
edaniels, line 44: [nit] drop this into validate and add tests

pexec/managed_process.go
```diff
@@ -40,17 +40,46 @@ type ManagedProcess interface {
...
+		stopSig = syscall.SIGUSR1
+	case "USR2", "SIGUSR2", "user defined signal 2", "12":
+		stopSig = syscall.SIGUSR1
+	case "TERM", "SIGTERM", "terminated", "terminate", "15":
+		stopSig = syscall.SIGTERM
+	default:
+		stopSig = syscall.SIGTERM
+	}
+
+	if config.StopTimeout == 0 {
```
edaniels, line 66: drop into validate too

pexec/managed_process.go
```diff
@@ -19,8 +19,8 @@ import (
 
 var errAlreadyStopped = errors.New("already stopped")
 
-// waitInterrupt is how long to wait after interrupting to move onto killing.
-const waitInterrupt = 3 * time.Second
+// defaultStopTimeout is how long to wait in seconds (all stages) between first signaling and finally killing.
+const defaultStopTimeout = 10
```
edaniels, line 23: make this a duration still

pexec/process.go
```diff
@@ -13,13 +13,15 @@ import (
...
-	LogWriter io.Writer `json:"-"`
+	ID          string    `json:"id"`
+	Name        string    `json:"name"`
+	Args        []string  `json:"args"`
+	CWD         string    `json:"cwd"`
+	OneShot     bool      `json:"one_shot"`
+	Log         bool      `json:"log"`
+	LogWriter   io.Writer `json:"-"`
+	StopSignal  string    `json:"stop_signal"`
+	StopTimeout float64   `json:"stop_timeout_seconds"`
```
edaniels, line 24: better to convert to a duration

pexec/managed_process.go
```diff
@@ -308,18 +341,32 @@ func (p *managedProcess) Stop() error {
...
+	timer := time.NewTimer(p.stopWaitInterval)
 	defer timer.Stop()
 	select {
 	case <-timer.C:
-		if err := p.cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) {
+		p.logger.Infof("stopping entire process group %d with %s", p.cmd.Process.Pid, p.stopSig.String())
+		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig.(syscall.Signal)); err != nil && !errors.Is(err, os.ErrProcessDone) {
+			return errors.Wrap(err, "error interrupting process")
+		}
+	case <-p.managingCh:
```
edaniels, line 360: stop timer in this case

pexec/managed_process.go
```diff
@@ -308,18 +341,32 @@ func (p *managedProcess) Stop() error {
...
+	// Lastly, kill everything in the process group that remains after a longer wait
+	timer2 := time.NewTimer(p.stopWaitInterval * 2)
+	defer timer2.Stop()
+	select {
+	case <-timer2.C:
+		p.logger.Infof("killing entire process group %d", p.cmd.Process.Pid)
+		if err := syscall.Kill(-p.cmd.Process.Pid, syscall.SIGKILL); err != nil && !errors.Is(err, os.ErrProcessDone) {
 			return errors.Wrap(err, "error killing process")
 		}
 	case <-p.managingCh:
```
edaniels, line 372: stop timer here too

pexec/managed_process.go
```diff
@@ -308,18 +341,32 @@ func (p *managedProcess) Stop() error {
...
+	// In case the process didn't stop, or left behind any orphan children in its process group,
+	// we now send a signal to everything in the process group after a brief wait.
+	timer := time.NewTimer(p.stopWaitInterval)
 	defer timer.Stop()
 	select {
 	case <-timer.C:
-		if err := p.cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) {
+		p.logger.Infof("stopping entire process group %d with %s", p.cmd.Process.Pid, p.stopSig.String())
+		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig.(syscall.Signal)); err != nil && !errors.Is(err, os.ErrProcessDone) {
+			return errors.Wrap(err, "error interrupting process")
```
edaniels, line 358: if your log has group, pid, sig, and all that, make sure to include in both errors here in case OS does not

pexec/managed_process.go
```diff
@@ -308,18 +341,32 @@ func (p *managedProcess) Stop() error {
...
-	// If after a while the process still isn't stopping, let's kill it.
-	timer := time.NewTimer(waitInterrupt)
+	// In case the process didn't stop, or left behind any orphan children in its process group,
+	// we now send a signal to everything in the process group after a brief wait.
+	timer := time.NewTimer(p.stopWaitInterval)
 	defer timer.Stop()
 	select {
 	case <-timer.C:
-		if err := p.cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) {
+		p.logger.Infof("stopping entire process group %d with %s", p.cmd.Process.Pid, p.stopSig.String())
```
edaniels, line 356: dont need to use .String in conjunction with %s, just `p.stopSig`

edaniels (changes requested):
I know this is in draft and needs tests but left a few comments. Nice work!

pexec/managed_process.go
Otterverse: Do you mean MOVE this to the config.Validate() method in process.go? Or just ALSO add tests/checks over there? I don't think it can be moved, as I'm parsing into an entirely different type than the string value from the config.

pexec/managed_process.go
Otterverse: Same as previous comment, not sure how this can be moved into ProcessConfig.Validate() (if that was what you mean?) because that's a pointer receiver and thus changes the "real" object, causing the robot to think configs have changed and thus reconfigure in an infinite loop. At least as I understand things. Am I missing some pattern that does this correctly? (E.g. sets a default when validating, not just rejects with an error.)

pexec/process.go
Otterverse: It's converted to a duration later, but perhaps I'm missing something to let me do it earlier. But from my understanding...

There's no native unmarshaller for time.Duration. It results in the underlying int64 type, so can't use decimal figures (I wanted seconds to be a fraction), and we definitely don't want users to have to input nanoseconds into a config.

Is there something I'm missing that makes this easier?

edaniels: Yes you can make your own unmarshal function. See https://github.com/viamrobotics/rdk/blob/main/config/config.go#L490.

pexec/managed_process.go
```diff
@@ -40,17 +37,28 @@ type ManagedProcess interface {
...
 func NewManagedProcess(config ProcessConfig, logger golog.Logger) ManagedProcess {
 	logger = logger.Named(fmt.Sprintf("process.%s_%s", config.ID, config.Name))
+
+	if err := config.Validate("new managed process"); err != nil {
+		logger.Warn(err)
+	}
+	// SMURF: Or better to do this then the log/warn above?
+	// if config.StopSignal == nil {
+	//	config.StopSignal = syscall.SIGTERM
+	//}
```
Otterverse, line 47: Trying to make sure this doesn't break anything in direct use in RDK (where a literal constructor might result in a nil signal) Please let me know what you think is the best way/place to handle that.

Otterverse:
Ugh, this breaks the config diffing stuff in RDK when I just tested... hold off on review.

Otterverse:
Okay, NOW it's passing in RDK as well, and therefore ready for review I believe.
Also, opened an issue to add the protobuf support in RDK afterwards. https://viam.atlassian.net/browse/RSDK-1596

edaniels (approved):

PR #118, opened 2022-12-21
Dial support for external auth with static credentials

AdamMagaluk:
- Support external auth with a static credential.
- Smart enough to figure out that when an external auth / signaler is used vs a local signaler when on mdns.

### DIFF
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -124,7 +124,7 @@ func dial(
 			// This path is also called by an mdns direct connection and ignores that case.
 			// This will skip all Authenticate/AuthenticateTo calls for the signaler.
 			if !dOpts.usingMDNS && dOpts.authMaterial == "" && dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial != "" {
-				logger.Debugw("using singaling's external auth as auth material")
+				logger.Debug("using signaling's external auth as auth material")
 				dOpts.authMaterial = dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial
 				dOpts.creds = Credentials{}
 			}
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -114,28 +114,20 @@ func dial(
 				// try WebRTC at same address
 				signalingAddress = address
 			}
-
-			var target string
-			var port uint16
-			if strings.Contains(signalingAddress, ":") {
-				host, portStr, err := net.SplitHostPort(signalingAddress)
-				if err != nil {
-					return nil, false, err
-				}
-				if strings.Contains(host, ":") {
-					host = fmt.Sprintf("[%s]", host)
-				}
-				target = host
-				portParsed, err := strconv.ParseUint(portStr, 10, 16)
-				if err != nil {
-					return nil, false, err
-				}
-				port = uint16(portParsed)
-			} else {
-				target = signalingAddress
-				port = 443
+			target, port, err := getWebRTCTargetFromAddressWithDefaults(signalingAddress)
+			if err != nil {
+				return nil, false, err
 			}
 			fixupWebRTCOptions(dOpts, target, port)
+
+			// When connecting to an external signaler for WebRTC, we assume we can use the external auth's material.
+			// This path is also called by an mdns direct connection and ignores that case.
+			// This will skip all Authenticate/AuthenticateTo calls for the signaler.
+			if !dOpts.usingMDNS && dOpts.authMaterial == "" && dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial != "" {
+				logger.Debugw("using singaling's external auth as auth material")
+				dOpts.authMaterial = dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial
+				dOpts.creds = Credentials{}
+			}
 		}
 
 		if dOpts.debug {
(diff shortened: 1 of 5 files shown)


rpc/dial.go
```diff
@@ -114,28 +114,19 @@ func dial(
...
-			} else {
-				target = signalingAddress
-				port = 443
+			target, port, err := getWebRTCTargetFromAddressWithDefaults(signalingAddress)
+			if err != nil {
+				return nil, false, err
 			}
 			fixupWebRTCOptions(dOpts, target, port)
+
+			// When connecting to a external signaler for WebRTC we assume we can use the external Auth's material.
```
edaniels, line 123:
```suggestion
			// When connecting to an external signaler for WebRTC, we assume we can use the external auth's material.
```

rpc/dial.go
```diff
@@ -114,28 +114,19 @@ func dial(
...
+			// This path is also called by a mdns direct connection and ignore that case.
```
edaniels, line 124:
```suggestion
			// This path is also called by an mdns direct connection and ignores that case.
```

rpc/dial.go
```diff
@@ -311,3 +314,26 @@ func fixupWebRTCOptions(dOpts *dialOptions, target string, port uint16) {
 		dOpts.webrtcOpts.SignalingCreds = dOpts.creds
 	}
 }
+
+func getWebRTCTargetFromAddressWithDefaults(signalingAddress string) (target string, port uint16, err error) {
```
edaniels, line 318: ty

rpc/dial_options.go
```diff
@@ -41,14 +41,20 @@ type dialOptions struct {
 	externalAuthAddr     string
 	externalAuthToEntity string
 	externalAuthInsecure bool
+	// static auth material used when an external auth service is used. This is also used for the signaler
+	// when the webrtc options are empty. See fixupWebRTCOptions
```
edaniels, line 45:
```suggestion
	// when the webrtc options are empty. See fixupWebRTCOptions.
```

rpc/dial_options.go
```diff
@@ -41,14 +41,20 @@ type dialOptions struct {
...
+	// static auth material used when directly connecting to the endpoint. If set all externalAuth options are ignored.
 	authMaterial string
 
 	// debug is helpful to turn on when the library isn't working quite right.
 	// It will output much more logs.
 	debug bool
 
 	mdnsOptions DialMulticastDNSOptions
+	// set when the connection is using mdns flow.
+	usingMdns bool
```
edaniels, line 57:
```suggestion
	usingMDNS bool
```

rpc/dial_options.go
```diff
@@ -162,6 +168,16 @@ func WithStaticAuthenticationMaterial(authMaterial string) DialOption {
 	})
 }
 
+// WithStaticExternalAuthenticationMaterial returns a DialOption which sets the already authenticated
+// material (opaque) to use for authenticated requests against external auth service. Ignored if
```
edaniels, line 172:
```suggestion
// material (opaque) to use for authenticated requests against an external auth service. Ignored if
```

rpc/dialer.go
```diff
@@ -282,12 +282,19 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts *dialOptions, log
...
+	} else if dOpts.creds.Type != "" || dOpts.externalAuthMaterial != "" {
 		rpcCreds = &perRPCJWTCredentials{
 			entity: dOpts.authEntity,
 			creds:  dOpts.creds,
 			debug:  dOpts.debug,
 			logger: logger,
+			// Note: don't set dialOptsCopy.authMaterial below as perRPCJWTCredentials will know to use
+			// its externalAccessToken to authenticateTo. This will result in both a connection level authorization
+			// added as well as an authorization header added from perRPCJWTCredentials, resulting in a failure.
+			externalAccessToken: dOpts.externalAuthMaterial,
```
edaniels, line 297: sug: call this field `externalAuthMaterial` for consistency

rpc/dialer.go
```diff
@@ -302,11 +309,18 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts *dialOptions, log
 			dialOptsCopy := *dOpts
 			dialOptsCopy.insecure = dOpts.externalAuthInsecure
 			dialOptsCopy.externalAuthAddr = ""
+			dialOptsCopy.externalAuthMaterial = ""
 			dialOptsCopy.creds = Credentials{}
+			dialOptsCopy.authEntity = ""
+
+			// reset the tls config that is used for the external Auth Service.
+			dialOptsCopy.tlsConfig = &tls.Config{MinVersion: tls.VersionTLS12}
```
edaniels, line 317: might want to single source this so we don't miss any version details in future updates

rpc/dialer.go
```diff
@@ -462,6 +480,8 @@ type perRPCJWTCredentials struct {
 	externalAuthToEntity string
 	creds                Credentials
 	accessToken          string
+	// The static external auth material used against the AuthenticateTo request to obtain final accessToken
```
edaniels, line 483: q: it says this is used in with AuthenticateTo, but doing the `With` option causes the externalAuthToEntity field to be cleared in the options. Is that correct?

edaniels (changes requested):
basically lgtm mod seeing the tests, some nits, and 1 q

rpc/dialer.go
AdamMagaluk: Ah good catch. I think that was from an early copy and paste when i added the With option.  I changed to to remove the authEntity and Credentials{} instead of the authTo. The tests didn't catch that and my e2e test locally add the options added in an order that didn't cause issues.

rpc/dial.go
```diff
@@ -114,28 +114,20 @@ func dial(
...
+			if err != nil {
+				return nil, false, err
 			}
 			fixupWebRTCOptions(dOpts, target, port)
+
+			// When connecting to an external signaler for WebRTC, we assume we can use the external auth's material.
+			// This path is also called by an mdns direct connection and ignores that case.
+			// This will skip all Authenticate/AuthenticateTo calls for the signaler.
+			if !dOpts.usingMDNS && dOpts.authMaterial == "" && dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial != "" {
+				logger.Debugw("using singaling's external auth as auth material")
```
edaniels, line 127:
```suggestion
				logger.Debug("using signaling's external auth as auth material")
```

edaniels (approved):
LGTM

PR #119, opened 2023-01-07
RSDK-1660 - Fix deadlock in cachedDialer

Otterverse:
Comments inline to explain, but tl;dr is lock contention. The cachedDialer can be put into the context, and thus gets reused by various dial functions. Dialing with Auth, for example, reuses the dialer from different layers, resulting in the possibility of nested refCountedConnWrapper and thus, deadlock on Close.

Note: RDK needs to be bumped with this update after this merges.

### DIFF
--- a/rpc/dialer.go
+++ b/rpc/dialer.go
@@ -140,9 +140,15 @@ func (cd *cachedDialer) DialFunc(
 
 func (cd *cachedDialer) Close() error {
 	cd.mu.Lock()
-	defer cd.mu.Unlock()
-	var err error
+	// need a copy of cd.conns as we can't hold the lock, since .Close() fires the onUnref()
+	// which uses the same lock and directly modifies cd.conns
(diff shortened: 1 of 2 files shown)


edaniels (approved):
LGTM after test change discussed offline

edaniels:
@Otterverse went deep on this and your fix is definitely correct. I added some more comments in the code as to what's going on. Current tests didn't catch this because they used DialDirect which avoids the multi path which can cause a double Dial (totally fine). The part that wasn't fine is if you do a DialFunc(multi) which does a DialFunc(grpc) and so on, you end up reference counting multiple times and one of the things you reference count might be a cached connection itself. That means we do need your fix for the legit case where the `actual` connection is something cached. Admittedly this points out a design flaw that is wasting memory by me trying to me smart with recursion :(. It's not worth fixing right now though.

We should still investigate RDK though.

PR #123, opened 2023-01-10
APP-1153 - Support using access token with external auth in JS

AdamMagaluk:
JS RPC library updated to allow authToAddress and authToEntity with accessToken set in the dial options. This will force the accessToken to be used for the AuthenticateTo request but use the returned access token in signaling or direct rpc calls.

If authTo options are omitted while an accessToken is present all Authenticate/AuthenticateTo calls will be bypassed.


edaniels (approved):
LGTM!

PR #127, opened 2023-01-13
[RSDK-1676] Replace . with - when advertising robot over mdns.

npmenard:
This PR replace any  '.' in robot's fqdn or local_fqdn with '-' ('.' are not skipped properly when sending the advertisement packets therefore violating mDNS RFC) . The robot's mDNS records can now be properly parsed by other mDNS discovery tools, the robot can also question record from other implementation (mini-rdk) properly
The former service advertisement is kept for backward compatibility.

### DIFF
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -199,31 +199,44 @@ func dialMulticastDNS(
 	logger golog.Logger,
 	dOpts *dialOptions,
 ) (ClientConn, bool, error) {
-	resolver, err := zeroconf.NewResolver(zeroconf.SelectIPRecordType(zeroconf.IPv4))
+	candidates := []string{address, strings.ReplaceAll(address, ".", "-")}
+	candidateLookup := func(ctx context.Context, candidates []string) (*zeroconf.ServiceEntry, error) {
+		resolver, err := zeroconf.NewResolver(zeroconf.SelectIPRecordType(zeroconf.IPv4))
+		if err != nil {
+			return nil, err
+		}
+		defer resolver.Shutdown()
+		for _, candidate := range candidates {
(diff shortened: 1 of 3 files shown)


rpc/server.go
```diff
@@ -413,35 +413,41 @@ func NewServer(logger golog.Logger, opts ...ServerOption) (Server, error) {
...
-					"local.",
-					mDNSAddress.Port,
-					hostname,
-					[]string{"127.0.0.1"},
-					supportedServices,
-					[]net.Interface{loopbackIfc},
-				)
-				if err != nil {
-					return nil, err
+				hosts := []string{host, strings.ReplaceAll(host, ".", "-")}
```
edaniels, line 416: where is this advertising old ones?

rpc/dial.go
```diff
@@ -199,31 +199,44 @@ func dialMulticastDNS(
...
-	// entries gets closed after lookupCtx expires or there is a real entry
-	entry := <-entries
 	if entry == nil {
 		return nil, false, ctx.Err()
 	}
 	var hasGRPC, hasWebRTC bool
 	for _, field := range entry.Text {
-		if field == "grpc" {
+		// mdns service may advertise TXT field following https://datatracker.ietf.org/doc/html/rfc1464 (ex grpc=)
+		if strings.Contains(field, "grpc") {
```
edaniels, line 236: ah

rpc/server.go
edaniels: oh I see heh

edaniels (approved):
LGTM but would like to find a way to test this, potentially just doing a query in a test once server is running

PR #128, opened 2023-01-18
pexec: better support windows

edaniels:
I'll add a runner later to support this but need to deal with other issues first. Essentially, killing in Windows is very hard without signals and a custom made binary.


Otterverse (approved):
LGTM, mod the failing checks/tests.

PR #132, opened 2023-01-26
Add mongodb opencensus metrics

AdamMagaluk:
See: https://opencensus.io/integrations/mongodb/go_driver/


AdamMagaluk:
This is actually all we need (once we bump the utils version in app).

The instrumentations looks like it is inside the go-driver already and reporting metrics to opencensus. This piece says our default metrics collection should include in the views for Mongodb. And lastly our app already has the opencensus code to export everything to Google Cloud Monitoring.

PR #134, opened 2023-01-31
pexec: improve logs

edaniels:
8) - any feedback on what character to use would be appreciated

Screenshot 2023-01-30 at 8 05 56 PM


edaniels:
@AdamMagaluk also curious what you think of this. it's much more readable but potentially harder to get into a log? I'm assuming we will make this better over time with which logs we actually send up with potentially separate process loggers.

Otterverse (approved):
Definitely like breaking this up into a new line, will make things more readable for sure!

The unicode character you picked through is small (half-width), so looks a little weird. I thought my browser was glitching and only rendering part of a glyph when I first saw the diff.  `ps fauxwww` does a `\n \_ ` which I kind of prefer (just regular ascii chars), but I'm fine with this as-is too.

edaniels:
oh I like \_ 🤦‍♂️

PR #137, opened 2023-02-09
Github Actions tests pass on skipped

bazile-clyde:
See https://github.com/viamrobotics/goutils/actions/runs/4137170424/jobs/7152865166#step:2:40

This issue is causing a failure in my PR https://github.com/viamrobotics/goutils/pull/136. `build_and_test` is `skipped`. `"skipped" != "success"`. Therefore, `test_passing` fails.

Edit: It's actually causing the same failure here. Not sure if there's a way to just push this through.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Check Results
         run: |
           echo Main Tests: ${{ needs.build_and_test.result }}
-          [ "${{ needs.build_and_test.result }}" == "success" ]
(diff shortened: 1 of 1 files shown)


edaniels:
We can't let a merge happen if tests get skipped. It looks like labeling is wrong.

edaniels:
Fixed the labeling issue in https://github.com/viamrobotics/goutils/commit/39301e05bdf92ca0fc631c59ecaa20f0e3cd532d

PR #140, opened 2023-02-22
Add more perf for MongoDB and add local metric exporter

edaniels:


perf/dev_exporter.go
```diff
@@ -0,0 +1,160 @@
...
+
+	spanID := hex.EncodeToString(sd.SpanID[:])
+	parentSpanID := hex.EncodeToString(sd.ParentSpanID[:])
+
+	if !reZero.MatchString(parentSpanID) {
+		e.children[parentSpanID] = append(e.children[parentSpanID], mySpanInfo{myinfo, spanID})
+		return
+	}
+
+	// i'm the top of the tree, go me
```
DTCurrie, line 157: 😄

DTCurrie (approved):
✅

perf/dev_exporter.go
edaniels: lol this is eliot :P

perf/mongodb.go
```diff
@@ -1,9 +1,127 @@
...
+	b, err := bson.MarshalExtJSON(evt.Command, false, false)
+	if err != nil {
+		utils.UncheckedError(err)
+		return
+	}
+	_, span := trace.StartSpan(ctx, evt.CommandName, trace.WithSampler(m.cfg.sampler))
+	span.AddAttributes(trace.StringAttribute(ext.DBInstance, evt.DatabaseName),
+		trace.StringAttribute("db.system", "mongodb"),
+		trace.StringAttribute("db.operation", string(b)),
+		trace.StringAttribute("db.connection_string", connString),
```
michaellee1019, line 60: Does this accidentally leak secrets when password is used in connection string? Maybe just use name of the cluster/host?

perf/dev_exporter.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"go.opencensus.io/metric/metricdata"
+	"go.opencensus.io/metric/metricexport"
+	"go.opencensus.io/trace"
+
+	"go.viam.com/utils"
+)
+
+// DevelopmentExporter exports metrics and span to log file.
+type DevelopmentExporter struct {
```
michaellee1019, line 23: Is this copy/paste from some other package? wondering how we can re-use centrally (can be handled later).

perf/mongodb.go
edaniels: no, see the function it derives it from

perf/dev_exporter.go
edaniels: yes, the comment below package says `based on "go.opencensus.io/examples/exporter"`. This makes some choices on it. Not sure what you mean by reuse centrally? This will be used in app.

PR #142, opened 2023-02-28
Allow for Lax Session Cookie.

AdamMagaluk:
- Allows being able to link from external web page to Viam.com and have auth work. Still protects against most CSRF assuming GET requests has no affect on system.


edaniels (approved):
🙇

PR #145, opened 2023-03-12
Refactor web auth to be more compliant with external auth

edaniels:
Wanted to start getting eyes on this. There's a lot of changes here but overall this ends up both consolidating concepts from web auth into external. Some auth concepts have also been refactored or ratified. Namely:

* Audiences can be specified by a server are always verified
* Issuers can be specified by a server (not verified)
* Auth handling is split into 3 separately specified options (AuthHandler, EntityDataLoader, TokenVerificationKeyProvider
  * This was to eliminate some interface embedding anti-patterns we had
 * More support for OIDC than just auth0's patterns

Still need to work on some tests


jwks/jwks.go
```diff
@@ -73,32 +76,50 @@ func (cp *cachingKeyProvider) Fetch(ctx context.Context) (KeySet, error) {
...
+// base domain and starts the auto refresh. Call CachingKeyProvider.Stop() to stop any
+// background goroutines.
+func NewCachingOIDCJWKKeyProvider(ctx context.Context, issuer string) (KeyProvider, error) {
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	httpClient := &http.Client{
+		Transport: httpTransport,
+	}
+	defer httpTransport.CloseIdleConnections()
+
+	wellKnown := strings.TrimSuffix(issuer, "/") + oidc.DiscoveryEndpoint
```
edaniels, line 89: Now we can support a lot more OIDC providers

proto/rpc/v1/auth.proto
```diff
@@ -9,9 +9,11 @@ import "google/api/annotations.proto";
 // authentication. Its sole Authenticate method should be used prior to any
 // other services that a gRPC server has to offer.
 service AuthService {
-	// Authenticate attempts to authenticate the caller. The resulting
-	// response contains an access token that should be used for future
-	// requests.
+	// Authenticate attempts to authenticate the caller claiming to be
```
edaniels, line 12: Hopefully this clears up some questions for future readers.

edaniels: I should have never called these entities but oh well

rpc/dialer.go
```diff
@@ -307,14 +307,15 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts *dialOptions, log
...
 		}
-		if dOpts.debug {
+		if dOpts.debug && dOpts.externalAuthAddr != "" && dOpts.externalAuthToEntity != "" {
 			logger.Debugw("will eventually authenticate as entity", "entity", dOpts.authEntity)
 		}
 		if dOpts.externalAuthAddr != "" {
-			if dOpts.debug && dOpts.externalAuthToEntity != "" {
-				logger.Debugw("will eventually externally authenticate to entity", "entity", dOpts.externalAuthToEntity)
+			if dOpts.externalAuthToEntity == "" {
+				return nil, false, errors.New("external auth address set but no authenticate to option set")
```
edaniels, line 315: This is new. We've never used this particular combination of settings and I don't see how it's useful to use your externally got credentials without using authTo. Felt like a bug to me but I had tests written to support it.

rpc/server_auth.go
```diff
@@ -277,70 +274,48 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (string, interface{},
...
-		// reset the claims to the handlers version
-		claims = provider.CreateClaims()
-		if claims == nil {
-			return "", nil, status.Error(codes.Internal, "invalid implementation of TokenCustomClaimProvider, cannot return nil")
+	// Audience verification is critical for security. Without it, we have a higher chance
+	// of validating a JWT is valid, but not that it is intended for us. Of course, that means
+	// we trust whomever owns the private keys to signing access tokens.
+	audVerified := false
+	for _, allowdAud := range ss.authAudience {
+		if claims.RegisteredClaims.VerifyAudience(allowdAud, true) {
```
edaniels, line 286: This is new and will require some robot/client restarts. I don't think it will cause any issues but flagging it.

rpc/auth.go
```diff
@@ -13,35 +13,49 @@ import (
 	"github.com/golang-jwt/jwt/v4"
```
edaniels, line 13: I'd say this file is where all the interesting API changes are happening. We don't have this module versioned so we are free to break things. I'd call this v2 of rpc auth with respect to the API. Using this feels much better to me now and makes more sense to me.

rpc/server_auth.go
michaellee1019: [minor] To reduce issues we could just `VerifyAudience(allowdAud) || VerifyAudience(claims.RegisteredClaims.Subject)` for 24hrs and then remove the 2nd check in a cleanup PR?

rpc/server_options.go
```diff
@@ -179,12 +190,15 @@ func WithStreamServerInterceptor(streamInterceptor grpc.StreamServerInterceptor)
 }
 
 // WithInstanceNames returns a ServerOption which sets the names for this
-// server instance. These names will be used for mDNS service discovery to
-// report the server itself. If unset the value is the address set by
-// WithExternalListenerAddress, WithInternalBindAddress, or the localhost and random port address,
-// in preference order from left to right.
+// server instance. These names will be used for auth token issuance (first name) and
```
michaellee1019, line 193:
```suggestion
// server instance. These names will be used for auth token issuance (first item in list) and
```

proto/rpc/v1/auth.proto
michaellee1019: [nit] need some generic name anyways to support multiple types (users, robots, service_account, etc)

proto/rpc/v1/auth.proto
```diff
@@ -26,9 +28,12 @@ service AuthService {
...
-	// The resulting response contains an access token that should be used for future
-	// requests. This assumes that the caller is already authenticated to the
+	// AuthenticateTo attempts to allow the caller to authenticate to another entity.
+	// The resulting response contains an access token that with the subject
+	// as the calling entity, the audience as the other entity, and the issuer
+	// as the provider of this service. This token should be used for all
+	// future RPC requests to the other entity on the services it provides.
+	// This assumes that the caller is already authenticated to the
 	// server implementing this service.
 	rpc AuthenticateTo(AuthenticateToRequest) returns (AuthenticateToResponse) {
```
michaellee1019, line 38: [opt] Do you think we should/could change `AuthenticateTo` to something else so the API name represents more of what it is doing? `AssumeEntity`, `RepresentEntity`, `ImpersonateUser` etc

rpc/dialer.go
michaellee1019: [q] So our customers always have to AuthTo a user in their org in order to use the API?

michaellee1019 (changes requested):

proto/rpc/v1/auth.proto
```diff
@@ -26,9 +28,12 @@ service AuthService {
...
 // entity's gRPC server has to offer.
 service ExternalAuthService {
-	// AuthenticateTo attempts to authenticate the caller on behalf of an entity.
-	// The resulting response contains an access token that should be used for future
-	// requests. This assumes that the caller is already authenticated to the
+	// AuthenticateTo attempts to allow the caller to authenticate to another entity.
+	// The resulting response contains an access token that with the subject
+	// as the calling entity, the audience as the other entity, and the issuer
+	// as the provider of this service. This token should be used for all
+	// future RPC requests to the other entity on the services it provides.
```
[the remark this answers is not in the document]
edaniels: Keep in mind this is in goutils, not app. As an open source utility, I think this is true.

proto/rpc/v1/auth.proto
edaniels: That would be breaking the proto which I'm trying to avoid here. I like AssumeEntity though. If we want to change, we should file a follow up ticket since the TS SDK would have to change too.

jwks/jwks.go
edaniels: It's open source code, not for our product necessarily.

rpc/dialer.go
edaniels: Nope, just to control/use to a robot if they are passing JWTs to the robot directly (not using our webrtc signaling in the cloud)

rpc/server_auth.go
edaniels: I'm not sure that would fix it. We are now in general verifying all audiences so I think your suggestion would still fail?

rpc/auth.go
```diff
@@ -65,105 +76,84 @@ type Claims interface {
...
-	return customClaimAuthHandler{AuthHandler: handler, claimFunc: claimFunc}
+// MakePublicKeyProvider returns a TokenVerificationKeyProvider that provides a public key for JWT verification.
+func MakePublicKeyProvider(pubKey *rsa.PublicKey) TokenVerificationKeyProvider {
+	return TokenVerificationKeyProviderFunc(
+		func(ctx context.Context, token *jwt.Token) (interface{}, error) {
+			if _, ok := token.Method.(*jwt.SigningMethodRSA); !ok {
+				return nil, fmt.Errorf("unexpected signing method %q", token.Method.Alg())
+			}
+
+			return pubKey, nil
```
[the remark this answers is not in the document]
edaniels: not sure what you mean. the pubKey comes from the caller here and a TokenVerificationKeyProvider does not return rsa pub keys necessarily. so we either do a closure like we do here or have a struct that implements the provider and returns the public key (still a closure)

rpc/server_auth.go
```diff
@@ -122,20 +127,23 @@ func (ss *simpleServer) AuthenticateTo(ctx context.Context, req *rpcpb.Authentic
...
 	forType CredentialsType,
-	entity, subject string,
+	audience []string,
+	entity string,
 	authMD map[string]string,
 ) (string, error) {
-	// TODO(RSDK-890): use the correct subject, not the audience (entity)
+	// TODO(GOUT-12): refresh token
+	// TODO(GOUT-9): more complete info
 	token := jwt.NewWithClaims(jwt.SigningMethodRS256, JWTClaims{
```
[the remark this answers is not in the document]
edaniels: log what?

proto/rpc/v1/auth.proto
edaniels: Might be a good task for you as well!

michaellee1019: True and sounds good. I'd be concerned if we don't do this before our first External API customer (not partner) uses it.

edaniels: Totally agree

michaellee1019: Added a Jira to not forget: https://viam.atlassian.net/browse/APP-1444

rpc/auth.go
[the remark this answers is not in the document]
michaellee1019: Eh its fine, ignore me. I'm new to go functions with closures and just don't like how some variables don't get passed into the function

rpc/server_auth.go
michaellee1019: Ok. I was just looking at the diff. previously the Audience attribute was set to `Audience:  jwt.ClaimStrings{entity}` and now is set to `Audience:  audience` on https://github.com/viamrobotics/goutils/pull/145/files#diff-988c4b375bc405125e428ecc73b8b3ee9da068e30a541214b93d20fced98e856R139

So I thought the incompatibility you were talking about is that for 24hrs the wrong audience value would be validated.

michaellee1019 (approved):

PR #153, opened 2023-04-10
mongo-test-utils: Add the ability to get randomized db / col names from real namespace

RoxyFarhad:
For the Billing Test on App, we need to add default data to the db i.e. billing tiers and cost information which is a write method we don't want to expose on our data api. Because we randomize db and column names, i need some testing utils that allow me to access the collection from the app. 

This change allows me to do so by just storing the mapping from original db / col names <> randomized names, and creating a function that returns the randomized collection name from the original. With the knowledge of the randomized name. I can access the collection from the APP to do the setup i require. 

I tested this via APP to get the billing tiers collection, and running the setup correctly.

### DIFF
--- a/mongo/namespace.go
+++ b/mongo/namespace.go
@@ -12,7 +12,13 @@ var (
 	namespaces          = map[*string][]*string{}
 	unmanagedNamespaces = map[string][]string{}
 	namespacesMu        sync.Mutex
+	oldNamespaces       = map[RandomizedName][]RandomizedName{}
 )
+
+type RandomizedDbAndName struct {
+	RandomizedDbName  string
+	RandomizedColName string
+}
 
 // RegisterNamespace globally registers the given database and collection as in use
 // with MongoDB. It will error if there's a duplicate registration.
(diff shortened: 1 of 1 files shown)


RoxyFarhad:
Yeah, I wasn't sure if this was the best way of doing it but i wasn't sure how else to do this. 

Yes seeding the data would be the best way of doing this. I wasn't sure if we have a way of doing this right now as I couldn't see an example of APP, how would you suggest going about this? 

I agree that the data should be able to use the collection without knowing the details, but how is this currently possible? On our data api we only expose I/O functions to the DB & Collections, and we don't expose the actual collections. My issue is that I don't want to expose functions to 1. create seed data (because that shouldn't be exposed in APP) 2. test the write methods because that requires functionality that we don't actually need in APP. 

> You can also always access the DB/Coll name via exported pointers that will be updated to the randomization.

How do we do this now? I thought my change was allowing for that.

edaniels:
So for example, we do `mongoutils.MustRegisterNamespace(&DBName, &someCollName)`. In this case, anything that refers to those strings after that call will have the randomized dbname and coll name. In this case, DBName is exported and accessible, but coll is not (but could be)

RoxyFarhad:
@edaniels Ah okay exposing this names worked thank you! I didn't realise that the local collection names would be updated  to the randomized name space. Thank you! Will close this.

PR #154, opened 2023-04-17
rpc: wait for reset background goroutine on channel close

edaniels:


rpc/wrtc_client_channel_test.go
```diff
@@ -412,6 +410,8 @@ func TestWebRTCClientChannelResetStream(t *testing.T) {
 	err = clientCh.Invoke(ctx, "thing", someStatus.Proto(), &respStatus)
 	test.That(t, err, test.ShouldBeError, context.Canceled)
 	test.That(t, &respStatus, test.ShouldResemble, &pbstatus.Status{})
+
+	<-resetCh
```
edaniels, line 414: verified that test hangs if i remove reset code

PR #155, opened 2023-04-20
RSDK-2425 Add OnCrashHandler to managed processes

benjirewis:
[RSDK-2425](https://viam.atlassian.net/browse/RSDK-2425)

Adds `OnCrashHandler` to `ProcessConfig`, a `func() bool` to be called when the `manage` goroutine detects a crash (unexpected stop) of the process. If the returned bool is true, the `manage` goroutine will attempt to restart the process. Otherwise, the `manage` goroutine will simply return. Adds tests for `OnCrashHandler()` in various use-cases.

Useful generally, but more specifically for handling crashed modules, which are run as `ManagedProcess`es.


pexec/managed_process.go
```diff
@@ -173,12 +175,13 @@ func (p *managedProcess) Start(ctx context.Context) error {
...
-// manage is the watchdog of the process. Any time it detects
-// the process has ended unexpectedly, it will restart it. It's
-// possible and okay for a restart to be in progress while a Stop
-// is happening. As a means simplifying implementation, a restart
-// spawns new goroutines by calling Start again and lets the original
-// goroutine die off.
+// manage is the watchdog of the process. If the process has ended
+// unexpectedly, onCrashHandler will be called. If onCrashHandler is unset or
+// returns true, manage will restart the process. Note that onCrashHandler may
+// be called multiple times if it returns true. It's possible and okay for a
```
benjirewis, line 181: > Not that onCrashHandler may be called multiple times if it returns true.

I'm not sure how I feel about this behavior. At first, I thought we would only call the `onCrashHandler` once when processes crash. If the process crashes, though, `manage` will restart both the process _and_ another `manage` goroutine; this leads to potentially infinite restarts, and I can't think of a way to differentiate between sequential quick crashes where `onCrashHandler` should only be called once and long runs of the process with intermittent crashing where `onCrashHandler` should maybe be called on each crash.

pexec/managed_process_test.go
```diff
@@ -193,6 +209,63 @@ func TestManagedProcessManage(t *testing.T) {
...
+			proc := NewManagedProcess(ProcessConfig{
+				Name:           "bash",
+				Args:           []string{"-c", "exit 1"},
+				OnCrashHandler: func() bool { onCrashHandlerCallCount++; return false },
+			}, logger)
+			test.That(t, proc.Start(context.Background()), test.ShouldBeNil)
+
+			// OnCrashHandler should be called once, as program should crash once and
+			// not restart.
+			time.Sleep(2 * time.Second)
```
benjirewis, line 259: This test is bad, and would love any other ideas on better ways to write it. 

The trouble is I'm trying to assert that `onCrashHandler` is called once _and only_ once when it returns false. There's nothing to "wait" for here that will tell me once the process has actually finished. Using a strategy like that of the test above (waiting for writes to a file) does not work: even after the write is awaited (`<-watcher.Events`), the `onCrashHandler` will not have been called yet. The `Sleep` "works" in that it tests that `onCrashHandlerCount` is called once within the two second window, but it's racy and doesn't really test that `onCrashHandlerCount` will be called _only_ once.

pexec/managed_process.go
```diff
@@ -265,6 +268,14 @@ func (p *managedProcess) manage(stdOut, stdErr io.ReadCloser) {
 	default:
 	}
 
+	// Run onCrashHandler if it exists. Do not attempt restart if onCrashHandler
+	// returns false.
+	if p.onCrashHandler != nil {
+		if shouldRestart := p.onCrashHandler(); !shouldRestart {
+			return
+		}
+	}
```
Otterverse, line 277:
```suggestion
	if p.onCrashHandler != nil && !p.onCrashHandler() {
		return
	}
```
Pretty sure that simplifies it correctly.

pexec/managed_process_test.go
```diff
@@ -193,6 +209,63 @@ func TestManagedProcessManage(t *testing.T) {
...
+			watcher, err := fsnotify.NewWatcher()
+			test.That(t, err, test.ShouldBeNil)
+			defer watcher.Close()
+			watcher.Add(tempFile.Name())
+
+			var onCrashHandlerCallCount int
+			proc := NewManagedProcess(ProcessConfig{
+				Name:           "bash",
+				Args:           []string{"-c", fmt.Sprintf("echo hello >> '%s'\nexit 1", tempFile.Name())},
+				OnCrashHandler: func() bool { onCrashHandlerCallCount++; return true },
```
Otterverse, line 228: What stops this from just restarting indefinitely? Shouldn't it return false after, say, a count of 5 or something?

pexec/managed_process_test.go
Otterverse: I don't think this is TOO bad... a two second sleep isn't the end of the world here.

That said, I would consider a dynamic function in the test per my suggestion above. Maybe combine both these tests... have it restart five times, then sleep a couple seconds, and verify that the count is EXACTLY five.

pexec/managed_process_test.go
Otterverse: To be clear, it _looks_ like it's a race condition... the watcher gets three hits and the call count is checked RIGHT after that, even though the process would continue to restart until proc.Stop(). If that's NOT the case, then please explain where I'm wrong.

pexec/managed_process_test.go
```diff
@@ -193,6 +209,63 @@ func TestManagedProcessManage(t *testing.T) {
...
+
+			tempFile := testutils.TempFile(t, "something.txt")
+			defer tempFile.Close()
+
+			watcher, err := fsnotify.NewWatcher()
+			test.That(t, err, test.ShouldBeNil)
+			defer watcher.Close()
+			watcher.Add(tempFile.Name())
+
+			var onCrashHandlerCallCount int
```
Otterverse, line 224: Make and use this as an atomic, as it's getting incremented into a routine where it COULD be accessed simultaneously. Currently anyway. If you modify to fail after a count of X, and only read after that, then it's probably okay as is.

Otterverse (changes requested):
Looks mostly good. A few minor things, mostly in testing.

benjirewis:
@Otterverse thanks for the feedback. Combined the tests into a single test with some safer concurrency (channel and an atomic int). The previous tests did have race conditions (failed with `-race`); this one seems fine. Let me know what you think.

pexec/managed_process_test.go
benjirewis: > Maybe combine both these tests... have it restart five times, then sleep a couple seconds, and verify that the count is EXACTLY five.

Did exactly this with some safer concurrency.

Otterverse (approved):
LGTM, save one optional nit to make it a little cleaner if you want.

pexec/managed_process_test.go
```diff
@@ -193,6 +210,40 @@ func TestManagedProcessManage(t *testing.T) {
 			test.That(t, err.Error(), test.ShouldContainSubstring, "exit status 1")
 		}
 	})
+	t.Run("OnCrashHandler", func(t *testing.T) {
+		logger := golog.NewTestLogger(t)
+
+		crashHandlerCalledEnough := make(chan struct{})
+		var onCrashHandlerCallCount uint64
```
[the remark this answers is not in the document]
benjirewis: Ah great point. Switched to `atomic.Unit64` and associated methods.

pexec/managed_process.go
```diff
@@ -265,6 +268,12 @@ func (p *managedProcess) manage(stdOut, stdErr io.ReadCloser) {
 	default:
 	}
 
+	// Run onCrashHandler if it exists. Do not attempt restart if onCrashHandler
+	// returns false.
+	if p.onCrashHandler != nil && !p.onCrashHandler() {
```
edaniels, line 273: I think `onUnexpectedExit` would be nicer. Also using `on` obviates the need for `Handler`.

edaniels: Can (should?) we pass the exit code in?

edaniels (changes requested):
Done with first pass. basically lgtm

pexec/managed_process_test.go
```diff
@@ -193,6 +210,38 @@ func TestManagedProcessManage(t *testing.T) {
...
+				return true
+			},
+		}, logger)
+		test.That(t, proc.Start(context.Background()), test.ShouldBeNil)
+
+		<-crashHandlerCalledEnough
+
+		// Assert onCrashHandlerCallCount is exactly five even after waiting a few
+		// seconds (no further restarts).
+		time.Sleep(2 * time.Second)
```
edaniels, line 236: It seems like you should not need this sleep. by calling stop later and switching it to >= 5, you know test messed up if it double closes

pexec/managed_process.go
benjirewis: Changed to `OnUnexpectedExit` and made the signature `func(int) bool` to have access to exit code. Would be nice for exit handler to better understand what happened to the crashed process.

edaniels: definitelyl

edaniels (approved):
LGTM!

edaniels:
just need to update PR title :)

PR #157, opened 2023-04-21
RSDK-2799 Get additional turn servers

maximpertsov:


rpc/js/src/dial.ts
```diff
@@ -218,6 +261,18 @@ export async function dialWebRTC(signalingAddress: string, host: string, opts?:
 	validateDialOptions(opts);
 
 	const webrtcOpts = opts?.webrtcOptions;
+  if (webrtcOpts?.rtcConfig?.iceServers) {
```
edaniels, line 264: This should basically follow https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_client.go#L160-L175 exactly. We always try to extend the list, not when servers are already specified.

edaniels (changes requested):

rpc/js/src/dial.ts
```diff
@@ -217,7 +260,42 @@ interface WebRTCConnection {
...
+      username: ice.username,
+    }
+  });
+
+  if (!opts) {
+    opts = {};
+  }
+
+	let webrtcOpts: DialWebRTCOptions;
+  if (!opts.webrtcOptions) {
```
maximpertsov, line 279:
```suggestion
  let webrtcOpts: DialWebRTCOptions;
  if (!opts.webrtcOptions) {
```

format

rpc/js/src/dial.ts
```diff
@@ -217,7 +260,42 @@ interface WebRTCConnection {
...
+    return {
+      urls: ice.urlsList,
+      credential: ice.credential,
+      username: ice.username,
+    }
+  });
+
+  if (!opts) {
+    opts = {};
+  }
```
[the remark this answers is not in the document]
maximpertsov: will address as part of follow-up to ensure js/golang logic are in parity

rpc/js/src/dial.ts
```diff
@@ -209,6 +209,49 @@ interface WebRTCConnection {
...
+        'rpc-host': host,
+      },
+      host: signalingAddress,
+      transport: directTransport,
+      onEnd: (resp: grpc.UnaryOutput<OptionalWebRTCConfigResponse>) => {
+        const { status, statusMessage, message } = resp;
+        if (status === grpc.Code.OK && message) {
+          result = message.getConfig();
+          if (!result) {
+            pReject("no config");
```
edaniels, line 237: I would presolve with nothing and have that be optional checked. since this is all optional, rejection feels odd (no one likes rejection!)

edaniels (approved):

rpc/js/src/dial.ts
maximpertsov: sure, will just return an empty/default config in this case: https://github.com/viamrobotics/goutils/pull/157/commits/a7b5f22f45590ea2e261dd5d8bd4abedae081c41

PR #163, opened 2023-04-28
RSDK-2545 Allow fail-through for mDNS

Otterverse:
This covers a case where the loopback adapter doesn't support multicast, such as in some docker containers (usually cross-architecture emulation.)

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -41,7 +41,10 @@ import (
 	webrtcpb "go.viam.com/utils/proto/rpc/webrtc/v1"
 )
 
-const generatedRSAKeyBits = 4096
+const (
+	generatedRSAKeyBits = 4096
+	mDNSerr             = "mDNS setup failed; continuing with mDNS disabled"
+)
 
 // A Server provides a convenient way to get a gRPC server up and running
 // with HTTP facilities.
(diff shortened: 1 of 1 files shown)


rpc/server.go
```diff
@@ -429,24 +429,26 @@ func NewServer(logger golog.Logger, opts ...ServerOption) (Server, error) {
...
-						mDNSAddress.Port,
-						hostname,
-						[]string{"127.0.0.1"},
-						supportedServices,
-						[]net.Interface{loopbackIfc},
-						logger,
-					)
-					if err != nil {
-						return nil, err
+			if loopbackIfc.Index > 0 {
```
Otterverse, line 432: This is really the only change... the the rest is identical, just indented inside this if block now.

edaniels (approved):
LGTM if tested locally in and out of docker on linux+mac

rpc/server.go
```diff
@@ -419,14 +422,12 @@ func NewServer(logger golog.Logger, opts ...ServerOption) (Server, error) {
 			if err != nil {
 				return nil, err
 			}
-			var loopbackIfc net.Interface
+			var loopbackIfaces []net.Interface
 			for _, ifc := range ifcs {
-				if (ifc.Flags&net.FlagUp) == 0 ||
-					(ifc.Flags&net.FlagLoopback) == 0 ||
-					(ifc.Flags&net.FlagMulticast) == 0 {
+				if (ifc.Flags&net.FlagUp) == 0 || (ifc.Flags&net.FlagLoopback) == 0 {
```
Otterverse, line 427: So by default linux doesn't have the "multicast" flag set on the loopback interface, so I removed that check. It appears that you can join a group on an interface without that being set explicitly, and all tests still pass.

Otterverse (commented):
Rewrote from first version with fail-through. (Previous solution disabled mDNS even on working linux systems, so was a no-go.)

edaniels (approved):

PR #166, opened 2023-05-09
APP-1218: Adding unauth option for health check

RoxyFarhad:
We want to add a health check to grpc for deployment. To do so, we need to bypass the authentication methods required for the rest of the app. 

This is part of 3 prs that will need to be merged, but this is the first:
https://github.com/viamrobotics/api/pull/258

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -44,6 +44,8 @@ import (
 const (
 	generatedRSAKeyBits = 4096
 	mDNSerr             = "mDNS setup failed; continuing with mDNS disabled"
+	healthCheckMethod   = "/grpc.health.v1.Health/Check"
+	healthWatchMethod   = "/grpc.health.v1.Health/Watch"
 )
 
 // A Server provides a convenient way to get a gRPC server up and running
@@ -364,6 +366,11 @@ func NewServer(logger golog.Logger, opts ...ServerOption) (Server, error) {
 		}
 		// Update this if the proto method or path changes
 		server.exemptMethods["/proto.rpc.v1.AuthService/Authenticate"] = true
+	}
+
+	if sOpts.allowUnauthenticatedHealthCheck {
+		server.exemptMethods[healthCheckMethod] = true
+		server.exemptMethods[healthWatchMethod] = true
 	}
 
 	if sOpts.authToHandler != nil {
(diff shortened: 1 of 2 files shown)


rpc/server.go
```diff
@@ -366,6 +366,11 @@ func NewServer(logger golog.Logger, opts ...ServerOption) (Server, error) {
 		server.exemptMethods["/proto.rpc.v1.AuthService/Authenticate"] = true
 	}
 
+	if sOpts.allowUnauthenticatedHealthCheck {
+		server.exemptMethods["/viam.health.v1.Health/Check"] = true
```
edaniels, line 370: why not use the grpc health service https://github.com/grpc/grpc/blob/master/doc/health-checking.md? also, could we run a grpc server on a separate port for this so we dont expose this to the public for app.viam

RoxyFarhad: We can use this: https://pkg.go.dev/google.golang.org/grpc/health to implement the check, but i wanted to added a check for the mongo DB too as we have had deployments cause the server to go down in the past.

edaniels: so what I'm saying is we just implement that service instead of using the default code.

edaniels: and having a health check port option

rpc/server.go
```diff
@@ -44,6 +44,8 @@ import (
 const (
 	generatedRSAKeyBits = 4096
 	mDNSerr             = "mDNS setup failed; continuing with mDNS disabled"
+	healthCheckMethod   = "/grpc.health.v1.Health/Check"
```
RoxyFarhad, line 47: taken directly from the grpc.health pacakge

edaniels:
This is an after thought after mulling over the security of this, so my apologies. So I think overall we can just run a gRPC server at a port we define in our code inside our app server, not here. That will avoid this package and prevent any changes needed here.

RoxyFarhad:
> This is an after thought after mulling over the security of this, so my apologies. So I think overall we can just run a gRPC server at a port we define in our code inside our app server, not here. That will avoid this package and prevent any changes needed here.

So running a completely separate grpc server that app initializes? Is there any chance that running this not on the same rpc Server that the actual app/data/etc services run on will actually indicate valid checks? 

One thing that I am also slightly confused about is how to make sure there are no port collisions between the health check port and the the regular server port because that seems to be set randomly? Is there any way to access the port value for the server?

edaniels (approved):

PR #167, opened 2023-05-09
rpc: mitigate some client close error races

edaniels:
This also makes sure we serialize writes coming in more clearly

### DIFF
--- a/rpc/wrtc_base_channel.go
+++ b/rpc/wrtc_base_channel.go
@@ -144,11 +144,11 @@ func (ch *webrtcBaseChannel) closeWithReason(err error) error {
 	ch.closedReason = err
 	ch.cancel()
 	ch.bufferWriteCond.Broadcast()
-	ch.activeBackgroundWorkers.Wait()
 	return ch.peerConn.Close()
 }
 
 func (ch *webrtcBaseChannel) Close() error {
+	defer ch.activeBackgroundWorkers.Wait()
 	return ch.closeWithReason(nil)
 }
 
--- a/rpc/wrtc_base_channel.go
+++ b/rpc/wrtc_base_channel.go
@@ -203,7 +203,6 @@ func (ch *webrtcBaseChannel) write(msg proto.Message) error {
 		ch.bufferWriteCond.L.Unlock()
 		break
 	}
-
 	if err := ch.dataChannel.Send(data); err != nil {
 		if strings.Contains(err.Error(), "sending payload data in non-established state") {
 			return io.ErrClosedPipe
--- a/rpc/wrtc_base_channel.go
+++ b/rpc/wrtc_base_channel.go
@@ -144,11 +144,11 @@ func (ch *webrtcBaseChannel) closeWithReason(err error) error {
 	ch.closedReason = err
 	ch.cancel()
 	ch.bufferWriteCond.Broadcast()
+	ch.activeBackgroundWorkers.Wait()
 	return ch.peerConn.Close()
 }
 
 func (ch *webrtcBaseChannel) Close() error {
-	defer ch.activeBackgroundWorkers.Wait()
 	return ch.closeWithReason(nil)
 }
 
@@ -203,6 +203,7 @@ func (ch *webrtcBaseChannel) write(msg proto.Message) error {
 		ch.bufferWriteCond.L.Unlock()
 		break
 	}
+
 	if err := ch.dataChannel.Send(data); err != nil {
 		if strings.Contains(err.Error(), "sending payload data in non-established state") {
 			return io.ErrClosedPipe
--- a/rpc/wrtc_client_stream.go
+++ b/rpc/wrtc_client_stream.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"io"
 	"math"
-	"sync"
 
 	"github.com/edaniels/golog"
 	protov1 "github.com/golang/protobuf/proto" //nolint:staticcheck
@@ -26,7 +25,6 @@ type webrtcClientStream struct {
 	*webrtcBaseStream
 	ctx              context.Context
 	cancel           func()
-	mu               sync.RWMutex
 	ch               *webrtcClientChannel
 	headers          metadata.MD
 	trailers         metadata.MD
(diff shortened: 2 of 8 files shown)


rpc/wrtc_base_channel.go
```diff
@@ -144,11 +144,11 @@ func (ch *webrtcBaseChannel) closeWithReason(err error) error {
 	ch.closedReason = err
 	ch.cancel()
 	ch.bufferWriteCond.Broadcast()
+	ch.activeBackgroundWorkers.Wait()
```
edaniels, line 147: we may as well wait after we cancel

rpc/wrtc_base_channel.go
```diff
@@ -203,6 +203,12 @@ func (ch *webrtcBaseChannel) write(msg proto.Message) error {
 		ch.bufferWriteCond.L.Unlock()
 		break
 	}
+	ch.mu.Lock()
+	defer ch.mu.Unlock()
+	if ch.closed {
```
edaniels, line 208: we should be only allowing one write at a time

rpc/wrtc_client_stream.go
```diff
@@ -83,8 +84,7 @@ func newWebRTCClientStream(
 //
 // It is safe to have a goroutine calling SendMsg and another goroutine
 // calling RecvMsg on the same stream at the same time, but it is not safe
-// to call SendMsg on the same stream in different goroutines. It is also
-// not safe to call CloseSend concurrently with SendMsg or resetStream.
+// to call SendMsg on the same stream in different goroutines.
```
edaniels, line 87: users should get proper errors now if they do things concurrently and in a bad way

rpc/wrtc_signaling_answerer.go
```diff
@@ -76,6 +76,16 @@ func (ans *webrtcSignalingAnswerer) Start() {
 	}
 }
 
+func checkExceptionalError(err error) error {
```
edaniels, line 79: just a helper

rpc/wrtc_server_stream.go
```diff
@@ -337,7 +338,12 @@ func (s *webrtcServerStream) processMessage(msg *webrtcpb.RequestMessage) {
 	}
 }
 
-func (s *webrtcServerStream) closeWithSendError(err error) error {
+func (s *webrtcServerStream) closeWithSendError(err error) (writeErr error) {
+	defer func() {
+		if writeErr == nil || errors.Is(writeErr, sctp.ErrStreamClosed) {
```
edaniels, line 343: found this along the way

rpc/wrtc_server_channel.go
```diff
@@ -106,14 +106,14 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
 	serverStream, ok := ch.streams[id]
 	if !ok {
 		if len(ch.streams) == WebRTCMaxStreamCount {
-			ch.webrtcBaseChannel.logger.Error(errWebRTCMaxStreams)
+			logger.Error(errWebRTCMaxStreams)
```
edaniels, line 109: better to use this logger

rpc/wrtc_server_channel.go
```diff
@@ -106,14 +106,14 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
...
 			ch.mu.Unlock()
 			return
 		}
 		// peek headers for timeout
 		headers, ok := req.Type.(*webrtcpb.Request_Headers)
 		if !ok || headers.Headers == nil {
-			ch.webrtcBaseChannel.logger.Debugf("expected headers as first message but got %T, discard request", req.Type)
+			logger.Debugf("expected headers as first message but got %T, discard request", req.Type)
 			ch.mu.Unlock()
```
edaniels, line 117: this case is largely avoided now

rpc/wrtc_client_stream.go
```diff
@@ -194,8 +208,18 @@ func init() {
...
 
 func (s *webrtcClientStream) writeMessage(m interface{}, eos bool) (err error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.sendClosed {
+		return io.ErrClosedPipe
+	}
+	if eos == true {
+		s.sendClosed = true
```
edaniels, line 218: both reset and this cause send closures preventing some double write errors on a closed stream

rpc/wrtc_client_stream.go
```diff
@@ -130,40 +130,56 @@ func (s *webrtcClientStream) Trailer() metadata.MD {
...
+			// checkWriteErrForRecvClose did not do it for us.
+			s.close()
+		}
 	}()
 	return s.ch.writeReset(s.webrtcBaseStream.stream)
 }
 
 func (s *webrtcClientStream) close() {
 	s.cancel()
+	s.webrtcBaseStream.close()
```
edaniels, line 174: this was mising

rpc/wrtc_client_stream.go
```diff
@@ -130,40 +130,56 @@ func (s *webrtcClientStream) Trailer() metadata.MD {
...
 	}
 	s.webrtcBaseStream.closeRecvWithError(err)
+	return err
 }
 
 // resetStream cancels the stream and sends a reset signal.
-// It is also not safe to call resetStream
-// concurrently with SendMsg or CloseSend.
 func (s *webrtcClientStream) resetStream() (err error) {
+	s.mu.Lock()
```
edaniels, line 152: we now lock in each of these 3 main areas of reset, headers, and message

rpc/wrtc_client_stream.go
```diff
@@ -194,8 +212,27 @@ func init() {
 }
 
 func (s *webrtcClientStream) writeMessage(m interface{}, eos bool) (err error) {
+	s.mu.RLock()
```
edaniels, line 215: `writeMessage` is safe for concurrent use now. for multiple writers non EOS it grabs a read lock. for an EOS, it grabs a write lock

rpc/wrtc_client_stream.go
edaniels: @cheukt seems like no harm having this

rpc/wrtc_base_channel.go
edaniels: This is actually probably wrong. Since a read and write can't be context guarded really in go, we need to rely on the close to kill the operation. Will see if I can work around the errors resulting in the race here

cheukt (approved):
LGTM

PR #170, opened 2023-05-15
limit cipher suits

Kschappacher:
We currently allow for weak ciphers in tls 1.2. This limits the ciphers to only the strong ones.


edaniels (approved):
LGTM, just comment in code with link to maintained list of secure versions and suites

PR #171, opened 2023-05-16
RSDK-3026 Ignore and log user-initiated abort chunk

benjirewis:
[RSDK-3026](https://viam.atlassian.net/browse/RSDK-3026)

Ignores and debug-logs `User Initiated Abort: Close called` errors. Also ignores `dtls.ConnIsClosed` errors that occur when the peer connection has already been closed by the other side.

Certain browsers close RTC peer connections through an ["upper-layer request"](https://www.tech-invite.com/y45/tinv-ietf-rfc-4960-3.html#e-3-3-10-12) that sends an abort chunk error that, [as of relatively recently](https://github.com/pion/sctp/commit/e4e606d1fff223734f7aa74c50eb82d50932b3f4), is correctly packaged by the underlying `sctp` library and handed to goutils. That abort chunk error was being unnecessarily logged by both the data channel and negotiation channel.

[RSDK-3026]: https://viam.atlassian.net/browse/RSDK-3026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rpc/wrtc_base_channel.go
+++ b/rpc/wrtc_base_channel.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	"github.com/edaniels/golog"
+	"github.com/pion/dtls/v2"
 	"github.com/pion/sctp"
 	"github.com/pion/webrtc/v3"
 	"google.golang.org/protobuf/proto"
@@ -144,7 +145,13 @@ func (ch *webrtcBaseChannel) closeWithReason(err error) error {
 	ch.closedReason = err
 	ch.cancel()
 	ch.bufferWriteCond.Broadcast()
-	return ch.peerConn.Close()
+
+	// Underlying connection may already be closed; ignore "conn is closed"
+	// errors.
+	if err := ch.peerConn.Close(); !errors.Is(err, dtls.ErrConnClosed) {
+		return err
+	}
+	return nil
 }
 
 func (ch *webrtcBaseChannel) Close() error {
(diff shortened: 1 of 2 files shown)


rpc/wrtc_base_channel.go
```diff
@@ -174,8 +181,19 @@ func (ch *webrtcBaseChannel) onChannelClose() {
 	}
 }
 
+// isUserInitiatedAbortChunkErr returns true if the error is an abort chunk
+// error that the user initiated through Close. Certain browsers (Safari,
+// Chrome and potentially others) close RTCPeerConnections with this type of
+// abort chunk that is not indicative of an actual state of error.
+func isUserInitiatedAbortChunkErr(err error) bool {
+	return err != nil && errors.Is(err, sctp.ErrChunk) &&
+		strings.Contains(err.Error(), "User Initiated Abort: Close called")
```
edaniels, line 190: for later, it may be nice to export the error in https://github.com/pion/sctp/blob/d6446e3b0d0495d9974e68fec9b7fdb79c018ca1/error_cause_user_initiated_abort.go#LL25C6-L25C34 or to provide a function check if it is such an error

rpc/wrtc_peer.go
```diff
@@ -207,6 +207,8 @@ func newPeerConnectionForServer(
 	if err != nil {
 		return pc, dataChannel, err
 	}
+	defer utils.UncheckedError(negotiationChannel.Close())
```
edaniels, line 210: [major] this should only close on peer/clientconn close. We use this negotiation channel actively throughout the connection

edaniels (changes requested):

rpc/wrtc_base_channel.go
benjirewis: > or to provide a function check if it is such an error

yeah good idea, I'll open a PR to try to add a function. it's not exactly an "error" and more an _error cause_ that gets marshaled and included in the list of error causes for the chunk (and a packet is a slice of chunks), so I don't know if exporting that `errorCauseUserInitiatedAbort` type will help.

rpc/wrtc_peer.go
benjirewis: oh whoops; I thought this channel wasn't needed beyond `newPeerConnectionForServer`. I'm not sure we should `Close` it at all then. the modified `initialDataChannelOnError` function already ignores the "User Initiated Abort: Close called" error, so the "premature" error is still gone even without this deferred close. I could write a new `negotiationChannelOnError` function that can close the negotiation channel on reception of "User Initiated Abort: Close called" but up to you—I'm a little out of my depth

edaniels: yeah that's fine to not close then

edaniels (approved):

benjirewis:
Ticket to file mentioned PR: https://viam.atlassian.net/browse/RSDK-3153.

PR #172, opened 2023-05-16
RSDK-906: Upload test result and coverage data into a MDB backend.

dgottlieb:
The goutils repository can currently output code coverage data into a file that is slurped into a Github PR comment. This patch moves that code into test executaion rather than being contained as a github action. That change will result in an additional coverage file, `code-coverage-results.md` (gitignored) to be created in the goutils repo root when running `make cover`.

### DIFF
--- a/etc/analyzetests/main.go
+++ b/etc/analyzetests/main.go
@@ -1,0 +1,154 @@
+// Package main is a go test analyzer that publishes results to a MongoDB database.
+package main
+
+import (
+	"context"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/edaniels/golog"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/mongo/readpref"
+	"go.uber.org/multierr"
+	"go.viam.com/utils"
+	"gotest.tools/gotestsum/testjson"
+)
+
+var logger = golog.NewDebugLogger("analyzetests")
+
+func main() {
+	utils.ContextualMain(mainWithArgs, logger)
+}
+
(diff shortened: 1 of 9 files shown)


etc/analyzetests/main.go
```diff
@@ -0,0 +1,154 @@
+// Package main is a go test analyzer that publishes results to a MongoDB database.
```
edaniels, line 1: for these tests/coverage commands, I'm assuming no changes were made from the copy in RDK?

etc/test.bash
```diff
@@ -0,0 +1,35 @@
...
+ROOT_DIR="$DIR/../"
+cd $ROOT_DIR
+
+if [[ "$1" == "cover" ]]; then
+	COVER=-coverprofile=coverage.txt
+fi
+
+# race isn't supported on the pi or jetson (and possibly other arm boards)
+# https://github.com/golang/go/issues/29948
+if [ "$(uname -m)" != "aarch64" ] || [ "$(uname)" != "Linux" ]; then
```
edaniels, line 13: we don't test on pi/jetson here so let's remove this and always run with race

etc/test.bash
```diff
@@ -0,0 +1,35 @@
...
+	RACE=-race
+fi
+
+gotestsum --format standard-verbose --jsonfile json.log -- -tags=no_skip $RACE $COVER ./...
```
edaniels, line 17: our scheme of grabbing the pid, trapping, and all that is not needed since this is one command now. We should be able to reduce this to the gotestsum command here followed by the analyzers

edaniels (changes requested):
Basically looks good but we can slim the test.sh some

edaniels (approved):
LGTM

PR #178, opened 2023-05-23
[RSDK-3286] Fix go routine leak in wrtc ice candidate function

Otterverse:
After staring at this code for far too long, I think the only way the leak shown in RSDK-3286 could have happened was by this function returning via the select because of a race with other instances of itself. @edaniels Please take a look though, as I'm not 100% certain how this waitgroup pattern is supposed to function (deferring in a later anonymous function feels odd.)

### DIFF
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -220,7 +220,7 @@ func dialWebRTC(
 			return nil, err
 		}
 
-		var callFlowWG sync.WaitGroup
+		var pendingCandidates sync.WaitGroup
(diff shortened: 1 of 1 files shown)


edaniels (approved):
This is 100% correct. A matching add to a wait group should have a corresponding defer done as close as possible to the goroutine associated with it. Good catch! If you want to make the change, that waitgroup ought to be called pendingCandidates.

Because the test didn't time out, this actually reveals that we ought to have waitgroup or synchronization logic in the teardown of the dial function here. It's not really important though in practice.

PR #179, opened 2023-05-24
RSDK-3045: Fall back to regular dns when mDNS fails.

dgottlieb:


rpc/dial.go
```diff
@@ -109,7 +109,11 @@ func dial(
 
 	if !dOpts.mdnsOptions.Disable && tryLocal && isJustDomain {
 		conn, cached, err := dialMulticastDNS(ctx, address, logger, dOpts)
-		if err != nil || conn != nil {
+		if err != nil {
+			logger.Warnf("Error performing mDNS. Falling back to regular DNS. Err: %v\n", err)
```
edaniels, line 113:
```suggestion
			logger.Warnw("error dialing with mDNS; falling back to other methods", "error", err)
```

- Not sure if linter will catch it but we keep the first letter lower case in logs and errors typically
- We use the `Levelw` method call to do structured JSON logging for easier ingestion of logs elsewhere (GCP Logging for example)

edaniels (approved):
LGTM. Have just a log change I'm requesting.

PR #183, opened 2023-06-08
RSDK-3297: Server streams must hold the stream mutex when closing receivers.

dgottlieb:
A server stream object (and its underlying incoming message channel) is shared by two goroutines. We will name those goroutines the "receiving" goroutine and the "processing" goroutine. Typically, the processing goroutine will be responsible for closing resources when the stream is finished. A stream can finish for (broadly) two reasons:
- A message with the eos (end-of-stream) flag is processed.
- Processing a message results in an error.

Notably in the case where processing a message encounters an error (e.g: session expiration), the code to close the stream can happen concurrently with a new message being received. The stream mutex must be held to reliably wait for any in-flight requests to be processed or canceled.

### DIFF
--- a/rpc/wrtc_base_stream.go
+++ b/rpc/wrtc_base_stream.go
@@ -113,12 +113,14 @@ func (s *webrtcBaseStream) RecvMsg(m interface{}) error {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


edaniels (approved):
LGTM

PR #189, opened 2023-06-22
Data-1289: Add alerting on data apis

gloriacai01:
Adding organization tag and view to grpc metrics in order to be able to add alerting policies for specific orgs.
Ticket: https://viam.atlassian.net/browse/DATA-1289

### DIFF
--- a/perf/grpc.go
+++ b/perf/grpc.go
@@ -11,16 +11,21 @@ import (
 // See: https://opencensus.io/guides/grpc/go/#1
 
 func registerGrpcViews() error {
-	OrgID, _ := tag.NewKey("org-id")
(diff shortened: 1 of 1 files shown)


edaniels (changes requested):
I think you need to do this exact change but in app, not goutils. goutils is for our public, open source code for golang utilities

gloriacai01:
> I think you need to do this exact change but in app, not goutils. goutils is for our public, open source code for golang utilities
@edaniels Do you have a recommendation of where in app I should be making this change?

edaniels:
talked offline to just use @AdamMagaluk's statz lib

PR #191, opened 2023-06-26
APP-2028: Add get last login method

anjinai:
**20230625_APP-2028: Add get last login method**

- Extends the “UserInfo” struct with a new method, GetLastLogin.
- GetLastLogin is needed to help provide last login time of users for display in the members table of the org settings page.

Figma: https://www.figma.com/file/2hs9zGWN9nTLqcm8pR38ip/Management?type=design&node-id=3039-21808

### DIFF
--- a/web/user.go
+++ b/web/user.go
@@ -57,6 +57,8 @@ func (u *UserInfo) GetBool(name string) bool {
 }
 
 // GetLastLogin gets the user last login time.
+// Updated_at is a timestamp indicating when the user's profile was last updated/modified.
+// Changes to last_login are considered updates, so most of the time, updated_at will match last_login.
 func (u *UserInfo) GetLastLogin() string {
 	if u.Properties == nil {
 		return ""
--- a/web/user.go
+++ b/web/user.go
@@ -56,19 +56,17 @@ func (u *UserInfo) GetBool(name string) bool {
 	return b
 }
 
-// GetLastLogin gets the user last login time.
-// Updated_at is a timestamp indicating when the user's profile was last updated/modified.
-// Changes to last_login are considered updates, so most of the time, updated_at will match last_login.
-func (u *UserInfo) GetLastLogin() string {
+// GetUpdatedAt gets the timestamp indicating when the user's auth0 profile was last updated/modified.
+func (u *UserInfo) GetUpdatedAt() string {
 	if u.Properties == nil {
 		return ""
 	}
 
-	lastLogin, ok := u.Properties["updated_at"].(string)
+	updatedAt, ok := u.Properties["updated_at"].(string)
 	if !ok {
 		return ""
 	}
-	return lastLogin
+	return updatedAt
 }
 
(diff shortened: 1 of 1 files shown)


web/user.go
```diff
@@ -56,6 +56,19 @@ func (u *UserInfo) GetBool(name string) bool {
 	return b
 }
 
+// GetLastLogin gets the user last login time.
+func (u *UserInfo) GetLastLogin() string {
+	if u.Properties == nil {
+		return ""
+	}
+
+	ll, ok := u.Properties["updated_at"].(string)
```
ehhong, line 65: can we be verbose and name the variable `lastLogin`?
is there a reason why there's a discrepancy between the name `updated_at` and `last_login`?

anjinai: The reason for the discrepancy between the names 'updated_at' and 'last_login' in the code is because 'last_login' wasn't directly accessible on the auth0 user object. In order to retrieve the timestamp indicating when the user's profile was last updated or modified, I had to use the 'updated_at' field instead. Changes to 'last_login' are considered updates, so 'updated_at' will match 'last_login'. Therefore, I decided to use 'updated_at' to fulfill the purpose of retrieving the user's last login time.

ehhong: will they ever be out of sync? for example, will `updated_at` potentially be more recent than `last_login`?

anjinai: In most cases, 'updated_at' and 'last_login' will match because changes to 'last_login' are considered updates and will update the 'updated_at' timestamp as well. However, there could be scenarios where 'updated_at' might be more recent than 'last_login'. One such scenario could be when the user's profile is modified without triggering a login event. For example, if an administrator updates the user's profile information without the user logging in, 'updated_at' would be updated while 'last_login' remains unchanged.

ehhong: oh okay that does seem unusual. sounds good then, maybe just leave a comment here or link to info when `updated_at` is updated?

anjinai: Good call, thanks - added in a comment explaining 'updated_at.'

web/user.go
```diff
@@ -56,6 +56,21 @@ func (u *UserInfo) GetBool(name string) bool {
 	return b
 }
 
+// GetLastLogin gets the user last login time.
+// Updated_at is a timestamp indicating when the user's profile was last updated/modified.
+// Changes to last_login are considered updates, so most of the time, updated_at will match last_login.
+func (u *UserInfo) GetLastLogin() string {
```
michaellee1019, line 62: I think on the goutils side, I would recommend keeping the names equal. Returning `updated_at` should be called GetUpdatedAt. On the app side, we can name it differently. 

Just following what Eric has told me: Keep goutils as generic as possible and don't put App specific logic in there (such as naming, mappings in this case).

michaellee1019 (changes requested):

web/user.go
anjinai: Thanks, makes sense!  Updated name to GetUpdatedAt.

michaellee1019 (approved):

PR #192, opened 2023-06-27
RSDK-3476 Fix distributed logging

RyanBoring:
This PR fixes an issue in applications that use OpenCensus to maintain traces of spans which are spawned in distinct processes. Without this PR, when a client starts a span and initiates a gRPC request to a server which starts its own span, the server-side span will not mark the client-side span as its parent. This causes the trace to incorrectly report that the server-side span is unrelated to the client-side span.

This PR fixes that issue by adding the current SpanID, TraceID, and TraceOptions to the HTTP headers that are sent in every gRPC request from the client-side. When the server receives those headers, it extracts the span metadata and inserts it into the current context as a new span called `server_root`. Later spans created in the server logic will automatically be made children of `server_root`.

This PR was tested by instrumenting the rpc/examples/echo client and server and exporting the generated trace to GCP. Before the fix, the server-side spans would appear unrelated to the client-side span. After the fix, the server-side spans are correctly displayed as children of the client-side span. This PR was also tested with an uninstrumented client and server, and it behaved as it did before the fix was applied, as desired.

[Here](https://console.cloud.google.com/traces/list?project=av-canary&tid=16253e8832ff2f285e23d4bc6173ac51&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22P14D%22,%22customValue%22:null))) are [two](https://console.cloud.google.com/traces/list?project=av-canary&tid=6a336d2950204b48c0c0d183a8d6d211&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22P14D%22,%22customValue%22:null))) processes that should've had a parent-child relationship
[Here's a link to GCP recording the correct server/client span relationship for the rpc/examples/echo example](https://console.cloud.google.com/traces/list?project=av-canary&tid=b4d0a9918ff1eb24c58bc6930bd0d571&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22P7D%22,%22customValue%22:null)))

### DIFF
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -33,6 +33,10 @@ func StreamClientTracingInterceptor() grpc.StreamClientInterceptor {
 
 func contextWithSpanMetadata(ctx context.Context) context.Context {
 	span := trace.FromContext(ctx)
+	if span == nil {
+		return ctx
+	}
+
 	ctx = metadata.AppendToOutgoingContext(
 		ctx,
 		"trace-id", span.SpanContext().TraceID.String(),
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -9,16 +9,22 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+// UnaryClientTracingInterceptor adds the current Span's metadata to the context.
 func UnaryClientTracingInterceptor() grpc.UnaryClientInterceptor {
-	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	return func(ctx context.Context, method string, req, reply interface{},
+		cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption,
+	) error {
 		ctx = contextWithSpanMetadata(ctx)
 		err := invoker(ctx, method, req, reply, cc, opts...)
 		return err
 	}
 }
 
+// StreamClientTracingInterceptor adds the current Span's metadata to the context.
 func StreamClientTracingInterceptor() grpc.StreamClientInterceptor {
-	return func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	return func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn,
+		method string, streamer grpc.Streamer, opts ...grpc.CallOption,
+	) (grpc.ClientStream, error) {
 		ctx = contextWithSpanMetadata(ctx)
 		stream, err := streamer(ctx, desc, cc, method, opts...)
 		return stream, err
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -1,0 +1,38 @@
+package rpc
+
+import (
+	"context"
+	"fmt"
+
+	"go.opencensus.io/trace"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+func UnaryClientTracingInterceptor() grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		ctx = contextWithSpanMetadata(ctx)
+		err := invoker(ctx, method, req, reply, cc, opts...)
+		return err
+	}
+}
+
+func StreamClientTracingInterceptor() grpc.StreamClientInterceptor {
+	return func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+		ctx = contextWithSpanMetadata(ctx)
+		stream, err := streamer(ctx, desc, cc, method, opts...)
+		return stream, err
+	}
+}
+
+func contextWithSpanMetadata(ctx context.Context) context.Context {
+	span := trace.FromContext(ctx)
+	md, ok := metadata.FromOutgoingContext(ctx)
+	if !ok {
+		md = metadata.New(map[string]string{})
+	}
+	md.Append("trace-id", span.SpanContext().TraceID.String())
+	md.Append("span-id", span.SpanContext().SpanID.String())
+	md.Append("trace-options", fmt.Sprint(span.SpanContext().TraceOptions))
+	return metadata.NewOutgoingContext(ctx, md)
+}
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -33,12 +33,11 @@ func StreamClientTracingInterceptor() grpc.StreamClientInterceptor {
 
 func contextWithSpanMetadata(ctx context.Context) context.Context {
 	span := trace.FromContext(ctx)
-	md, ok := metadata.FromOutgoingContext(ctx)
-	if !ok {
-		md = metadata.New(make(map[string]string))
-	}
-	md.Append("trace-id", span.SpanContext().TraceID.String())
-	md.Append("span-id", span.SpanContext().SpanID.String())
-	md.Append("trace-options", fmt.Sprint(span.SpanContext().TraceOptions))
-	return metadata.NewOutgoingContext(ctx, md)
+	ctx = metadata.AppendToOutgoingContext(
+		ctx,
+		"trace-id", span.SpanContext().TraceID.String(),
+		"span-id", span.SpanContext().SpanID.String(),
+		"trace-options", fmt.Sprint(span.SpanContext().TraceOptions),
+	)
+	return ctx
 }
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -35,7 +35,7 @@ func contextWithSpanMetadata(ctx context.Context) context.Context {
 	span := trace.FromContext(ctx)
 	md, ok := metadata.FromOutgoingContext(ctx)
 	if !ok {
-		md = metadata.New(map[string]string{})
+		md = metadata.New(make(map[string]string))
 	}
 	md.Append("trace-id", span.SpanContext().TraceID.String())
 	md.Append("span-id", span.SpanContext().SpanID.String())
--- a/rpc/server_interceptors.go
+++ b/rpc/server_interceptors.go
@@ -14,6 +14,7 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// UnaryServerTracingInterceptor starts a new Span if Span metadata exists in the context.
 func UnaryServerTracingInterceptor(logger golog.Logger) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 		remoteSpanContext, err := remoteSpanContextFromContext(ctx)
@@ -39,13 +40,14 @@ func UnaryServerTracingInterceptor(logger golog.Logger) grpc.UnaryServerIntercep
 	}
 }
 
+// StreamServerTracingInterceptor starts a new Span if Span metadata exists in the context.
 func StreamServerTracingInterceptor(logger golog.Logger) grpc.StreamServerInterceptor {
 	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
 		remoteSpanContext, err := remoteSpanContextFromContext(stream.Context())
 		if err == nil {
 			newCtx, span := trace.StartSpanWithRemoteParent(stream.Context(), "server_root", remoteSpanContext)
 			defer span.End()
-			stream = WrapServerStream(stream, newCtx)
+			stream = wrapServerStream(newCtx, stream)
 		} else {
 			logger.Warnf("client did not send valid Span metadata in headers, local Spans will not be linked to client. reason: %w", err)
 		}
@@ -64,17 +66,17 @@ func StreamServerTracingInterceptor(logger golog.Logger) grpc.StreamServerInterc
 	}
 }
 
-type ServerStreamWrapper struct {
+type serverStreamWrapper struct {
 	grpc.ServerStream
 	ctx context.Context
 }
 
-func (s *ServerStreamWrapper) Context() context.Context {
+func (s *serverStreamWrapper) Context() context.Context {
 	return s.ctx
 }
 
-func WrapServerStream(stream grpc.ServerStream, ctx context.Context) *ServerStreamWrapper {
-	s := ServerStreamWrapper{ServerStream: stream, ctx: ctx}
+func wrapServerStream(ctx context.Context, stream grpc.ServerStream) *serverStreamWrapper {
+	s := serverStreamWrapper{ServerStream: stream, ctx: ctx}
 	return &s
 }
 
--- a/rpc/server_interceptors.go
+++ b/rpc/server_interceptors.go
@@ -6,16 +6,16 @@ import (
 	"fmt"
 	"strconv"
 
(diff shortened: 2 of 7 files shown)


rpc/wrtc_client_channel.go
```diff
@@ -223,7 +225,16 @@ func (ch *webrtcClientChannel) newClientStream(ctx context.Context, method strin
 }
 
 func makeRequestHeaders(ctx context.Context, method string) *webrtcpb.RequestHeaders {
-	headersMD, _ := metadata.FromOutgoingContext(ctx)
+	headersMD, ok := metadata.FromOutgoingContext(ctx)
+	if !ok {
+		headersMD = metadata.New(map[string]string{})
+	}
+
+	span := trace.FromContext(ctx)
```
edaniels, line 233: I think we may want to support this as a client and server interceptor so that we can use this not just with webrtc but grpc as well. https://gist.github.com/charithe/b5956620fa6bb0caf668f857d70a5fc4 may be useful as well

edaniels (changes requested):

rpc/wrtc_client_channel.go
RyanBoring: Okay, I think the change I made addresses that. I tested the change by adding `rpc.WithUnaryClientInterceptor(rpc.UnaryClientTracingInterceptor())` to the DialOptions on the client side and the equivalent for the server side and it behaved as expected: [GCP Trace](https://console.cloud.google.com/traces/list?project=av-canary&tid=a22b0033d5e85c64f3e4ac10e79d59e5&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22PT6H%22,%22customValue%22:null))&start=1687954823333&end=1687976423333)

RyanBoring: Sure, I'll look into that

rpc/server_interceptors.go
```diff
@@ -0,0 +1,126 @@
...
+// UnaryServerTracingInterceptor starts a new Span if Span metadata exists in the context.
+func UnaryServerTracingInterceptor(logger golog.Logger) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		remoteSpanContext, err := remoteSpanContextFromContext(ctx)
+		var span *trace.Span
+		if err == nil {
+			ctx, span = trace.StartSpanWithRemoteParent(ctx, "server_root", remoteSpanContext)
+			defer span.End()
+		} else {
+			logger.Warnf("client did not send valid Span metadata in headers, local Spans will not be linked to client. reason: %w", err)
```
bazile-clyde, line 26: This should be a key-value pair instead of using `%w`. See https://github.com/viamrobotics/goutils/blob/main/jwks/jwksutils/utils.go#L82

rpc/server_interceptors.go
```diff
@@ -0,0 +1,126 @@
...
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+// UnaryServerTracingInterceptor starts a new Span if Span metadata exists in the context.
+func UnaryServerTracingInterceptor(logger golog.Logger) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		remoteSpanContext, err := remoteSpanContextFromContext(ctx)
+		var span *trace.Span
+		if err == nil {
```
bazile-clyde, line 22: [[style](https://viam.atlassian.net/wiki/spaces/ENG/pages/262371/Viam+Engineering+Guidelines)]: you probably want to do `err != nil` checks and fail-fast when it [reduces nesting](https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting) and limits the [scope of variables](https://github.com/uber-go/guide/blob/master/style.md#reduce-scope-of-variables) here and throughout. Line 30 looks like an obvious exception since it looks it would do the opposite.

rpc/server_interceptors.go
```diff
@@ -0,0 +1,126 @@
...
+		} else {
+			logger.Warnf("client did not send valid Span metadata in headers, local Spans will not be linked to client. reason: %w", err)
+		}
+
+		resp, err := handler(ctx, req)
+		if err == nil {
+			return resp, nil
+		}
+		if _, ok := status.FromError(err); ok {
+			return nil, err
```
bazile-clyde, line 34: Rather than leaving the status info out, which might be helpful, you could wrap the error with it. Something like
```
import "github.com/pkg/errors"
...
		if s, ok := status.FromError(err); ok {
			return nil, errors.Wrapf(err, s.Message())
		}


```

rpc/server_interceptors.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+}
+
+type serverStreamWrapper struct {
+	grpc.ServerStream
+	ctx context.Context
+}
+
+func (s *serverStreamWrapper) Context() context.Context {
```
bazile-clyde, line 74: I think our linter requires all public methods to have documentation. You can check locally by running `make lint-go`. Nevertheless, it's good practice. I'd just copy the docs from the method you're overriding.

rpc/server_interceptors.go
```diff
@@ -0,0 +1,126 @@
...
+	var spanID trace.SpanID
+	copy(spanID[:], spanIDBytes)
+
+	// Extract trace-options
+	traceOptionsMetadata := md.Get("trace-options")
+	if len(traceOptionsMetadata) == 0 {
+		return trace.SpanContext{}, errors.New("trace-options is missing from metadata")
+	}
+
+	traceOptionsUint, err := strconv.ParseUint(traceOptionsMetadata[0], 10, 32)
```
bazile-clyde, line 119: What these values (`10` and `32`) mean, or why they were chosen, was initially unclear to me, and I'd imagine an eng without context would also scratch their head about this. Could this be documented?

rpc/wrtc_client_channel.go
```diff
@@ -224,6 +224,7 @@ func (ch *webrtcClientChannel) newClientStream(ctx context.Context, method strin
 
 func makeRequestHeaders(ctx context.Context, method string) *webrtcpb.RequestHeaders {
 	headersMD, _ := metadata.FromOutgoingContext(ctx)
+
```
bazile-clyde, line 227: Did you mean to leave this from your previous changes?

rpc/wrtc_server_stream.go
```diff
@@ -292,6 +292,15 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
 		return
 	}
 
+	md, ok := metadata.FromIncomingContext(s.ctx)
+	if !ok {
+		md = metadata.New(map[string]string{})
+	}
+	md.Append("trace-id", headers.Metadata.Md["trace-id"].Values[0])
```
bazile-clyde, line 299: It looks this line is causing the test failure in our [GitHub actions](https://github.com/viamrobotics/goutils/actions/runs/5404691236/jobs/9819197882?pr=192#step:8:1158)
```
panic: runtime error: invalid memory address or nil pointer dereference
```
It may be because the `Md` isn't guaranteed to have any of these values.

bazile-clyde (commented):
Looks great so far! I left a few comments and called out the line that's causing the test failures.

rpc/wrtc_client_channel.go
RyanBoring: I've got a few ideas of how I could approach unit testing for this fix but I'm not sure exactly what you were looking for. I could:
1. Test the interceptors in isolation. That is, pass a handler to the interceptor which checks that the interceptor functioned as intended (created a new Span or injected the tracing metadata into the context)
2. Test the interceptors as part of a webrtc/grpc communication. That would mean constructing a whole client and server to facilitate sending and receiving messages and inspecting program behavior along the way, like what you did for [this server test](https://github.com/viamrobotics/goutils/blob/main/rpc/server_test.go), but stripping away authentication and other details. That feels more like an integration test than a unit test, and I don't know if that much boilerplate is warranted for verifying the interceptors' behavior when (1) is also an option, but I will if you think it's appropriate.
3. Another option?
I looked through the tests in the same directory I'm working in and they don't touch on inspecting interceptor behavior because that functionality is usually provided through the grpc library itself. I'm not sure if we ever define our own interceptors like this and test their functionality like I'm trying to, but if you know of a test you can point me to, I can model my tests after that

edaniels: I think option 2 is a great choice! RDK session and opid tests have good examples of testing these with unary and steam interceptors. Message me on slack if you wanna work through more details!

edaniels:
@RyanBoring if you're ready for another review, just click the refresh looking button next to my name in reviewers to request another review!

RyanBoring:
Will do @edaniels, but there's a bug in the test file that causes some requests spawned by grpc.DialContext to not trigger the client interceptor, causing the test to fail. I'm still not sure what could be causing the issue - Clyde took a look through grpc.DialContext and he said he thought that every request it makes would be passed through the client interceptor as they should be, but when he paused execution during debugging, the tests would sometimes fail and sometimes pass. I uploaded a test that highlights the issue [here](https://github.com/RyanBoring/goutils/blob/e0e0366b5860db1da03da835bc50eaa88786bef4/rpc/interceptors_test.go). Do you think you'd be able to take a look into that?

edaniels:
So it looks like it fails due to a data race:
```WARNING: DATA RACE
Write at 0x00c00011e07f by goroutine 19:
  go.viam.com/utils/rpc.TestTracingInterceptors.func5()
      /Users/eric/goutils/rpc/interceptors_test.go:128 +0x48
  go.viam.com/utils/rpc.TestTracingInterceptors()
      /Users/eric/goutils/rpc/interceptors_test.go:159 +0xe8c
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.20.4/libexec/src/testing/testing.go:1576 +0x188
  testing.(*T).Run.func1()
      /opt/homebrew/Cellar/go/1.20.4/libexec/src/testing/testing.go:1629 +0x40

Previous read at 0x00c00011e07f by goroutine 95:
  go.viam.com/utils/rpc.TestTracingInterceptors.func2()
      /Users/eric/goutils/rpc/interceptors_test.go:63 +0xac
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1.1.1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:49 +0x6c
  go.viam.com/utils/rpc.StreamServerTracingInterceptor.func1()
      /Users/eric/goutils/rpc/server_interceptors.go:53 +0x290
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1.1.1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:49 +0x6c
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:58 +0xec
  go.viam.com/utils/rpc.NewServer.func4()
      /Users/eric/goutils/rpc/server.go:333 +0x110
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1.1.1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:49 +0x6c
  go.viam.com/utils/rpc.streamServerCodeInterceptor.func1()
      /Users/eric/goutils/rpc/server.go:864 +0x50
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1.1.1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:49 +0x6c
  github.com/grpc-ecosystem/go-grpc-middleware/logging/zap.StreamServerInterceptor.func1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/logging/zap/server_interceptors.go:53 +0x12c
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1.1.1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:49 +0x6c
  github.com/grpc-ecosystem/go-grpc-middleware/recovery.StreamServerInterceptor.func1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/recovery/interceptors.go:51 +0xbc
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1.1.1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:49 +0x6c
  github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1()
      /Users/eric/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:58 +0xec
  google.golang.org/grpc.(*Server).processStreamingRPC()
      /Users/eric/go/pkg/mod/google.golang.org/grpc@v1.54.0/server.go:1646 +0x17e8
  google.golang.org/grpc.(*Server).handleStream()
      /Users/eric/go/pkg/mod/google.golang.org/grpc@v1.54.0/server.go:1726 +0xbc8
  google.golang.org/grpc.(*Server).serveStreams.func1.2()
      /Users/eric/go/pkg/mod/google.golang.org/grpc@v1.54.0/server.go:966 +0xa4```

That explains the inconsistency you're seeing. Let me know if you want to debug this further in person. I ran `go test -race -run TestTracingInterceptors -count=3` to get this. You may need to increase the count.

**Comment by edaniels** [comment]
@bazile-clyde should the interceptors always be installed or should a user need to add them themselves? I feel like it would be nice if they were there automatically

File: rpc/server_interceptors.go [hunk 9]
```diff
@@ -0,0 +1,125 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+// UnaryServerTracingInterceptor starts a new Span if Span metadata exists in the context.
+func UnaryServerTracingInterceptor(logger golog.Logger) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if remoteSpanContext, err := remoteSpanContextFromContext(ctx); err != nil {
+			logger.Warnf("client did not send valid Span metadata in headers, local Spans will not be linked to client", "error", err)
```
edaniels, line 21: needs to be warnw as discussed for both but I also think we may not need to log this at all since not all clients will always send spans

edaniels: no worries!

bazile-clyde:
> @bazile-clyde should the interceptors always be installed or should a user need to add them themselves? I feel like it would be nice if they were there automatically

@edaniels I think having them there automatically is the way to go. 

@RyanBoring That's how it's currently implemented right? If there's a parent span, then we automatically use it. Otherwise, we operate as normal without tracing data.

edaniels:
Okay cool. Let's get them in as standalone here and then have a follow up PR to do it automatically (with an option to enable). better to opt into this I think?

rpc/client_interceptors.go
```diff
@@ -0,0 +1,43 @@
...
+		stream, err := streamer(ctx, desc, cc, method, opts...)
+		return stream, err
+	}
+}
+
+func contextWithSpanMetadata(ctx context.Context) context.Context {
+	span := trace.FromContext(ctx)
+	ctx = metadata.AppendToOutgoingContext(
+		ctx,
+		"trace-id", span.SpanContext().TraceID.String(),
```
edaniels, line 38: [minor] we'll need to make sure we don't append empty headers. so let's check we have an actual span

edaniels:
> Okay cool. Let's get them in as standalone here and then have a follow up PR to do it automatically (with an option to enable). better to opt into this I think?

nvm fine to do in this PR and not make it an option

edaniels (changes requested):
Just one minor comment on when we do the attaching

edaniels (approved):

PR #194, opened 2023-07-18
App 2076: Creating public methods for a server

RoxyFarhad:
For Modular Resource Registry, we want some endpoints to not require auth checks but still be valid. This PR allows for that by allowing Servers to take a list of grpc method definitions, and bypassing auth if that method is in the list on server instantiation.


rpc/js/package-lock.json
RoxyFarhad, line 1: not sure why this is in here, i think there is an issue with my linter will fix.

rpc/server_auth_test.go
```diff
@@ -525,6 +525,15 @@ func TestServerAuthJWTAudienceAndID(t *testing.T) {
 	}
 }
 
+func TestServerPublicMethods(t *testing.T) {
```
RoxyFarhad, line 528: TODO: just wanted a review of the code

michaellee1019 (commented):
LGTM so far. My only thought was to rename and refactor `ensureAuthed` into both `ensureAuthed` and `tryAuth`. try would attempt to authenticate and return nothing. ensure would require authentication and would only be called when `canByPassAuthCheck` returns false. However that is a lot of refactoring just to avoid a status code check.

zaporter-work (changes requested):
edit: removing in favor of next review

rpc/server.go
```diff
@@ -375,6 +379,12 @@ func NewServer(logger golog.Logger, opts ...ServerOption) (Server, error) {
 		server.exemptMethods[healthWatchMethod] = true
 	}
 
+	if len(sOpts.publicMethods) > 0 {
+		for _, method := range sOpts.publicMethods {
+			server.publicMethods[method] = true
+		}
+	}
```
zaporter-work, line 386:
```suggestion

		for _, method := range sOpts.publicMethods {
			server.publicMethods[method] = true
		}

```

rpc/server_auth.go
```diff
@@ -162,6 +162,12 @@ func (ss *simpleServer) signAccessTokenForEntity(
 	return tokenString, nil
 }
 
+func (ss *simpleServer) canByPassAuthCheck(
```
zaporter-work, line 165:
```suggestion
func (ss *simpleServer) isPublicMethod(
```
[opt]
if we are going to keep this function, `canBypassAuthCheck` seems like a complex name for this simple function

rpc/server_auth.go
```diff
@@ -171,6 +177,9 @@ func (ss *simpleServer) authUnaryInterceptor(
 	if !ss.exemptMethods[info.FullMethod] {
 		nextCtx, err := ss.ensureAuthed(ctx)
 		if err != nil {
+			if status, _ := status.FromError(err); status.Code() == codes.Unauthenticated && !ss.canByPassAuthCheck(info.FullMethod) {
```
zaporter-work, line 180:
```suggestion
			if status, _ := status.FromError(err); status.Code() == codes.Unauthenticated && ss.canByPassAuthCheck(info.FullMethod) {
```

Shouldn't this be run if we `can` bypass auth checks rather than if we cant bypass auth checks

rpc/server_auth.go
```diff
@@ -187,6 +196,9 @@ func (ss *simpleServer) authStreamInterceptor(
 	if !ss.exemptMethods[info.FullMethod] {
 		nextCtx, err := ss.ensureAuthed(serverStream.Context())
 		if err != nil {
+			if status, _ := status.FromError(err); status.Code() == codes.Unauthenticated && !ss.canByPassAuthCheck(info.FullMethod) {
```
zaporter-work, line 199: Same comment

rpc/server_options.go
```diff
@@ -33,6 +33,12 @@ type serverOptions struct {
 	// allowUnauthenticatedHealthCheck allows the server to have an unauthenticated healthcheck endpoint
 	allowUnauthenticatedHealthCheck bool
 
+	// allowPublicMethods allows for nested public API methods on the server
+	// i.e. will disable auth checks for users
+	allowPublicMethods bool
```
zaporter-work, line 38: This is never checked anywhere

michaellee1019:
> Then we don't have to have this complex distinction between a `publicMethod` and an `exemptMethod` which are almost identical but not really

I see value in having 3 distinct mechanisms:
1. Authentication is required (api is in neither list): All requests are authenticated and result in Unauthenticated error code when not. Needed for the majority of pages that require login
2. Authentication is optional (api is in `publicMethod`): If the user is authenticated then retrieve the user and show user information. Otherwise allow the API to still be called but without the context of a user. The API and pages that use the API have to be designed to retrieve data based on both cases.
3. Authentication is not done (api is in `exemptMethod`: We don't plan on ever authenticating this route, because no user information is needed. Say a homepage that is not personalized, or a page that is designed for SEO bots.

Having 3 fit into number 2 means that we will grab user information even when we don't need it, even though the underlying API doesn't need this information. It could result in extra DB calls and other work that isn't required. So I think that is the only strong case for having 3 options.

zaporter-work:
Mhmm -- okay -- I agree on the unnecessary db calls to do the auth.

In that case, how do we feel about `publicMethod` -> `opportunisticallyAuthenticated` (or something similar that indicates that auth is actually done but it is not required.)?

RoxyFarhad:
@zaporter-work +1 to what @michaellee1019 Said about keeping them separate. While exempt methods are a sub-part of public methods, they are different in that we shouldn't be trying to do the auth work required for the non-exempt endpoints for the exempt endpoints. It requires doing a lot of complex auth work which for health checks (major exempt method), should not be being called and just return the critical path. 

I agree they definitely are similar just because of how they store data and where they touch in the codebase but what they avoid is different. 

I do think that `public` is succinct and clear to show that it is not a private API which is what all our APP endpoints are by default. I can leave a comment to explain more clearly that this means `optional_auth` essentially but if you have a strong feeling about this can discuss

rpc/server_options.go
RoxyFarhad: sorry yeah, something i need to remove. I had it in there originally but the map is enough.

rpc/server_auth.go
RoxyFarhad: my bad! I originally made ss. canByPassAuthCheck ss.requiresAuthCheck so forgot to remove the negation

rpc/server.go
```diff
@@ -296,9 +299,10 @@ func NewServer(logger golog.Logger, opts ...ServerOption) (Server, error) {
 			info *grpc.UnaryServerInfo,
 			handler grpc.UnaryHandler,
 		) (interface{}, error) {
-			if server.exemptMethods[info.FullMethod] {
+			if server.exemptMethods[info.FullMethod] || server.canByPassAuthCheck(info.FullMethod) {
```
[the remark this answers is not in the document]
RoxyFarhad: i actually think it is better to have a function in this situation. I think we should avoid these map lookups in line just so its clear when someone is looking at the code what the lookup is actually doing.

rpc/server_options.go
RoxyFarhad: hm weird actually, my commit didn't push up with this removed

RoxyFarhad:
>  My only thought was to rename and refactor ensureAuthed into both ensureAuthed and tryAuth. try would attempt to authenticate and return nothing. ensure would require authentication and would only be called when canByPassAuthCheck returns false. However that is a lot of refactoring just to avoid a status code check.

i agree with this ^ but for time yeah I think it would be a lot of refactoring also considering to separate it out completely. Can add a helper function to wrap around enureAuthed to catch the status error but yeah am hesitant to go directly into that right now.

rpc/server.go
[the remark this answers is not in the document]
RoxyFarhad: thinking about this more, we should leave out the unaryInterceptor check in here. It should be up to the server implementation to know whether the endpoint should avoid middleware.

rpc/server_auth.go
```diff
@@ -236,6 +272,17 @@ var validSigningMethods = []string{
 	"RS256",
 }
 
+// tryAuth is called for public methods where auth is not required but preferable.
+func (ss *simpleServer) tryAuth(ctx context.Context) (context.Context, error) {
```
RoxyFarhad, line 276: @michaellee1019 what do you think of this?

rpc/server_auth.go
RoxyFarhad: changed to IsPublicMethod

michaellee1019:
> hesitant to go directly into that right now.

Yeah some future task is OK with me. Not entirely convinced the refactor would result in more clarity and might be more confusing.

rpc/server_auth.go
michaellee1019: much better! It adds a ton of clarity.

michaellee1019 (approved):
LGTM!

rpc/server.go
```diff
@@ -119,10 +119,12 @@ type simpleServer struct {
 	signalingServer         *WebRTCSignalingServer
 	mdnsServers             []*zeroconf.Server
 	exemptMethods           map[string]bool
```
zaporter-work, line 121:
```suggestion
        // exempt methods do not perform authentication
	exemptMethods           map[string]bool
```

rpc/server.go
```diff
@@ -119,10 +119,12 @@ type simpleServer struct {
...
-	tlsConfig               *tls.Config
-	firstSeenTLSCertLeaf    *x509.Certificate
-	stopped                 bool
-	logger                  golog.Logger
+	// public methods is a list of methods that have optional auth
+	publicMethods        map[string]bool
```
zaporter-work, line 123:
```suggestion
	// public methods attempt, but do not require, authentication
	publicMethods        map[string]bool
```

rpc/context.go
```diff
@@ -66,6 +66,10 @@ func ContextWithAuthEntity(ctx context.Context, authEntity EntityInfo) context.C
 
 // ContextAuthEntity returns the entity (e.g. a user) associated with this authentication context.
 func ContextAuthEntity(ctx context.Context) (EntityInfo, bool) {
+	authEntityValue := ctx.Value(ctxKeyAuthEntity)
+	if authEntityValue == nil {
+		return EntityInfo{}, false
+	}
```
zaporter-work, line 72: [q] What is this for?

zaporter-work (approved):
LGTM!

rpc/server_options.go
```diff
@@ -33,6 +33,9 @@ type serverOptions struct {
 	// allowUnauthenticatedHealthCheck allows the server to have an unauthenticated healthcheck endpoint
 	allowUnauthenticatedHealthCheck bool
 
+	// publicMethods is a list of publicMethods that are allowed to bypass auth for
```
zaporter-work, line 36:
```suggestion
	// publicMethods are api routes that attempt, but do not require, authentication
```

[nit] typo in original comment

PR #195, opened 2023-08-14
App 997: Pass an update robot part function to the call queue

RoxyFarhad:
This is required for this work: https://github.com/viamrobotics/app/pull/2175

Went through a lot of iterations for this:
1. using the signaling answer functionality but that only indicated when the connection was created + dropped
2. Creating a change stream for the operator table (complex set up)

This definitely made the most sense, and was the most concise change


rpc/wrtc_call_queue_mongodb.go
```diff
@@ -87,6 +87,8 @@ type mongoDBWebRTCCallQueue struct {
 	csStateUpdates              chan changeStreamStateUpdate
 	csCtxCancel                 func()
 
+	// function to update access times on robot parts based on this call queue
+	updateRobotPartConnectivity *func(hostnames []string)
```
michaellee1019, line 91: Lets not name it app specific. 
```suggestion
	activeAnswerersfunc *func(hostnames []string)
```

michaellee1019 (approved):
Just a naming concern and the rest LGTM!

rpc/wrtc_call_queue_mongodb_test.go
```diff
@@ -14,14 +14,33 @@ import (
 	"go.viam.com/utils/testutils"
 )
 
+var activeAnswerersTrigger chan int
+
+func init() {
+	activeAnswerersTrigger = make(chan int, 10)
+}
+
+func activeAnswerersStub(hostnames []string) {
```
RoxyFarhad, line 23: @michaellee1019 this was my way to test that passing in a function would be called when running the call_queue. Not sure if a channel was the best way of going here but essentially I just wanted something that would be able to be pinged inside the operator goroutine.

rpc/wrtc_call_queue_mongodb_test.go
```diff
@@ -153,4 +155,31 @@ func TestMongoDBWebRTCCallQueueMulti(t *testing.T) {
 		t.Logf("received offer %d=%s", (maxHostAnswerersSize*2)+1, exchange.UUID())
 		exchanges[0] = exchange
 	})
+
+	t.Run("ActiveAnswerer", func(t *testing.T) {
```
RoxyFarhad, line 159: @michaellee1019 this is the updated test that creates a go routine with a single answerer

michaellee1019: Nice! good job figuring this out. I like the simple length test.

michaellee1019 (approved):

PR #196, opened 2023-08-28
RSDK-4281: Shutting down viam-server causes intel realsense module to seg-fault

bazile-clyde:
I've manually tested that this fixes the errors referenced in the ticket. I've also confirmed that the module returns an exit code of 0 and throws no exceptions when this happens with something like
```
int main(int argc, char* argv[]) try {
    return ... // always zero
} catch (const std::exception& e) {
         // no exceptions caught
}

```

This error message should also be ignored since the module does not provide an exit code, exception, or any other status information. The problem is likely at the boundary between the module and the C++ SDK miscommunicating what it means to "[shut] down gracefully" on [close](https://github.com/viamrobotics/viam-cpp-sdk/blob/816344a7932260307fbcfe0255a0c08afeb448f6/src/viam/sdk/module/service.cpp#L223). 

Debugging the C++ SDK along with process signals is way out of the scope of this ticket. I'll create a new ticket and link it in the TODO here if this gets approved, but I think this represents a larger issue of how we shut down processes.

### DIFF
--- a/pexec/managed_process_unix.go
+++ b/pexec/managed_process_unix.go
@@ -106,12 +106,14 @@ func (p *managedProcess) kill() (bool, error) {
 	return forceKilled, nil
 }
 
-func isWaitErrUnknown(err string, forceKilled bool) bool {
-	// This can easily happen if the process does not handle interrupts gracefully
-	// and it won't provide us any exit code info.
+func isWaitErrUnknown(err string, _ bool) bool {
+	// TODO: ensure processes handle interrupts gracefully.
+	//  For now, ignore signals caused by improper handling of interrupts since they provide no exit code, exception, or
+	//  other status information.
 	switch err {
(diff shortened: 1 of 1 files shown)


Otterverse (changes requested):
So I personally don't think this is a good idea. The module requirement list here https://pkg.go.dev/go.viam.com/rdk/module#hdr-Module_Protocol_Requirements explicitly states `[the module] Cleanly exits when sent a SIGINT or SIGTERM signal.`

Additionally, a segfault is very definitively an error, not just an "unknown status" like the other codes here. If a module segfaults on close, something is wrong with the module, not our process management here in goutils. I firmly believe we should fix errors, not ignore them. Doubly so because this isn't just one special case for the camera, but is going to ignore all segfaults, and that could be a critical error for other modules people use, one they DO need to know is happening.

PR #198, opened 2023-08-29
APP-2534: Create a TTL index on the web.sessions collection

Oabraham1:
This is required for: [Create TTL index on web.sessions in app.viam](https://viam.atlassian.net/browse/APP-2535)

In this PR, I added code to ensure that a TTL index is created on the `web.sessions` collections as this helps me close out work on the `app.viam` on creating a TTL index that removes sessions documents that are over 30 days old.

### DIFF
--- a/web/sessions.go
+++ b/web/sessions.go
@@ -4,16 +4,28 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
-	"errors"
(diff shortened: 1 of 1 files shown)


RoxyFarhad:
lgtm! the only thing you need to do to fix the test error is just run `make build lint` and then commit the file changes :)

PR #203, opened 2023-09-19
APP-2592:  Goutils: implement auth handler that supports (id, key) pairs

Oabraham1:
[Ticket](https://viam.atlassian.net/browse/APP-2592)

## Description:
The goal of this PR is to add the `MakeSimpleMultiAuthPairHandler` which implements an auth handler for an api key and api key id pair. The goal is described more in-depth [here](https://docs.google.com/document/d/1o1Eq0bQBun4LADDaOWGjUTI9dfOnI4wenTe6JtChe3Y/edit#bookmark=id.ff76wzf60oxk)

## Testing:
- [x] Added unit test to test proper software behavior.

### DIFF
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -200,6 +200,40 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
 	})
 }
 
+// SimpleMultiAuthPairHandler works similarly to `MakeSimpleMultiAuthHandler` with the addtion of
+// supporting a key, id pair used to ensure that a key that maps to the 
+func MakeSimpleMultiAuthPairHandler(forEntities, expectedPayloads []string) AuthHandler {
+	if len(expectedPayloads) == 0 {
+		panic("expected at least one payload")
+	}
+	entityChecker := MakeEntitiesChecker(forEntities)
+	return AuthHandlerFunc(func(ctx context.Context, entity, payload string) (map[string]string, error) {
+		if err := entityChecker(ctx, entity); err != nil {
+			if errors.Is(err, errCannotAuthEntity) {
+				return nil, errInvalidCredentials
+			}
+			return nil, err
+		}
+
+		payloadB := []byte(payload)
+		payloadKeyB := []byte(payload)
+		for _, expectedPayload := range expectedPayloads {
+			if subtle.ConstantTimeCompare(payloadB, []byte(expectedPayload)) == 1 {
+				// check that the key it maps to is the same as the key that was passed in with the request
+				if subtle.ConstantTimeCompare(payloadKeyB, []byte("")) == 1 {
+					return map[string]string{}, nil
+				}
+			}
+		}
+		return nil, errInvalidCredentials
+	})
+	// We are given a keyID and a key by the user
+	// Perform a lookup of the keyID in the list of pre-populated keyIDs
+	// If the keyID is found, lookup the corresponding key to the ID found in the list of keyIDs
+	// Compare that that ID is the same as the ID passed in the request
+	// Done
+}
+
 // MakeEntitiesChecker checks a list of entities against a given one for use in an auth handler.
 func MakeEntitiesChecker(forEntities []string) func(ctx context.Context, entities ...string) error {
 	return func(ctx context.Context, entities ...string) error {
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -203,52 +203,19 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
 // MakeSimpleMultiAuthPairHandler works similarly to MakeSimpleMultiAuthHandler with the addition of
 // supporting a key, id pair used to ensure that a key that maps to the id matches the key passed
 // during the function call.
-func MakeSimpleMultiAuthPairHandler(forEntities, expectedPayloads []map[string]string) AuthHandler {
+
+func MakeSimpleMultiAuthPairHandler(expectedPayloads map[string]string) AuthHandler {
 	if len(expectedPayloads) == 0 {
 		panic("expected at least one payload")
 	}
 
-	// Create a list of KeyIDs for the entities from our given map since MakeEntitiesChecker takes in a list of entities
-	var forEntitiesList []string
-	for _, entity := range forEntities {
-		for key := range entity {
-			forEntitiesList = append(forEntitiesList, key)
-		}
-	}
-
-	entityChecker := MakeEntitiesChecker(forEntitiesList)
 	return AuthHandlerFunc(func(ctx context.Context, entity, payload string) (map[string]string, error) {
-		if err := entityChecker(ctx, entity); err != nil {
-			if errors.Is(err, errCannotAuthEntity) {
-				return nil, errInvalidCredentials
-			}
-			return nil, err
+		if _, ok := expectedPayloads[entity]; !ok {
+			return nil, errInvalidCredentials
 		}
 
-		// set the keyID to the current payload
-		payloadKeyID := []byte(payload)
-
-		// Find the key that maps to the current payload
-		var payloadKey string
-		for _, pk := range expectedPayloads {
-			for key, value := range pk {
-				if subtle.ConstantTimeCompare([]byte(key), payloadKeyID) == 1 {
-					payloadKey = value
-				}
-			}
-		}
-
-		for _, expectedPayloadMap := range expectedPayloads {
-			for key, value := range expectedPayloadMap {
-				expectedPayloadKeyID := key
-				expectedPayloadKey := value
-				if subtle.ConstantTimeCompare(payloadKeyID, []byte(expectedPayloadKeyID)) == 1 {
-					// check that the key it maps to is the same as the key that was passed in with the request
-					if subtle.ConstantTimeCompare([]byte(payloadKey), []byte(expectedPayloadKey)) == 1 {
-						return map[string]string{}, nil
-					}
-				}
-			}
+		if subtle.ConstantTimeCompare([]byte(expectedPayloads[entity]), []byte(payload)) == 1 {
+			return map[string]string{}, nil
 		}
 		return nil, errInvalidCredentials
 	})
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -225,8 +225,10 @@ func MakeSimpleMultiAuthPairHandler(forEntities, expectedPayloads []map[string]s
 			return nil, err
 		}
 
-		// Find the key that maps to the current payload which is a keyID
+		// set the keyID to the current payload
 		payloadKeyID := []byte(payload)
+
+		// Find the key that maps to the current payload
 		var payloadKey string
 		for _, pk := range expectedPayloads {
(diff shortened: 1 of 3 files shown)


rpc/auth.go
```diff
@@ -200,6 +200,60 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
 	})
 }
 
+// MakeSimpleMultiAuthPairHandler works similarly to MakeSimpleMultiAuthHandler with the addition of
+// supporting a key, id pair used to ensure that a key that maps to the id matches the key passed
+// during the function call.
+func MakeSimpleMultiAuthPairHandler(forEntities, expectedPayloads []map[string]string) AuthHandler {
```
jr22, line 206: could you explain a bit to me about how we're going to be calling this? will `forEntities` be a list of robot FQDNs? and then the payload will be a map {keyId: key}? want to make sure i'm fully understanding the use case.

the reason why I ask is, for all the sdks, it seems like we're passing in they api key ID as the "entity" so am wondering how that would work with this structure.

jr22, line 206: (sorry left a comment here originally and then deleted it while i was looking at the pr a bit more but re-adding something in the similar vein) 

can you explain a bit about how you are planning on calling this? for example, what is an expected values for the `forEntities` map and the `expectedPayloads` map? seems like what we would want to accept multiple keyid-key pairs (where the keyid is the entity and the key is the payload) in only one map?  

once i have clarity on this I feel like itll help me comment more on the rest

rpc/auth.go
```diff
@@ -200,6 +200,60 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
...
+	if len(expectedPayloads) == 0 {
+		panic("expected at least one payload")
+	}
```
ohEmily, line 209: [important] I believe this line is in the original `MakeSimpleMultiAuthHandler` above because before this PR, simple auth was the only form of auth, i.e. if there were nothing loaded into `MakeSimpleMultiAuthHandler`, there would be no way to control the robot, so it made sense to have the process panic. Now that we are having 2 different kinds of auth, I think this `panic` should only confirm if there is neither an entry in `MakeSimpleMultiAuthHandler` NOR in `MakeSimpleMultiAuthPairHandler`.

rpc/auth.go
```diff
@@ -200,6 +200,60 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
...
+	// Create a list of KeyIDs for the entities from our given map since MakeEntitiesChecker takes in a list of entities
+	var forEntitiesList []string
+	for _, entity := range forEntities {
+		for key := range entity {
+			forEntitiesList = append(forEntitiesList, key)
+		}
+	}
+
+	entityChecker := MakeEntitiesChecker(forEntitiesList)
+	return AuthHandlerFunc(func(ctx context.Context, entity, payload string) (map[string]string, error) {
```
ohEmily, line 220: [q] what is the advantage of having this as an inline function?

ohEmily: Ah I see this is done above. Uh. Ok I have no idea why this is; seems to obfuscate the code.

rpc/auth.go
```diff
@@ -200,6 +200,60 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
...
+				if subtle.ConstantTimeCompare(payloadKeyID, []byte(expectedPayloadKeyID)) == 1 {
+					// check that the key it maps to is the same as the key that was passed in with the request
+					if subtle.ConstantTimeCompare([]byte(payloadKey), []byte(expectedPayloadKey)) == 1 {
+						return map[string]string{}, nil
+					}
+				}
+			}
+		}
+		return nil, errInvalidCredentials
+	})
```
ohEmily, line 254: [q] what's the advantage of calling `subtle.ConstantTimeCompare` on the key versus just indexing into the map with the string you already have? and then calling `subtle.ConstantTimeCompare` on the key you got back and comparing it to the value mapped to the key you started with? Don't fully understand this algorithm.

rpc/auth.go
jr22: aren't we only ever going to call one or the other though? my understanding was that each credential type only had one auth handler so we wouldnt be calling both. example from rdk: https://github.com/jr22/rdk/blob/55cab43ce0a2f5bf08caf929761afc5e750a6650/robot/web/web.go#L1042-L1044

or are you saying that the changes in rdk will be such that the api-key credential type will have a new type of auth handler that calls both?

rpc/auth.go
Oabraham1: forEntities would contain a bucket of (keyID, key) mappings that we are basically comparing the `expectedPayloads` against to 1) ensure that the expectedPayload(keyID, key mapping) exists in bucket 2) ensure that the keyID found has a key that matches the key passed into the expectedPayload. The payload is a keyId and key pair and our goal here is to ensure that that keyID exists in a bucket of keyIDs that we're searching through and when we find that keyID, we want to verify that the key passed in the payload is the same as the one stored in the bucket. Not sure if this helps, more than happy to discuss in person.

jr22: so expectedEntities being a map of keyID to key makes sense to me. but expectedPayloads will be passed in when this auth handler is initialized right? so we are comparing the arguments of `AuthHandlerFunc` to `forEntities` and `expectedPayloads`? 

basically, what i'm getting at, is why do there have to be two maps of keyId to key? in `AuthHandlerFunc` the entity should be a keyId and `payload` should be the key itself, so seems like one map would suffice here?

jr22: also can discuss monday in person as it may be easier to explain then

rpc/auth.go
jr22: `MakeSimpleMultiAuthPairHandler` is basically defining this as the callback function right?

rpc/auth.go
Oabraham1: +1 to Julia's question. I assumed that we would be calling only one AuthHandler at a time either with a string or map of keyID and key?

rpc/auth.go
Oabraham1: @ohEmily no real advantage that I see here. Was following the convention in the `MakeSimpleMultiAuthHandler` that is similar to `MakeSimpleMultiAuthPairHandler`

rpc/auth.go
ohEmily: I think that as we are migrating over, it's possible to have both. We aren't going to delete secrets just because we added API keys. But we should eventually support removing the secrets.

jr22: oh right because we need to support those legacy robot API keys and i guess there is no way to know when setting up the auth handler which one its gonna be

rpc/auth.go
```diff
@@ -200,6 +200,60 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
...
+			return nil, err
+		}
+
+		// set the keyID to the current payload
+		payloadKeyID := []byte(payload)
+
+		// Find the key that maps to the current payload
+		var payloadKey string
+		for _, pk := range expectedPayloads {
+			for key, value := range pk {
```
piokasar, line 234: I think this loop could be combined with the one below.

rpc/auth_test.go
```diff
@@ -137,6 +137,53 @@ func TestMakeSimpleMultiAuthHandler(t *testing.T) {
...
+	t.Run("with no entities should always fail", func(t *testing.T) {
+		handler := MakeSimpleMultiAuthPairHandler(nil, []map[string]string{{"something": "nothing"}})
+		_, err := handler.Authenticate(context.Background(), "", "something")
+		test.That(t, err, test.ShouldNotBeNil)
+		_, err = handler.Authenticate(context.Background(), "entity", "something")
+		test.That(t, err, test.ShouldNotBeNil)
+	})
+
+	t.Run("should validate entities and (keyID, key) mappings", func(t *testing.T) {
+		expectedEntitiesMap := []map[string]string{{"one": "1"}, {"two": "2"}, {"three": "3"}}
```
piokasar, line 154: This might not need to be a map as it looks like you're discarding the values anyways.

rpc/auth.go
```diff
@@ -200,6 +200,60 @@ func MakeSimpleMultiAuthHandler(forEntities, expectedPayloads []string) AuthHand
...
+			}
+		}
+
+		for _, expectedPayloadMap := range expectedPayloads {
+			for key, value := range expectedPayloadMap {
+				expectedPayloadKeyID := key
+				expectedPayloadKey := value
+				if subtle.ConstantTimeCompare(payloadKeyID, []byte(expectedPayloadKeyID)) == 1 {
+					// check that the key it maps to is the same as the key that was passed in with the request
+					if subtle.ConstantTimeCompare([]byte(payloadKey), []byte(expectedPayloadKey)) == 1 {
```
piokasar, line 247: In this check, I think the second "if" will always be true because you're using the expected payload value for the corresponding key. I think the issue may be that the AuthHandlerFunc is only taking a string as the payload, rather than a pair that would represent the api key id and the api key. Let me know if I'm completely misunderstanding though and also happy to talk about it in person.

rpc/auth.go
Oabraham1: So instead of the first `subtle.ConstantTimeCompare`, we just check that the values that `payloadKeyID` and `expectedPayloadKeyID` map to are the same? That makes perfect sense and simplifies that better for me :)

Oabraham1:
@ohEmily I had a pair programming session with @jr22 and @piokasar to help me better understand the code path and resolve a lot of comments.

rpc/auth.go
ohEmily: I think I might have missed an in-person resolution to this commit. Did you all establish that it was OK to have the `panic` in both auth handlers?

Oabraham1: Yes, during the meeting we talked about how we would include branching logic in the rdk code path to use one of the handlers depending on the credential type in the setup hence the panic would confirm the absence of a payload in `MakeSimpleMultiAuthPairHandler` if we provide an empty map. In the case of branching logic, if our `handler.Config` [here](https://github.com/viamrobotics/rdk/blob/55cab43ce0a2f5bf08caf929761afc5e750a6650/robot/web/web.go#L1033-L1059) is called with the old api keys the we call `MakeSimpleMultiAuthHandler`, if it is called with the new api keys then we call, `MakeSimpleMultiAuthPairHandler` if it called with both then we call with some hybrid function that calls both in the background?

ohEmily: kk thank you!

rpc/auth.go
ohEmily: ah fwiw I figured this out. It's because the return type of the method is actually a function; we are returning the callback rather than calling the function.

jr22: ^ thats what i was trying to get at above. shouldve explained that better

ohEmily (approved):
Thank you!

rpc/auth.go
Oabraham1: I should've also explained it better

PR #204, opened 2023-09-29
Add struct to structpb method that ignores omit empty tags

vijayvuyyuru:
Hey yall, as part of queryable sensor data , we would like users to be able to see all fields of their data regardless of it is an "empty value". If a boolean is false, it can be dropped from the collector which is not ideal. The testing for the new method is fairly barebones as its fundamentally the same method, just added a case where it shouldn't be dropping a variable. I'm happy to add more if required.

### DIFF
--- a/protoutils/protoutils.go
+++ b/protoutils/protoutils.go
@@ -16,7 +16,7 @@ func InterfaceToMap(data interface{}) (map[string]interface{}, error) {
 	return interfaceToMapHelper(data, true)
 }
 
-func interfaceToMapHelper(data interface{}, shouldOmitEmpty bool) (map[string]interface{}, error) {
+func interfaceToMapHelper(data interface{}, ignoreOmitEmpty bool) (map[string]interface{}, error) {
 	if data == nil {
 		return nil, errors.New("no data passed in")
 	}
@@ -28,12 +28,12 @@ func interfaceToMapHelper(data interface{}, shouldOmitEmpty bool) (map[string]in
 	var err error
 	switch t.Kind() {
 	case reflect.Struct:
-		res, err = structToMap(data, shouldOmitEmpty)
+		res, err = structToMap(data, ignoreOmitEmpty)
 		if err != nil {
 			return nil, err
 		}
 	case reflect.Map:
-		res, err = marshalMap(data, shouldOmitEmpty)
+		res, err = marshalMap(data, ignoreOmitEmpty)
 		if err != nil {
 			return nil, err
 		}
@@ -68,8 +68,8 @@ func StructToStructPb(i interface{}) (*structpb.Struct, error) {
 func StructToStructPbIgnoreOmitEmpty(i interface{}) (*structpb.Struct, error) {
 	encoded, err := interfaceToMapHelper(i, false)
 	if err != nil {
-		return nil, errors.Wrap(err,
-			fmt.Sprintf("unable to convert interface %v to a form acceptable to structpb.NewStruct", i))
+		return nil, errors.Wrapf(err,
+			"unable to convert interface %v to a form acceptable to structpb.NewStruct", i)
 	}
 	ret, err := structpb.NewStruct(encoded)
 	if err != nil {
@@ -79,7 +79,7 @@ func StructToStructPbIgnoreOmitEmpty(i interface{}) (*structpb.Struct, error) {
 }
 
 // takes a go type and tries to make it a better type for converting to grpc.
-func toInterface(data interface{}, shouldOmitEmpty bool) (interface{}, error) {
+func toInterface(data interface{}, ignoreOmitEmpty bool) (interface{}, error) {
 	if data == nil {
 		return data, nil
 	}
@@ -95,17 +95,17 @@ func toInterface(data interface{}, shouldOmitEmpty bool) (interface{}, error) {
 	var err error
 	switch t.Kind() {
 	case reflect.Struct:
-		newData, err = structToMap(data, shouldOmitEmpty)
+		newData, err = structToMap(data, ignoreOmitEmpty)
 		if err != nil {
 			return nil, err
 		}
 	case reflect.Map:
-		newData, err = marshalMap(data, shouldOmitEmpty)
+		newData, err = marshalMap(data, ignoreOmitEmpty)
 		if err != nil {
 			return nil, err
 		}
 	case reflect.Slice:
-		newData, err = marshalSlice(data, shouldOmitEmpty)
+		newData, err = marshalSlice(data, ignoreOmitEmpty)
 		if err != nil {
 			return nil, err
 		}
@@ -148,9 +148,8 @@ func isEmptyValue(v reflect.Value) bool {
 }
 
 // structToMap attempts to coerce a struct into a form acceptable by grpc.
-// shouldOmitEmpty specifies whether to ignore empty values if they are tagged omitempty or
-// keep all values.
-func structToMap(data interface{}, shouldOmitEmpty bool) (map[string]interface{}, error) {
+// ignoreOmitEmpty specifies whether to ignore the omitEmpty tag.
+func structToMap(data interface{}, ignoreOmitEmpty bool) (map[string]interface{}, error) {
 	t := reflect.TypeOf(data)
 	if t.Kind() == reflect.Ptr {
 		t = t.Elem()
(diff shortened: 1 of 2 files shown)


protoutils/protoutils_test.go
```diff
@@ -332,24 +330,33 @@ func TestStructToStructPb(t *testing.T) {
 	}
 }
 
+func TestStructToStructPbOmitEmpty(t *testing.T) {
```
agreenb, line 333: Can you add the same test to TestStructToStructPb and ensure that it instead equals {} there?

protoutils/protoutils.go
```diff
@@ -128,8 +178,10 @@ func isEmptyValue(v reflect.Value) bool {
 	}
 }
 
-// structToMap attempts to coerce a struct into a form acceptable by grpc. Ignores omitempty tags.
-func structToMap(data interface{}) (map[string]interface{}, error) {
+// structToMap attempts to coerce a struct into a form acceptable by grpc.
+// shouldOmit specifies whether to ignore empty values if they are tagged omit empty or
```
agreenb, line 182: [nit] shouldOmit -> shouldOmitEmpty, and `tagged omit empty` -> `tagged omitempty`

protoutils/protoutils.go
```diff
@@ -24,12 +26,45 @@ func InterfaceToMap(data interface{}) (map[string]interface{}, error) {
...
+		fallthrough
+	default:
+		return nil, errors.Errorf("data of type %T and kind %s not a struct or a map-like object", data, t.Kind().String())
+	}
+	return res, nil
+}
+
+//nolint:dupl
+func interfaceToMapKeepEmpty(data interface{}) (map[string]interface{}, error) {
+	if data == nil {
```
agreenb, line 51: To not duplicate the inner logic, could you have a interfaceToMap helper which takes `data` and `shouldOmitEmpty`? InterfaceToMap can call that with `true`, and then you can get rid of `interfaceToMapKeepEmpty` and just call the helper directly from `StructToStructPbIgnoreOmitEmpty`.

protoutils/protoutils_test.go
```diff
@@ -182,24 +181,24 @@ func TestInterfaceToMap(t *testing.T) {
 
 func TestMarshalMap(t *testing.T) {
 	t.Run("not a valid map", func(t *testing.T) {
-		_, err := marshalMap(simpleStruct)
+		_, err := marshalMap(simpleStruct, true)
```
agreenb, line 184: Can extract `true` into a `shouldOmitEmpty` const that's used throughout

agreenb (changes requested):
Nice work on updating the goutils code to work for our desired capture/sync behavior! Just a few things, but overall looks good.

protoutils/protoutils_test.go
```diff
@@ -370,7 +379,7 @@ type (
 )
 
 type OmitStruct struct {
-	X float64 `json:"x"`
+	X float64 `json:"x,omitempty"`
```
agreenb, line 382: Ah I missed this before - you can revert and I see that there's a test for `{"omit struct", OmitStruct{}, map[string]interface{}{"x": 0.0}`, thanks!

protoutils/protoutils.go
```diff
@@ -59,8 +63,23 @@ func StructToStructPb(i interface{}) (*structpb.Struct, error) {
 	return ret, nil
 }
 
+// StructToStructPbIgnoreOmitEmpty converts an arbitrary Go struct to a *structpb.Struct. Only exported fields are included in the
+// returned proto and any omitempty tag is ignored.
+func StructToStructPbIgnoreOmitEmpty(i interface{}) (*structpb.Struct, error) {
+	encoded, err := interfaceToMapHelper(i, false)
+	if err != nil {
+		return nil, errors.Wrap(err,
+			fmt.Sprintf("unable to convert interface %v to a form acceptable to structpb.NewStruct", i))
```
dmhilly, line 72: [minor] You can use `errors.Wrapf` here to remove the need for fmt.Sprintf.

protoutils/protoutils_test.go
```diff
@@ -332,24 +332,33 @@ func TestStructToStructPb(t *testing.T) {
 	}
 }
 
+func TestStructToStructPbOmitEmpty(t *testing.T) {
+	expected := map[string]interface{}{"x": 0.0, "y": 0.0}
+	t.Run("StructToStructPBIgnoreOmitEmpty includes all values", func(t *testing.T) {
+		data, err := StructToStructPbIgnoreOmitEmpty(OmitStruct{})
```
dmhilly, line 338: I think I'm missing something; where is OmitStruct defined?

protoutils/protoutils_test.go
```diff
@@ -332,24 +332,33 @@ func TestStructToStructPb(t *testing.T) {
 	}
 }
 
+func TestStructToStructPbOmitEmpty(t *testing.T) {
+	expected := map[string]interface{}{"x": 0.0, "y": 0.0}
+	t.Run("StructToStructPBIgnoreOmitEmpty includes all values", func(t *testing.T) {
```
dmhilly, line 337: [nit] Just an FYI that it's okay to not run this as a subtest (within t.Run) since you only have logic for one test in this function. t.Run just lets you differentiate between different subtests within a given Test function (and do useful things like run them in parallel, in isolation, etc.). If you did this in anticipation of adding future tests or to have a more descriptive subtest name then it's okay to keep it the way you have it here, just wanted you to be aware!

dmhilly (changes requested):
Nice work Vijay! Left a couple comments.

AaronCasas (changes requested):
Just one small comment on the `shouldPmitEmpty` naming, and +1 on Devin's question about OmitEmpty. Overall looks great

protoutils/protoutils.go
```diff
@@ -13,6 +13,10 @@ import (
 // InterfaceToMap attempts to coerce an interface into a form acceptable by structpb.NewStruct.
 // Expects a struct or a map-like object.
 func InterfaceToMap(data interface{}) (map[string]interface{}, error) {
+	return interfaceToMapHelper(data, true)
+}
+
+func interfaceToMapHelper(data interface{}, shouldOmitEmpty bool) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
vijayvuyyuru: Yes you are correct. I'll change the name

protoutils/protoutils_test.go
vijayvuyyuru: Ah makes sense, ill change it.

protoutils/protoutils_test.go
vijayvuyyuru: https://github.com/viamrobotics/goutils/blob/c6d1b7e6309fff9e05367c8c0f0628a6dcb74c88/protoutils/protoutils_test.go#L372 Its at the bottom of the file. It just has a field that is tagged as omitEmpty.

AaronCasas (approved):
LGTM once tests are passing

agreenb (approved):
Really nice work!

protoutils/protoutils.go
```diff
@@ -13,6 +13,10 @@ import (
 // InterfaceToMap attempts to coerce an interface into a form acceptable by structpb.NewStruct.
 // Expects a struct or a map-like object.
 func InterfaceToMap(data interface{}) (map[string]interface{}, error) {
+	return interfaceToMapHelper(data, true)
```
dmhilly, line 16: I'm confused, now it looks like there is a mixture of ignoreOmitEmpty and shouldOmitEmpty. If we're going with ignoreOmitEmpty, then this should be false, correct?

protoutils/protoutils.go
```diff
@@ -59,8 +63,23 @@ func StructToStructPb(i interface{}) (*structpb.Struct, error) {
 	return ret, nil
 }
 
+// StructToStructPbIgnoreOmitEmpty converts an arbitrary Go struct to a *structpb.Struct. Only exported fields are included in the
+// returned proto and any omitempty tag is ignored.
+func StructToStructPbIgnoreOmitEmpty(i interface{}) (*structpb.Struct, error) {
+	encoded, err := interfaceToMapHelper(i, false)
```
dmhilly, line 69: Should this be `true`?

protoutils/protoutils_test.go
dmhilly: Spoke with Vijay offline; he didn't end up adding a test for this because one already exists [here](https://github.com/viamrobotics/goutils/pull/204/files#diff-9c3c3101f109f0429d739b33ceb0acd4463bad9e25bc8c8d1b2769d000a7c501R75)

Alexa had commented this elsewhere but I missed it

agreenb (commented):
Request to run all RDK tests locally while referencing this goutils version before merging, to ensure that all still passes. Nice work Vijay!

dmhilly (approved):
LGTM! Agree with Alexa about running tests locally. Nice work 😄

PR #205, opened 2023-09-29
DATA-2031 windows-specific build

abe-winter:
## What
- move signal registration for ContextualMain to architecture-specific file
- note: I'm not sure what this is used for and it's possible ContextualMain won't work correctly on windows (but seems like SIGTERM will still support some use cases)
- this is for CLI, which doesn't use ContextualMain. another option here would be to move the entire ContextualMain logic to runtime_posix.go
## Why
- this is to allow a windows-specific build of the CLI in rdk repo

### DIFF
--- a/runtime_windows.go
+++ b/runtime_windows.go
@@ -1,0 +1,7 @@
+package utils
+
+import "os"
+
+func notifySignals(channel chan os.Signal) {
+	println("skipping notifySignals on windows platform")
+}
--- a/etc/.golangci.yaml
+++ b/etc/.golangci.yaml
@@ -76,6 +76,9 @@ linters-settings:
       - shadow
   lll:
     line-length: 140
+  revive:
+    disable:
+    - package-comments
 issues:
   exclude:
     - composites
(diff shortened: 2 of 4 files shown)


runtime_windows.go
```diff
@@ -0,0 +1,7 @@
+package utils
```
maximpertsov, line 1: Do we need a directive to only build on windows platforms here?

```suggestion
//go:build windows

package utils
```

maximpertsov (changes requested):
looks good! just one question regarding build directives from me.

to your comment of

> this is for CLI, which doesn't use ContextualMain. another option here would be to move the entire ContextualMain logic to runtime_posix.go

I think the way you did it is fine

runtime_windows.go
abe-winter: you can use GOOS + GOARCH in file suffixes to get build constraints

https://pkg.go.dev/go/build#hdr-Build_Constraints

(search for `source_windows.go` on that page)

maximpertsov (approved):

PR #206, opened 2023-10-03
add ProcessConfig.Username field

abe-winter:
## What
- add Username in managedProcess + ProcessConfig
- when this field is present, managedProcess looks up the user and runs as them
- add a new CI step to test this (because it needs to run with setuid capability)
## Why
- to support running non-privileged subprocess (if necessary in the future)

### DIFF
--- a/pexec/managed_process_test.go
+++ b/pexec/managed_process_test.go
@@ -212,8 +212,8 @@ func TestManagedProcessStart(t *testing.T) {
 				Log:      true,
 			}, golog.NewTestLogger(t))
 			test.That(t, proc.Start(context.Background()), test.ShouldBeNil)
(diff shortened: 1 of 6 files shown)


pexec/managed_process_test.go
```diff
@@ -22,6 +25,22 @@ import (
...
+func subprocUser() *user.User {
+	var userInfo *user.User
+	var err error
+	if usernameFromEnv := os.Getenv("TEST_SUBPROC_USER"); len(usernameFromEnv) > 0 {
+		userInfo, err = user.Lookup(usernameFromEnv)
+	} else {
+		userInfo, err = user.Current()
+	}
+	if err != nil {
+		panic(err)
```
benjirewis, line 39: Can you just return the error and fail the test? Why panic?

abe-winter: pythonism that leaked into my go

fixed in ace033ba

benjirewis (approved):
LGTM

PR #207, opened 2023-10-04
RSDK-4793: Remote logging fails to transmit certain error messages

hexbabe:
https://viam.atlassian.net/browse/RSDK-4793

This probably won't catch all the complex types in the future, but it solves the one James ran into on the ticket. I think for the future we can continue adding cases that the `reflect` package fails to catch before the switch case statements.


.github/workflows/proto-checks.yml
```diff
@@ -22,9 +22,17 @@ jobs:
...
     - uses: actions/checkout@v3
     - uses: bufbuild/buf-setup-action@v1
+      with:
+        username: soyybean
+        password: dckr_pat_kXtItMP6C7iN17oWJFCZ2JseVsg
     # Run all Lint runs
     - uses: bufbuild/buf-lint-action@v1
+      with:
+        username: soyybean
+        password: dckr_pat_kXtItMP6C7iN17oWJFCZ2JseVsg
```
hexbabe: i immediately revoked access to this! no sweat 😅

vijayvuyyuru (changes requested):
Great work on this Sean! Just one small comment about testing.

protoutils/protoutils_test.go
```diff
@@ -361,6 +362,13 @@ func TestToInterfaceWeirdBugUint8(t *testing.T) {
 	test.That(t, x, test.ShouldEqual, a)
 }
 
+func TestToInterfaceErrno(t *testing.T) {
```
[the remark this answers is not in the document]
hexbabe: I did some possibly questionable things in order for the test value decoding to work. Let me know if there are things that you think can be handled better @vijayvuyyuru

protoutils/protoutils_test.go
```diff
@@ -127,6 +131,12 @@ var (
 			map[string]interface{}{"UintValue": uint(1)},
 			SingleUintStruct{},
 		},
+		{
+			"struct with errno",
+			errnoStruct,
+			map[string]interface{}{"Errno": float64(errnoVal)}, // cast float64 because pb to map conversion supports double for nums
+			ErrnoStructReturn{},
```
hexbabe, line 138: Decoder fails when it needs to decode a uintptr (syscall.Errno), so we change the return type to something it can parse (int)   

```
 /Users/seanyu/Projects/goutils/protoutils/protoutils_test.go:263: Expected: nil
        Actual:   '1 error(s) decoding:
        
        * Errno: unsupported type: uintptr'
```

vijayvuyyuru (approved):
LGTM, nice work on this!

PR #208, opened 2023-10-04
RSDK-4473 Make stream message reception sensitive to peer finishing

benjirewis:
Could be related to https://viam.atlassian.net/browse/RSDK-5235 and/or https://viam.atlassian.net/browse/RSDK-4473.

Cancels channel context when peer has finished (connection was closed, or connection state moved to disconnected, failed or closed). Respects channel context in base stream `RecvMsg`, so streams 
(both client and server) will not hang if connection gets closed mid-operation-invocation.

### DIFF
--- a/rpc/wrtc_base_stream.go
+++ b/rpc/wrtc_base_stream.go
@@ -17,12 +17,9 @@ import (
 )
 
 type webrtcBaseStream struct {
-	mu     sync.RWMutex
-	ctx    context.Context
-	cancel context.CancelFunc
-	// channelCtx is the context from the underlying channel (either a
-	// webrtcClientChannel or a webrtcServerChannel).
-	channelCtx    context.Context
+	mu            sync.RWMutex
+	ctx           context.Context
+	cancel        context.CancelFunc
 	stream        *webrtcpb.Stream
 	msgCh         chan []byte
 	onDone        func(id uint64)
@@ -41,21 +38,16 @@ type webrtcBaseStream struct {
 func newWebRTCBaseStream(
 	ctx context.Context,
 	cancelCtx func(),
-	// Disable revive linter that complains about this context.Context not being
-	// the first parameter of the function.
-	//nolint:revive
-	channelCtx context.Context,
 	stream *webrtcpb.Stream,
 	onDone func(id uint64),
 	logger golog.Logger,
 ) *webrtcBaseStream {
 	bs := webrtcBaseStream{
-		ctx:        ctx,
-		cancel:     cancelCtx,
-		channelCtx: channelCtx,
-		stream:     stream,
-		onDone:     onDone,
-		logger:     logger,
+		ctx:    ctx,
+		cancel: cancelCtx,
+		stream: stream,
+		onDone: onDone,
+		logger: logger,
 	}
 	bs.msgCh = make(chan []byte, 1)
 	return &bs
@@ -106,11 +98,13 @@ func (s *webrtcBaseStream) RecvMsg(m interface{}) error {
 		}
 	}
 
-	// RSDK-4473: There are two ways a stream can be signaled that it should give
-	// up receiving a message:
-	//   - `s.ctx` is canceled due to a timeout or some other grpc/webrtc error.
-	//   - `s.channelCtx` is canceled when the underlying webrtc data channel
+	// RSDK-4473: There are three ways a stream can be signaled that it should
+	// give up receiving a message:
+	//   - `s.ctx` errors due to a timeout or some other grpc/webrtc error.
+	//   - `s.ctx` is canceled when the underlying webrtc data channel
 	//      connection experiences an error.
+	//   - `s.msgCh` is closed when the stream, channel or connection has been
+	//      closed.
 	select {
 	case <-s.ctx.Done():
 		msgBytes, err := checkLastOrErr(s.ctx.Err())
@@ -121,15 +115,6 @@ func (s *webrtcBaseStream) RecvMsg(m interface{}) error {
 			return proto.Unmarshal(msgBytes, m.(proto.Message))
 		}
 		return s.ctx.Err()
-	case <-s.channelCtx.Done():
-		msgBytes, err := checkLastOrErr(s.channelCtx.Err())
-		if err != nil {
-			return err
-		}
-		if msgBytes != nil {
-			return proto.Unmarshal(msgBytes, m.(proto.Message))
-		}
-		return s.channelCtx.Err()
 	case msgBytes, ok := <-s.msgCh:
 		if ok {
 			return proto.Unmarshal(msgBytes, m.(proto.Message))
--- a/rpc/wrtc_base_stream.go
+++ b/rpc/wrtc_base_stream.go
@@ -17,9 +17,12 @@ import (
 )
 
 type webrtcBaseStream struct {
-	mu            sync.RWMutex
-	ctx           context.Context
-	cancel        context.CancelFunc
+	mu     sync.RWMutex
+	ctx    context.Context
+	cancel context.CancelFunc
+	// channelCtx is the context from the overlying channel (either a
+	// webrtcClientChannel or a webrtcServerChannel).
+	channelCtx    context.Context
 	stream        *webrtcpb.Stream
 	msgCh         chan []byte
 	onDone        func(id uint64)
@@ -38,16 +41,21 @@ type webrtcBaseStream struct {
 func newWebRTCBaseStream(
 	ctx context.Context,
 	cancelCtx func(),
+	// Disable revive linter that complains about this context.Context not being
+	// the first parameter of the function.
+	//nolint:revive
+	channelCtx context.Context,
 	stream *webrtcpb.Stream,
 	onDone func(id uint64),
 	logger golog.Logger,
 ) *webrtcBaseStream {
 	bs := webrtcBaseStream{
-		ctx:    ctx,
-		cancel: cancelCtx,
-		stream: stream,
-		onDone: onDone,
-		logger: logger,
+		ctx:        ctx,
+		cancel:     cancelCtx,
+		channelCtx: channelCtx,
+		stream:     stream,
+		onDone:     onDone,
+		logger:     logger,
 	}
 	bs.msgCh = make(chan []byte, 1)
 	return &bs
@@ -94,6 +102,7 @@ func (s *webrtcBaseStream) RecvMsg(m interface{}) error {
 			return nil, nil
 		}
 	}
+
 	select {
 	case <-s.ctx.Done():
 		msgBytes, err := checkLastOrErr(s.ctx.Err())
@@ -104,6 +113,15 @@ func (s *webrtcBaseStream) RecvMsg(m interface{}) error {
 			return proto.Unmarshal(msgBytes, m.(proto.Message))
 		}
 		return s.ctx.Err()
+	case <-s.channelCtx.Done():
+		msgBytes, err := checkLastOrErr(s.channelCtx.Err())
+		if err != nil {
+			return err
+		}
+		if msgBytes != nil {
+			return proto.Unmarshal(msgBytes, m.(proto.Message))
+		}
+		return s.channelCtx.Err()
 	case msgBytes, ok := <-s.msgCh:
 		if ok {
 			return proto.Unmarshal(msgBytes, m.(proto.Message))
--- a/rpc/wrtc_base_stream.go
+++ b/rpc/wrtc_base_stream.go
@@ -20,7 +20,7 @@ type webrtcBaseStream struct {
 	mu     sync.RWMutex
 	ctx    context.Context
 	cancel context.CancelFunc
-	// channelCtx is the context from the overlying channel (either a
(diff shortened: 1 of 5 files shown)


rpc/wrtc_base_stream.go
```diff
@@ -104,6 +113,15 @@ func (s *webrtcBaseStream) RecvMsg(m interface{}) error {
...
 		return s.ctx.Err()
+	case <-s.channelCtx.Done():
+		msgBytes, err := checkLastOrErr(s.channelCtx.Err())
+		if err != nil {
+			return err
+		}
+		if msgBytes != nil {
+			return proto.Unmarshal(msgBytes, m.(proto.Message))
+		}
+		return s.channelCtx.Err()
```
benjirewis, line 124: This code is redundant with the case above, but `fallthrough` does not exist for select, and you cannot have multiple `case` values for a single logical branch. Abstracting into a function felt confusing with the pre-existing `checkLastOrErr` inline function.

benjirewis (commented):
I could not think of a great way to test this behavior in CI. We really need more of an integration test to ensure that stream message reception does not hang when connections close unexpectedly in the middle of operations. Let me know if you disagree/have an idea for how to test here.

benjirewis:
I had mentioned offline that I'd add the same logic to `SendMsg` for client and server code. `SendMsg` does not work in the same way as `RecvMsg`, and interacts directly with the channel and underlying connection due to its need to send data across the wire. I could not find a place where `SendMsg` would get caught selecting between cases, unaware that the underlying connection was unhealthy.

rpc/wrtc_base_stream.go
dgottlieb: I think this is fine the way it is.

`checkLastOrErr` confuses me. It feels like an optimization. Like if we hit the timeout case, we should do one last check if we got any messages. If it's for correctness, I wish we knew why so we could document. If it's an optimization, I'm unconvinced it's helpful and removing it would elide us from worrying about this redundancy.

Can keep it in mind to poke at these code paths more the next time we have a reason to look at this area.

rpc/wrtc_base_stream.go
```diff
@@ -38,16 +41,21 @@ type webrtcBaseStream struct {
 func newWebRTCBaseStream(
```
dgottlieb, line 41: I'm inclined to leave a comment somewhere (but my guess was here was the best surfaced, but maybe RecvMsg?):

```
RSDK-4473: There are two ways a stream can be signaled that it should give up receiving a message:
- `ctx` is canceled due to a timeout or from a logical error.
- `channelCtx` is canceled when the underlying webrtc data channel connection experiences an error.
```

I don't feel great in my summarization of the former as a "logical error". It felt more appropriate than say "an error from the goutils grpc/webrtc code".

dgottlieb:
And let's associate the commit with a ticket. RSDK-4473 seems fine, but creating a new ticket just describing the problem "WebRTC channel errors can be missed".

rpc/wrtc_base_stream.go
benjirewis: >  Like if we hit the timeout case, we should do one last check if we got any messages.

That's how I understand it, too. If we fall into a context error branch, we see whether the msgCh contained a message, too. If it did, we throw that message into the `m` output parameter. If it didn't, we return the context error. If the channel was already closed, we try to return the last error stored on the stream, replacing an internal expected closure with `io.EOF`. The one branch I don't really understand is:
```
if origErr == nil {
   return nil, io.EOF
}
```
I don't know how we could ever fall into that case if the context is done and `ctx.Err()` was passed in as `origErr`.

In any case, I also can't describe the "correctness" that function is trying to enforce, and it feels more like an attempt to respect actual errors on the stream over context errors when returning an error to the user (hoping that actual errors are more informative/relevant than context errors). I've left a comment in `checkLastOrErr` to explain my understanding; let me know what you think.

rpc/wrtc_base_stream.go
benjirewis: Yeah good idea; left the comment you wrote out before the relevant `select` case; changed the wording slightly.

dgottlieb (approved):

maximpertsov (approved):
looks good!

[opt] would it be feasible to add a test for this new behavior? maybe in `wrtc_(client|server)_channel_test.sh`? totally cool punt since we are still trying out different solutions for this issue.

benjirewis:
> [opt] would it be feasible to add a test for this new behavior? maybe in wrtc_(client|server)_channel_test.sh? totally cool punt since we are still trying out different solutions for this issue.

Yeah it's a good question. I mucked around with this for a bit and found it to be hard, but let me try for another 30 and see if I can get something.

rpc/wrtc_base_stream.go
maximpertsov: [q] I'm totally happy with the current implementation, but just out of curiosity - is there a way to "merge" contexts in golang? not sure if that's even appropriate here, but say we did end up wanting to react to both contexts here the same way?

benjirewis: Another good q. There's no obvious function/method I know of to just "merge" two contexts, but there are two options I can think of that might be more elegant than storing two contexts on the same struct. The first method is often referred to as a "fan-in" of channels (see [this old article](https://go.dev/blog/pipelines) on go concurrency patterns); we _could_ fan-in the `Done` channels of the two contexts and check the aggregated channel. The second method is establishing a parent/child relationship between the channel context and the stream context. That way, the stream context's `Done` channel will receive a value when the channel context's does (but _not_ vice-versa). I'll try the latter method locally and see if it cleans things up.

benjirewis: > There's no obvious function/method I know of to just "merge" two contexts

...yes [there is](https://github.com/viamrobotics/goutils/blob/main/context.go#L12-L25) haha; it's in this repo. Things seem cleaner using `MergeContext*`, and the hang still disappears in my repro, so I'll use those methods. FWIW they use the "fan-in" method mentioned above. Thanks for the idea!

I also discovered that webRTC _server_ streams seem to already incorporate the server channel's context into the context passed to `RecvMsg` [here](https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_server_channel.go#L121). So, I don't think we need any extra server logic for server-side `RecvMsg` hangs.

rpc/wrtc_client_stream.go
```diff
@@ -44,7 +44,7 @@ func newWebRTCClientStream(
 	onDone func(id uint64),
 	logger golog.Logger,
 ) *webrtcClientStream {
-	ctx, cancel := context.WithCancel(ctx)
+	ctx, cancel := utils.MergeContext(channel.ctx, ctx)
```
dgottlieb, line 47: I can't tell this is a behavior change. The specific concern that I'm struggling to reason about:
* If the `ctx` (input parameter from the client/application code) is canceled, that results in [this cancel](https://github.com/viamrobotics/goutils/blob/45e9b67aa447d1d1d09aaf6390c4507b6ab44886/context.go#L48) being called.
* That cancel comes from this [construction of](https://github.com/viamrobotics/goutils/blob/45e9b67aa447d1d1d09aaf6390c4507b6ab44886/context.go#L13) `context.WithCancel(channel.ctx)`.
* Will that result in the `channel.ctx` being canceled?

It's fine if a channel context cancel propagates to streams/API calls (we presume the connection is dead). But it's not ok for a stream/API timeout to break the channel.

benjirewis: > Will that result in the channel.ctx being canceled?

Fair concern; I think the answer is no. The code of merged contexts is tough to follow, and I'm getting confused trying to explain my understanding with prose haha. Let me write you a test locally to try to prove that when merging contexts, the first context parameter cannot be canceled by canceling the second or by canceling the merged context.

benjirewis: So this test passes locally:

```
func TestMergedContext(t *testing.T) {
	channelCtx, cancelChannelCtx := context.WithCancel(context.Background())
	streamCtx, cancelStreamCtx := context.WithCancel(context.Background())
	mergedCtx, cancelMergedCtx := utils.MergeContext(channelCtx, streamCtx)

	cancelMergedCtx()
	test.That(t, mergedCtx.Err(), test.ShouldNotBeNil)
	test.That(t, mergedCtx.Err(), test.ShouldBeError, context.Canceled)
	test.That(t, streamCtx.Err(), test.ShouldBeNil)
	test.That(t, channelCtx.Err(), test.ShouldBeNil)

	cancelStreamCtx()
	test.That(t, mergedCtx.Err(), test.ShouldNotBeNil)
	test.That(t, mergedCtx.Err(), test.ShouldBeError, context.Canceled)
	test.That(t, streamCtx.Err(), test.ShouldNotBeNil)
	test.That(t, streamCtx.Err(), test.ShouldBeError, context.Canceled)
	test.That(t, channelCtx.Err(), test.ShouldBeNil)

	cancelChannelCtx()
	test.That(t, mergedCtx.Err(), test.ShouldNotBeNil)
	test.That(t, mergedCtx.Err(), test.ShouldBeError, context.Canceled)
	test.That(t, streamCtx.Err(), test.ShouldNotBeNil)
	test.That(t, streamCtx.Err(), test.ShouldBeError, context.Canceled)
	test.That(t, channelCtx.Err(), test.ShouldNotBeNil)
	test.That(t, channelCtx.Err(), test.ShouldBeError, context.Canceled)
}
```
I'm hoping that explains the relationships between the first, second and merged contexts of `MergeContext`. Context erroring propagates as first -> second -> merged.

rpc/wrtc_client_channel_test.go
```diff
@@ -538,3 +538,53 @@ func TestWebRTCClientChannelWithInterceptor(t *testing.T) {
...
+	// Regression test for RSDK-4473.
+	testutils.SkipUnlessInternet(t)
+	logger := golog.NewTestLogger(t)
+	pc1, pc2, dc1, dc2 := setupWebRTCPeers(t)
+
+	clientCh := newWebRTCClientChannel(pc1, dc1, logger, nil, nil)
+	defer func() {
+		test.That(t, clientCh.Close(), test.ShouldBeNil)
+	}()
+	serverCh := newBaseChannel(context.Background(), pc2, dc2, nil, logger)
```
benjirewis, line 552: We use base channels to represents server channels in this test file. We _could_ make a full server channel, but it requires [some extra logic](https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_server_channel_test.go#L31-L43) setting up the actual server, that I felt was unnecessary, as what we're testing here is really client-side.

rpc/wrtc_client_channel_test.go
```diff
@@ -538,3 +538,53 @@ func TestWebRTCClientChannelWithInterceptor(t *testing.T) {
 
 	wg.Wait()
 }
+
+func TestWebRTCClientChannelCanStopStreamRecvMsg(t *testing.T) {
```
benjirewis, line 542: This test sets up a client-server connection, sends a message from the client, does not respond from the server, closes the underlying client peer connection, and tries to `RecvMsg` from the client side. It passes with the changes in this PR, and hangs forever on `main`.

I don't think it's a _perfect_ representation of the issue, but it's probably close enough.

benjirewis (commented):
Ah sorry will have to re-request both of you given code changes + the new test. Appreciate the reviews 🙏🏻 .

rpc/wrtc_client_channel_test.go
maximpertsov: cool, i guess the main thing is does this result in a hang when we remove the new `select case` we added?

maximpertsov (approved):

rpc/wrtc_client_stream.go
dgottlieb: Cool, thanks for checking!

dgottlieb (approved):

rpc/wrtc_client_channel_test.go
```diff
@@ -538,3 +538,53 @@ func TestWebRTCClientChannelWithInterceptor(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+
+	test.That(t, clientStream.SendMsg(someStatus.Proto()), test.ShouldBeNil)
+
+	// Close client peer connection before receiving a client message. Assert
+	// that RecvMsg does not hang.
+	test.That(t, pc1.Close(), test.ShouldBeNil)
+	var respStatus pbstatus.Status
+	err = clientStream.RecvMsg(&respStatus)
+	test.That(t, err, test.ShouldBeError, context.Canceled)
```
dgottlieb, line 586: This is a clever test. I see the this is failing in CI -- getting an EOF instead of a context canceled. Once you understand what's happening and what should happen, can you add some details? I suspect an EOF here is actually bad as its often considered a "mutual" end of communication.

benjirewis: >  I suspect an EOF here is actually bad as its often considered a "mutual" end of communication.

That _sounds_ right to me; I assumed something weird was happening with that `checkLastOrErr` and we were replacing a context canceled with EOF due to an internal expected closure. That's not really the point of the test though, so definitely unexpected. I'll take a closer look and report back.

PR #211, opened 2023-10-16
RSDK-4293 wait for answerer to finish completely

maximpertsov:
Wait on all goroutines related to ice candidate negotiation to complete before closing the server.

Testing
- goleak reproduced in [this commit](https://github.com/viamrobotics/goutils/pull/211/commits/9e03120cf4087b105d7d62ae022c5a24a6a3cf0e)

### DIFF
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -306,7 +306,7 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 			return err
 		}
 
-		var callFlowWG sync.WaitGroup
+		var pendingCandidates sync.WaitGroup
 		waitOneHost := make(chan struct{})
 		var waitOneHostOnce sync.Once
 		pc.OnICECandidate(func(icecandidate *webrtc.ICECandidate) {
@@ -314,7 +314,7 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 				return
 			}
 			if icecandidate != nil {
-				callFlowWG.Add(1)
+				pendingCandidates.Add(1)
 				if icecandidate.Typ == webrtc.ICECandidateTypeHost {
 					waitOneHostOnce.Do(func() {
 						close(waitOneHost)
@@ -327,7 +327,7 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 				defer ans.activeBackgroundWorkers.Done()
 
 				if icecandidate != nil {
-					defer callFlowWG.Done()
+					defer pendingCandidates.Done()
 				}
 
 				select {
@@ -343,7 +343,7 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 						ans.logger.Debug("Sleeping to delay the end of the negotiation")
 						time.Sleep(1 * time.Second)
 					}
-					callFlowWG.Wait()
+					pendingCandidates.Wait()
 					if err := sendDone(); err != nil {
 						sendErr(err)
 					}
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -329,10 +329,6 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 					return
 				}
 				if icecandidate == nil {
-					// TODO(mp): remove this sleep! it's there to repro a flaky test
-					// failure:
-					// https://viam.atlassian.net/browse/RSDK-4293?focusedCommentId=20176
-					time.Sleep(2 * time.Second)
 					callFlowWG.Wait()
 					if err := sendDone(); err != nil {
 						sendErr(err)
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -326,6 +326,7 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 					return
 				}
 				if i == nil {
+					time.Sleep(2 * time.Second)
 					callFlowWG.Wait()
 					if err := sendDone(); err != nil {
 						sendErr(err)
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -320,6 +320,9 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 			}
 			// must spin off to unblock the ICE gatherer
 			utils.PanicCapturingGo(func() {
+				ans.activeBackgroundWorkers.Add(1)
+				defer ans.activeBackgroundWorkers.Done()
+
 				select {
 				case <-initSent:
 				case <-exchangeCtx.Done():
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -321,8 +321,8 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 			}
 			// must spin off to unblock the ICE gatherer
 			utils.PanicCapturingGo(func() {
-				// ans.activeBackgroundWorkers.Add(1)
-				// defer ans.activeBackgroundWorkers.Done()
+				ans.activeBackgroundWorkers.Add(1)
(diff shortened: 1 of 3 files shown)


rpc/wrtc_signaling_answerer.go
```diff
@@ -306,34 +307,42 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
...
+
 				select {
 				case <-initSent:
 				case <-exchangeCtx.Done():
 					return
 				}
-				if i == nil {
+				// there are no more candidates coming during this negotiation
+				if icecandidate == nil {
+					if _, ok := os.LookupEnv("TEST_DELAY_FINISH_NEGOTIATION"); ok {
```
dgottlieb, line 334: I think this is a neat idea making a conditional to replicate a bug based on an environment variable. Hadn't seen that done before. 

Did you intend to leave this in for future developers to be able to reproduce this bug? If so, can you mark this with a comment like:
```
RSDK-4293: Introducing a sleep here replicates the conditions for a prior goroutine leak.
```

maximpertsov:
> I'm also curious if this [fix to analogous looking](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579) code also applies here.

yeah this looks like the exact same fix but client-side - I can try to add a similar repro / test

maximpertsov:
> > I'm also curious if this [fix to analogous looking](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579) code also applies here.
> 
> yeah this looks like the exact same fix but client-side - I can try to add a similar repro / test

I [tried to replicate the goleak on the client side](https://github.com/maximpertsov/goutils/compare/RSDK-4293-fix-slow-answerer...maximpertsov:goutils:RSDK-3286-repro-goleak-client-side?expand=1) but it's not as easy to do with our current test cases.

The client side of the negotiation is constructed a bit differently - for example there is no high-level background worker waitgroup like we have on the server-side. Given all of that, I'm going to leave the client-side as is for now. We can revisit if this sort of goleak flake surfaces again.

maximpertsov:
> I'm also curious if this [fix to analogous looking](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579) code also applies here.

Sorry, I don't think I answered your immediate question. I'm actually not sure if the client-side fix is the correct approach - it fixes the goleak but it could introduce other issues related to timing. In that patch, we are reducing the waitgroup queue by 1 if we receive a nil candidates (which signals no more candidates). Does that still work if there are two candidates still being "processed" when we get a nil candidate? Is there an issue if somehow the nil candidate arrives out of order before another candidate? 

That said, in practice these issues might never come up since I assume we "process" candidates very quickly and the out-of-order scenario might not ever happen.

dgottlieb:
Sorry, I linked to the commit and not [the PR](https://github.com/viamrobotics/goutils/pull/178) so some conversation was lost there. Specifically the PR talks about [this early return](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3R245-R246). 

The bad sequence being:
* Having `i != nil`. Which [adds to the "call flow" waitgroup](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3L231)
* but we can [miss the "defer wg.Done"](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3L252)
* when we hit the [early return](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3L243).

dgottlieb:
> Does that still work if there are two candidates still being "processed" when we get a nil candidate? Is there an issue if somehow the nil candidate arrives out of order before another candidate?

My understanding is that the `nil` candidate will always be processed last. And that will happen after all of the prior `OnICECandidate` calls have returned. But because we [shove the meat of the callback into a goroutine](https://github.com/viamrobotics/goutils/pull/211/files#diff-5514904bc577d9192b1f4805ade57d369a09c437b0bc2036986efa4be3b2bd0bR325), those goroutines can certainly be processed out of order.

maximpertsov:
> Sorry, I linked to the commit and not [the PR](https://github.com/viamrobotics/goutils/pull/178) so some conversation was lost there. Specifically the PR talks about [this early return](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3R245-R246).
> 
> The bad sequence being:
> 
>     * Having `i != nil`. Which [adds to the "call flow" waitgroup](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3L231)
> 
>     * but we can [miss the "defer wg.Done"](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3L252)
> 
>     * when we hit the [early return](https://github.com/viamrobotics/goutils/commit/533a16f8455971900fddc3b2ef7c029397b06579#diff-b391ec2d0e42e6403a1a19da01e440025f59562d057f188df7535ec8e50800a3L243).

I see, I didn't consider this scenario. Also I think I misread the patch, we are only calling `wg.Done` earlier for non-nil candidates which is totally the right way to go.

rpc/wrtc_signaling_answerer.go
```diff
@@ -303,37 +306,50 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 			return err
 		}
 
-		var callFlowWG sync.WaitGroup
+		var pendingCandidates sync.WaitGroup
 		waitOneHost := make(chan struct{})
 		var waitOneHostOnce sync.Once
-		pc.OnICECandidate(func(i *webrtc.ICECandidate) {
+		pc.OnICECandidate(func(icecandidate *webrtc.ICECandidate) {
```
dgottlieb, line 312: Also I don't think I mentioned this before, but I really appreciate you renaming the ice candidate variable away from `i`! I have a 4-5 line limit where I'm ok with single letter variables.

rpc/wrtc_signaling_answerer.go
```diff
@@ -303,37 +306,50 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
...
+				pendingCandidates.Add(1)
+				if icecandidate.Typ == webrtc.ICECandidateTypeHost {
 					waitOneHostOnce.Do(func() {
 						close(waitOneHost)
 					})
 				}
 			}
 			// must spin off to unblock the ICE gatherer
 			utils.PanicCapturingGo(func() {
+				ans.activeBackgroundWorkers.Add(1)
```
dgottlieb, line 326: Sorry I missed this before, this `Add` probably needs to be outside the goroutine. I assume that `Wait`ing on `activeBackgroundWorkers` won't happen while inside an `OnICECandidate` callback, but it can happen after `OnICECandidate` returns, but before the goroutine it spawns completes. A small race certainly but this allows for the following interleaving:
```
| OnIceCandidate(Foo) | GoRoutine(Foo)                 | Answerer.activeBackgroundWorkers.Wait() |
|---------------------+--------------------------------+-----------------------------------------|
| PanicCapturingGo    |                                |                                         |
| Return              |                                |                                         |
|                     |                                | Call Wait() // Counter is 0             |
|                     |                                | Return from Wait()                      |
|                     | activeBackgroundWorkers.Add(1) |                                         |
```

dgottlieb (approved):
Pre-emptively approving. If you agree re: moving the `activeBackgroundWorkers.Add(1)` outside the goroutine, feel free to commit that + merge. But happy to discuss look further if what the right thing to do is unclear based on how this code gets used.

rpc/wrtc_signaling_answerer.go
maximpertsov: hah, also my brain is hardwired to treat `i` as an index so it had to change for my sanity. i think i'll do a flyby change for this in the client code as well.

PR #212, opened 2023-10-16
APP-2890: Refactor to make EnsureAuth Public

michaellee1019:
Allows the ability for usages of goutils to call EnsureAuth whenever auth checks are needed that don't follow the convention of RPC Credential Type (For example checks outside of gRPC calls).

### DIFF
--- a/rpc/server_auth.go
+++ b/rpc/server_auth.go
@@ -30,8 +30,8 @@ func (ss *simpleServer) authHandlers(forType CredentialsType) (credAuthHandlers,
 }
 
 const (
-	metadataFieldAuthorization     = "authorization"
-	authorizationValuePrefixBearer = "Bearer "
+	MetadataFieldAuthorization     = "authorization"
+	AuthorizationValuePrefixBearer = "Bearer "
 )
 
 // JWTClaims extends jwt.RegisteredClaims with information about the credentials as well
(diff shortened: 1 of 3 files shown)


ohEmily (approved):

rpc/server_auth.go
```diff
@@ -239,16 +241,16 @@ func (wrapped ctxWrappedServerStream) Context() context.Context {
...
 	}
-	authHeader := md.Get(metadataFieldAuthorization)
+	authHeader := md.Get(MetadataFieldAuthorization)
 	if len(authHeader) != 1 {
 		return "", status.Error(codes.Unauthenticated, "authentication required")
 	}
-	if !strings.HasPrefix(authHeader[0], authorizationValuePrefixBearer) {
-		return "", status.Errorf(codes.Unauthenticated, "expected Authorization: %s", authorizationValuePrefixBearer)
+	if !strings.HasPrefix(authHeader[0], AuthorizationValuePrefixBearer) {
+		return "", status.Errorf(codes.Unauthenticated, "expected Authorization: %s", AuthorizationValuePrefixBearer)
```
ohEmily, line 251: [minor] this is such a bizarre error message?

michaellee1019: Agreed! I updated it to be more specific.

PR #214, opened 2023-10-18
Encapsulate JS build step

maximpertsov:
* Add a `build` step to every package.json file - [this led to some confusion recently](https://viaminc.slack.com/archives/C03LCAHM2TT/p1697567590350979?thread_ts=1697468664.877159&cid=C03LCAHM2TT)
* Remove references to `vite` since we don't it use yet

### DIFF
--- a/rpc/examples/echo/frontend/package.json
+++ b/rpc/examples/echo/frontend/package.json
@@ -8,20 +8,20 @@
     "google-protobuf": "^3.14.0"
   },
   "devDependencies": {
-    "ansi-regex": ">=5.0.1",
-    "set-value": ">=4.0.1",
-    "@webpack-cli/serve": "^1.3.1",
-    "webpack": "^5.74.0",
-    "webpack-cli": "^4.10.0",
     "@types/google-protobuf": "^3.7.4",
+    "@webpack-cli/serve": "^1.3.1",
+    "ansi-regex": ">=5.0.1",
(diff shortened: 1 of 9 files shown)


rpc/examples/echo/frontend/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "scripts": {
     "prepare": "./etc/prepare.sh",
+    "build": "npx webpack",
```
micheal-parks, line 24: Instead of `npx webpack`, it would be a bit safer to `npm i --save-dev webpack` to get a pinned version in the devDependencies, then you could just:
```js
    "build": "webpack",
```

micheal-parks (approved):

rpc/examples/echo/frontend/package.json
maximpertsov: This seems to break stuff in CI: https://github.com/viamrobotics/goutils/pull/214/commits/f01c993645616882b474ea5b895d6706f1f8463d + https://github.com/viamrobotics/goutils/pull/214/commits/72354b728fed3344e5a999d366891be0d30ca9cd

I'm just going to keep it as `npx webpack` for now to mirror the setup that exists. Then hopefully move to vite soon after that.

micheal-parks: ah dang, agreed that's the best course of action then 👍

PR #215, opened 2023-10-24
RSDK-4952 - pass priority attribute

purplenicole730:
[Related TS SDK change](https://github.com/viamrobotics/viam-typescript-sdk/pull/181)
[Jira ticket](https://viam.atlassian.net/browse/RSDK-5228)

Allowing for a custom priority attribute to be pushed in.

Tested with a board running the MicroRDK to confirm default and customized priority works as expected. Also tested with a robot running the regular RDK (with and without a priority attribute) to confirm everything else still works. The echo example was also run to confirm it ran smoothly.

### DIFF
--- a/rpc/js/src/dial.d.ts
+++ b/rpc/js/src/dial.d.ts
@@ -15,7 +15,7 @@ export interface DialWebRTCOptions {
   signalingExternalAuthToEntity?: string;
   signalingCredentials?: Credentials;
   signalingAccessToken?: string;
-  priority?: number;
+  additionalSdpFields?: object;
 }
 export interface Credentials {
   type: string;
--- a/rpc/js/src/dial.d.ts
+++ b/rpc/js/src/dial.d.ts
@@ -15,6 +15,7 @@ export interface DialWebRTCOptions {
   signalingExternalAuthToEntity?: string;
   signalingCredentials?: Credentials;
   signalingAccessToken?: string;
+  priority?: number;
 }
 export interface Credentials {
   type: string;
--- a/rpc/js/src/dial.d.ts
+++ b/rpc/js/src/dial.d.ts
@@ -15,7 +15,7 @@ export interface DialWebRTCOptions {
   signalingExternalAuthToEntity?: string;
   signalingCredentials?: Credentials;
   signalingAccessToken?: string;
-  additionalSdpFields?: object;
+  additionalSdpFields?: Record<string, string | number>;
 }
 export interface Credentials {
   type: string;
--- a/rpc/js/src/dial.ts
+++ b/rpc/js/src/dial.ts
@@ -26,7 +26,7 @@ import {
   OptionalWebRTCConfigResponse,
 } from "./gen/proto/rpc/webrtc/v1/signaling_pb";
 import { SignalingService } from "./gen/proto/rpc/webrtc/v1/signaling_pb_service";
-import { addCustomSdpFields, newPeerConnectionForClient } from "./peer";
+import { addSdpFields, newPeerConnectionForClient } from "./peer";
 
 export interface DialOptions {
   authEntity?: string;
@@ -569,7 +569,7 @@ export async function dialWebRTC(
     client.start({ "rpc-host": host });
 
     const callRequest = new CallRequest();
-    const description = addCustomSdpFields(
+    const description = addSdpFields(
       pc.localDescription,
       opts.webrtcOptions?.additionalSdpFields
     );
--- a/rpc/js/src/dial.ts
+++ b/rpc/js/src/dial.ts
@@ -74,6 +74,10 @@ export interface DialWebRTCOptions {
   //
   // If enabled, other auth options have no affect. Eg. authEntity, credentials, signalingAuthEntity, signalingCredentials.
   signalingAccessToken?: string;
+
+  // `priority` is the priority of the current WebRTC connection. If only one WebRTC
+  // connection is allowed and another peer with a higher priority attempts to connect,
+  // the higher priority connection will unseat the lower priority connection.
   priority?: number;
 }
 
--- a/rpc/js/src/dial.ts
+++ b/rpc/js/src/dial.ts
@@ -74,6 +74,7 @@ export interface DialWebRTCOptions {
   //
   // If enabled, other auth options have no affect. Eg. authEntity, credentials, signalingAuthEntity, signalingCredentials.
   signalingAccessToken?: string;
+  priority?: number;
 }
 
 export interface Credentials {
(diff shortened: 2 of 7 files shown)


maximpertsov:
https://github.com/viamrobotics/viam-typescript-sdk/pull/181#issuecomment-1777730534

This comment actually applies to this PR more than the one I posted it on :slightly_smiling_face:

purplenicole730:
> [viamrobotics/viam-typescript-sdk#181 (comment)](https://github.com/viamrobotics/viam-typescript-sdk/pull/181#issuecomment-1777730534)
> 
> This comment actually applies to this PR more than the one I posted it on 🙂

Done!

rpc/js/src/peer.ts
```diff
@@ -60,9 +61,14 @@ export async function newPeerConnectionForClient(
...
         await peerConnection.setLocalDescription();
-        negotiationChannel.send(
-          btoa(JSON.stringify(peerConnection.localDescription))
-        );
+        let description = {
+          sdp: peerConnection.localDescription?.sdp,
+          type: peerConnection.localDescription?.type,
+        };
+        if (priority) {
+          description.sdp = description.sdp + `a=x-priority:${priority}\r\n`;
```
maximpertsov, line 69: same issue https://github.com/viamrobotics/goutils/pull/215/files#r1372023046

```suggestion
          description.sdp = [description.sdp, `a=x-priority:${opts.webrtcOptions?.priority}\r\n`].join("");
```

rpc/js/src/peer.ts
```diff
@@ -75,9 +81,14 @@ export async function newPeerConnectionForClient(
...
       await peerConnection.setLocalDescription();
-      negotiationChannel.send(
-        btoa(JSON.stringify(peerConnection.localDescription))
-      );
+      let description = {
+        sdp: peerConnection.localDescription?.sdp,
+        type: peerConnection.localDescription?.type,
+      };
+      if (priority) {
+        description.sdp = description.sdp + `a=x-priority:${priority}\r\n`;
```
maximpertsov, line 89: same issue https://github.com/viamrobotics/goutils/pull/215/files#r1372023046

```suggestion
        description.sdp = [description.sdp, `a=x-priority:${opts.webrtcOptions?.priority}\r\n`].join("");
```

maximpertsov (changes requested):
`undefined` + string concat issue, otherwise looks good

(it's not you, it's JavaScript)

maximpertsov (approved):
so sorry, I copy-pasted the same code suggestion for each case where we were appending, but the vars are slightly different. everything else looks good!

rpc/js/src/peer.ts
```diff
@@ -75,9 +81,14 @@ export async function newPeerConnectionForClient(
...
       await peerConnection.setLocalDescription();
-      negotiationChannel.send(
-        btoa(JSON.stringify(peerConnection.localDescription))
-      );
+      let description = {
+        sdp: peerConnection.localDescription?.sdp,
+        type: peerConnection.localDescription?.type,
+      };
+      if (priority) {
+        description.sdp = [description.sdp, `a=x-priority:${opts.webrtcOptions?.priority}\r\n`].join("");
```
maximpertsov, line 89: my apologies

```suggestion
        description.sdp = [description.sdp, `a=x-priority:${priority}\r\n`].join("");
```

rpc/js/src/peer.ts
```diff
@@ -60,9 +61,14 @@ export async function newPeerConnectionForClient(
...
         await peerConnection.setLocalDescription();
-        negotiationChannel.send(
-          btoa(JSON.stringify(peerConnection.localDescription))
-        );
+        let description = {
+          sdp: peerConnection.localDescription?.sdp,
+          type: peerConnection.localDescription?.type,
+        };
+        if (priority) {
+          description.sdp = [description.sdp, `a=x-priority:${opts.webrtcOptions?.priority}\r\n`].join("");
```
maximpertsov, line 69: my apologies

```suggestion
          description.sdp = [description.sdp, `a=x-priority:${priority}\r\n`].join("");
```

rpc/js/src/dial.d.ts
```diff
@@ -15,6 +15,7 @@ export interface DialWebRTCOptions {
   signalingExternalAuthToEntity?: string;
   signalingCredentials?: Credentials;
   signalingAccessToken?: string;
+  priority?: number;
```
cheukt, line 18: I don't think we should expose priority directly - we should have an `additionalSDPFields` or something (that is an object with key-value pairs) so that if we want to add future fields we don't have to make additional updates.

rpc/js/src/dial.ts
```diff
@@ -565,7 +571,17 @@ export async function dialWebRTC(
...
+    let description = {
+      sdp: pc.localDescription?.sdp,
+      type: pc.localDescription?.type,
+    };
+    if (opts.webrtcOptions?.priority) {
+      description.sdp = [
+        description.sdp,
+        `a=x-priority:${opts.webrtcOptions?.priority}\r\n`,
+      ].join("");
+    }
```
cheukt, line 583: should be factored out into a helper function

cheukt (changes requested):
looks pretty good! just want to make our additions as extensible as possible

rpc/js/src/peer.ts
```diff
@@ -3,9 +3,29 @@ interface ReadyPeer {
...
+) {
+  let description = {
+    sdp: localDescription?.sdp,
+    type: localDescription?.type,
+  };
+  if (sdpFields) {
+    Object.keys(sdpFields).forEach((key) => {
+      description.sdp = [
+        description.sdp,
+        `a=x-${key}:${sdpFields[key as keyof typeof sdpFields]}\r\n`,
```
cheukt, line 18: i don't think the x- should be automatically added

rpc/js/src/peer.d.ts
```diff
@@ -2,8 +2,16 @@ interface ReadyPeer {
   pc: RTCPeerConnection;
   dc: RTCDataChannel;
 }
+export declare function addCustomSdpFields(
+  sdpFields?: object,
+  localDescription?: RTCSessionDescription | null
```
cheukt, line 7: conventionally I think you would have localDescription first in args

rpc/js/src/peer.ts
```diff
@@ -3,9 +3,29 @@ interface ReadyPeer {
   dc: RTCDataChannel;
 }
 
+export function addCustomSdpFields(
+  sdpFields?: object,
+  localDescription?: RTCSessionDescription | null
+) {
+  let description = {
+    sdp: localDescription?.sdp,
```
cheukt, line 11: if localDesc is null, does this work?, what does `localDescription?.sdp` evaluate to?

cheukt (changes requested):
a few smaller nits

rpc/js/src/peer.ts
purplenicole730: If this is for custom sdp values, the `x-` is a standard to add in, according to https://www.ietf.org/rfc/rfc2327.txt.
> Unregistered attributes should begin with "X-" to
   prevent inadvertent collision with registered attributes.  In either
   case, if an attribute is received that is not understood, it should
   simply be ignored by the receiver.

rpc/js/src/peer.ts
purplenicole730: `localDescription?.sdp` would return undefined, which would then be `description.sdp = undefined`. The concatenation would still work as intended.

npmenard (approved):
LGTM

rpc/js/src/peer.ts
cheukt: ya but users of the utility may want to add attributes that are registered

purplenicole730: Ah, I thought we didn't want them to be able to change attributes until further notice. If that's the case, I will change the function name as well as get rid of the `x-`.

PR #216, opened 2023-10-25
APP-3031: make auth0.go generic to auth providers

ohEmily:
[APP-3031](https://viam.atlassian.net/browse/APP-3031)

([app PR](https://github.com/viamrobotics/app/pull/2784) that incorporates these changes)

Tiny PR that I believe (🤞) includes all the changes needed to `auth0.go` to make it so that we can cutover the `app` backend to FusionAuth without any further `goutils` changes.

I didn't rename `auth0.go` because it'll destroy the version history on this file if I do a squash and merge (only thing enabled for our repo afaik).

Tested that auth still works in `app` if I:
0) [local setup] add `replace go.viam.com/utils => ../goutils` and then `go mod tidy`
1) replace uses of Auth0Config with AuthProviderConfig in 2 places in `server/server.go`
```
diff --git a/server/server.go b/server/server.go
index 9b8a5643..29d00820 100644
--- a/server/server.go
+++ b/server/server.go
@@ -299,7 +299,7 @@ type simpleServer struct {
 	unauthenticatedMux    *goji.Mux
 	grpcServer            rpc.Server
 	auth0State            io.Closer
-	auth0Config           uweb.Auth0Config
+	auth0Config           uweb.AuthProviderConfig
 	webTokenKeyProvider   jwks.KeyProvider
 	jWKSKeyProvider       rpc.TokenVerificationKeyProvider
 	sessions              *uweb.SessionManager
```
and
```
@@ -640,7 +640,7 @@ func (ss *simpleServer) setupDependencies(ctx context.Context) error {
 		return errors.Wrap(err, "error unmarshaling auth0 data")
 	}

-	ss.auth0Config = uweb.Auth0Config{
+	ss.auth0Config = uweb.AuthProviderConfig{
 		Domain:     m["domain"],
 		ClientID:   m["clientId"],
 		Secret:     m["secret"],
```

I'll update `goutils` in app with the changes above once this is ready to merge.

[APP-3031]: https://viam.atlassian.net/browse/APP-3031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mcous:
Bringing in the council of Michaels on this one

ohEmily:
One thing I'm not clear on is whether I broke the automatic call to `Close()` by futzing around with utilization of the `io.Closer` interface. Any ideas? Could produce some resource leaks.

michaellee1019 (approved):
LGTM, some naming suggestions.

web/auth0.go
```diff
@@ -23,30 +23,43 @@ import (
 	"go.viam.com/utils"
 )
 
-// Auth0Config config for auth0.
-type Auth0Config struct {
+// AuthProviderConfig config options with constants that will probably need to be manually configured after
```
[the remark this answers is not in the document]
ohEmily: I didn't want to rename as part of this PR as it would destroy the commit history on the squash and merge. I think I should do a `git mv` as a follow up

mcous (dismissed):
LGTM

michaellee1019:
> One thing I'm not clear on is whether I broke the automatic call to `Close()` by futzing around with utilization of the `io.Closer` interface. Any ideas? Could produce some resource leaks.

As long as the app side still calls Close() like it does here: https://github.com/viamrobotics/app/blob/main/server/server.go#L1691, then I don't think there is anything to be concerned about. Its just refactoring and the Close call will be the same.

web/auth0.go
```diff
@@ -23,30 +23,43 @@ import (
...
+type AuthProvider struct {
+	io.Closer
+
+	config   AuthProviderConfig
 	sessions *SessionManager
 
-	authOIConfig       *oidc.Config
-	authConfig         oauth2.Config
-	auth0HTTPTransport *http.Transport
+	authOIConfig  *oidc.Config
```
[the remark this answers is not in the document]
ohEmily: pushed this change

ohEmily:
Finally gearing up to merge this since we moved forward the testing and doesn't look like we'll need more app changes?

web/auth0.go
```diff
@@ -92,57 +167,80 @@ func InstallAuth0(
...
+
+	// http.SetCookie(w, &http.Cookie{
+	// 	Name:     h.redirectStateCookieName,
+	// 	Value:    "",
+	// 	Path:     "/",
+	// 	MaxAge:   -1,
+	// 	Secure:   r.TLS != nil,
+	// 	SameSite: http.SameSiteLaxMode,
+	// 	HttpOnly: true,
+	// })
```
ohEmily, line 242: [q] Devin, so this should be turned off for FusionAuth only? Or for both Auth0 and FusionAuth?

web/auth0.go
```diff
@@ -185,6 +282,36 @@ func (h *callbackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	http.SetCookie(w, &http.Cookie{
+		Name:     "viam.auth.expiry",
+		Value:    token.Expiry.Format(time.RFC3339),
+		Path:     "/",
+		Expires:  token.Expiry,
+		Secure:   r.TLS != nil,
+		SameSite: http.SameSiteLaxMode,
+		HttpOnly: true,
+	})
```
ohEmily, line 313: @DTCurrie same here -- is there any harm in this being sent for Auth0 as well as FusionAuth? Trying to split this out into different auth providers.

DTCurrie: I didn’t actually comment out that first cookie originally, but so don’t think there is any harm in uncommenting it our now. It shouldn’t interfere with what I added.

The changes I made in app don’t really care about auth0 or fusion auth, so the added logic should conceivably work for both. Basically we set temporary cookies with the access token and it’s expiry, then on the front end we call the new token endpoint I added to return those values and clear the temp cookies. Then the front end caches those in local storage and uses them from there. If the expiry time has passed, it sends the user to log back in.

ohEmily: thanks for the lightning fast response!

PR #217, opened 2023-11-03
RSDK-5308-1 - standardize-prettier-config

purplenicole730:


rpc/js/.prettierrc.cjs
```diff
@@ -0,0 +1,2 @@
+// .prettierrc.cjs
```
mcous, line 1: This filename comment was just for the README in the config repo, it doesn't have to be here 😄. Since it's a CommonJS module, we _should_ include `use strict;`, though, or else our linter will yell at us later.

```suggestion
'use strict';
```

(The "use strict;" incantation is an old silly thing from darker times in JS. Add it if the linter ever tells you to, but don't worry about it otherwise because it's on its way out)

rpc/js/package.json
```diff
@@ -8,7 +8,8 @@
   },
   "devDependencies": {
     "@types/google-protobuf": "^3.7.4",
-    "prettier": "^2.8.8",
+    "@viamrobotics/prettier-config": "^0.3.4",
+    "prettier": "^3.0.3",
```
mcous, line 12: There's a `format` script below that has `{ts,js}`. We'll want to add `cjs` to that list since we've added `.prettierrc.cjs`

mcous (approved):

PR #218, opened 2023-11-06
RSDK-5308-2 - standardize type checking

purplenicole730:
Tested the output using the [procedure](https://github.com/viamrobotics/goutils/pull/218#issuecomment-1799522286) provided by @mcous below, and have made the necessary changes so that there are no differences in the outputted `main.cjs` file.

Ran the echo example as well just for further testing


mcous:
I'm working on a test procedure for this, and it's a bit of a slog because of how out of date this project's tooling is.

1. Manually modify `rpc/js/webpack.config.js` to disable source maps and minification
```diff
- devtool: 'inline-source-map',
+ devtool: false,
+ optimization: { minimize: false },
```
2. Run `make build-web` on `main`
3. Copy `rpc/js/dist/main.mjs` somewhere
4. Checkout this branch
5. Make the same `webpack.onfig.js` change
6. Run `make build-web` on this branch
7. Copy `rpc/js/dist/main.mjs` next to the copy from `main`
8. Diff the files

I'm getting a fairly substantial diff, which makes sense given we've upgraded TypeScript and this project uses the TypeScript compiler (via `ts-loader` to generate its bundle)

```shell
 main.mjs | 1486 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------------------------------------------------------------------------------------------------
 1 file changed, 482 insertions(+), 1004 deletions(-)
```

I'm combing through it, but my initial instincts are that we've probably removed a little more TS config than we want to in the switch. Will post updates here as I work through it

rpc/js/tsconfig.json
```diff
@@ -1,46 +1,12 @@
 {
+  "extends": "@viamrobotics/typescript-config/base.json",
   "compilerOptions": {
     "target": "ES5",
-    "useDefineForClassFields": true,
```
mcous, line 4: Discussed in Slack, looks like we need to keep this line

rpc/js/tsconfig.json
```diff
@@ -1,46 +1,12 @@
...
-    "declaration": true,
-    "strict": true,
-    "importHelpers": true,
```
mcous, line 7: Ditto

mcous (commented):
From our Slack converations

mcous (approved):
LGTM, don't see any difference in compiled output and CI is green

maximpertsov (approved):
We might want to run [the echo example](https://github.com/viamrobotics/goutils/tree/main/rpc/examples/echo) to ensure basic functionality still works, but I'll leave this as optional since the [diff no longer has any changes](https://github.com/viamrobotics/goutils/pull/218#pullrequestreview-1721198495).

PR #219, opened 2023-11-08
RSDK-5308-3 - add eslint configuration

purplenicole730:


maximpertsov (changes requested):
we don't actually run `eslint` in CI yet so we would need to add that

rpc/js/package.json
```diff
@@ -25,6 +32,7 @@
   "scripts": {
     "prepare": "./etc/prepare.sh",
     "typecheck": "tsc --noEmit",
+    "lint": "eslint '.*.cjs' '**/*.{ts,js,cjs}' --quiet",
```
mcous, line 35: I would recommend against using `--quiet` unless we need to; better to start off noisy than not noisy enough

rpc/js/.eslintrc.cjs
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+module.exports = {
+  root: true,
+  extends: ['@viamrobotics/eslint-config'],
+  parserOptions: {
+    project: ['./tsconfig.json'],
+  },
```
mcous, line 8: Since we're in a large, multilevel repo with a few package.json files, we've found things to be a little more consistent if we include the `tsconfigRootDir: __dirname` setting

```suggestion
  parserOptions: {
    tsconfigRootDir: __dirname,
    project: './tsconfig.json',
  },
```

Without it, ESLint can get a little confused about where to find the TSConfig

mcous (commented):
A couple quick thoughts, in advance of CI config

mcous:
@purplenicole730 ok did a little more testing:

- We'll need an `.eslintignore` file that excludes `src/gen` and probably also those silly `src/*.d.ts` files
- We'll also likely need to split apart out `tsconfig.json` into two: one for linting and typechecking and another for building
    - Follow the pattern in typescript-sdk of `tsconfig.json` including everything, and `tsconfig.build.json` only including the source files we want to build
    - The `webpack.config.js` then needs to be updated to use `tsconfig.build.json`
- This seems like too many errors to fix in this PR; **ignore my comment above about `--quiet`**
    - Add overrides in `.eslintrc.cjs` to downgrade everything that's erroring to a `warn`
    - Then, come back to them in another PR or series of PRs
    - It's very likely that actual bugs are going to be uncovered here, so it's probably best to tackle bugs in isolation without lumping them in with a bunch of style changes
 
> we don't actually run eslint in CI yet so we would need to add that

@maximpertsov looks like [it's running](https://github.com/viamrobotics/goutils/actions/runs/6804278716/job/18501443031?pr=219) in CI to me!

maximpertsov:
> looks like [it's running](https://github.com/viamrobotics/goutils/actions/runs/6804278716/job/18501443031?pr=219) in CI to me!

Yeah we just added it https://github.com/viamrobotics/goutils/pull/219/commits/fd3d269e6af8399baca1cb20008d62688f91a449 :)

maximpertsov:
> Add overrides in .eslintrc.cjs to downgrade everything that's erroring to a warn

cool - agree with that approach, basically mirrors how we're proceeding in the TS SDK

rpc/js/.eslintrc.cjs
```diff
@@ -0,0 +1,71 @@
...
+    'unicorn/prefer-ternary': 'warn',
+    'unicorn/custom-error-definition': 'warn',
+    'unicorn/no-useless-promise-resolve-reject': 'warn',
+    'prefer-const': 'warn',
+    'no-plusplus': 'warn',
+    'id-length': 'warn',
+    'class-methods-use-this': 'warn',
+    'no-bitwise': 'warn',
+    'guard-for-in': 'warn',
+    eqeqeq: 'warn',
```
maximpertsov, line 33: gosh, we don't use `===` somewhere?

maximpertsov (approved):
great! lots to fix but I'm happy to do in subsequent PRs

mcous (approved):
Reviewed in person, LGTM

PR #221, opened 2023-11-13
RSDK-851 rpc js tests

maximpertsov:
Run echo example in CI so we have some JS test coverage. 

* Note that the CI run below uses the workflow definition on `main` and doesn't include the new JS test. I ran the JS tests in a separate action: https://github.com/viamrobotics/goutils/actions/runs/7051309320

Also add a minimal UI to the example instead of just logging to the console:

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,7 +82,7 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: playwright-report
-        path: rpc/examples/echo/playwright-report/
+        path: $GITHUB_WORKSPACE/rpc/examples/echo/playwright-report/
         retention-days: 30
 
     - name: Test go library
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,10 +79,10 @@ jobs:
 
     - if: failure()
       name: Upload js test report
-      uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce #v3.1.2
+      uses: actions/upload-artifact@v3
       with:
         name: playwright-report
-        path: rpc/examples/echo/playwright-report/
+        path: ${{ github.workspace}}/rpc/examples/echo/playwright-report
         retention-days: 30
 
     - name: Test go library
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,28 +74,36 @@ jobs:
             exit 1
         fi
 
-    - name: Test web client
+    - name: Test js library
       run: sudo -u testbot bash -lc 'make test-web'
-    #
-    # - name: Test
-    #   env:
-    #     TEST_MONGODB_URI: ${{ secrets.TEST_MONGODB_URI }}
-    #     MONGODB_TEST_OUTPUT_URI: ${{ secrets.MONGODB_TEST_OUTPUT_URI }}
-    #   run: |
-    #     sudo -u testbot bash -lc 'echo "${{ secrets.ARTIFACT_GOOGLE_APPLICATION_CREDENTIALS }}" | base64 -d > artifact_google_creds.json'
-    #     export ARTIFACT_GOOGLE_APPLICATION_CREDENTIALS=`pwd`/artifact_google_creds.json
-    #     sudo -u testbot --preserve-env=ARTIFACT_GOOGLE_APPLICATION_CREDENTIALS,TEST_MONGODB_URI,MONGODB_TEST_OUTPUT_URI,GITHUB_SHA,GITHUB_RUN_ID,GITHUB_RUN_NUMBER,GITHUB_RUN_ATTEMPT,GITHUB_X_PR_BASE_SHA,GITHUB_X_PR_BASE_REF,GITHUB_X_HEAD_REF,GITHUB_X_HEAD_SHA,GITHUB_REPOSITORY bash -lc 'make cover'
-    #
-    # # this one runs as root for CAP_SETUID
-    # - name: test run-as-user
-    #   env:
-    #     TEST_SUBPROC_USER: testbot
-    #   run: go test -v ./pexec -run TestManagedProcessStart
-    #
-    # - name: Add Coverage PR Comment
-    #   uses: marocchino/sticky-pull-request-comment@v2.2.0
-    #   if: github.event_name == 'pull_request'
-    #   with:
-    #     recreate: true
-    #     path: code-coverage-results.md
-    #     GITHUB_TOKEN: ${{ secrets.GIT_ACCESS_TOKEN }}
+
+    - name: Test go library
+      env:
+        TEST_MONGODB_URI: ${{ secrets.TEST_MONGODB_URI }}
+        MONGODB_TEST_OUTPUT_URI: ${{ secrets.MONGODB_TEST_OUTPUT_URI }}
+      run: |
+        sudo -u testbot bash -lc 'echo "${{ secrets.ARTIFACT_GOOGLE_APPLICATION_CREDENTIALS }}" | base64 -d > artifact_google_creds.json'
+        export ARTIFACT_GOOGLE_APPLICATION_CREDENTIALS=`pwd`/artifact_google_creds.json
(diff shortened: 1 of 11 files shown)


rpc/examples/echo/playwright.config.ts
```diff
@@ -0,0 +1,78 @@
...
+      name: "chromium",
+      use: { ...devices["Desktop Chrome"] },
+    },
+
+    {
+      name: "firefox",
+      use: { ...devices["Desktop Firefox"] },
+    },
+
+    // TODO: resolve Safari tests
```
maximpertsov, line 45:
```suggestion
    // TODO(RSDK-4494): fix safari tests
```

mcous (commented):
Played around with the locally! Got some thoughts and ideas

rpc/examples/echo/playwright.config.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: "./tests",
+  outputDir: "/tmp/.playwright-test",
```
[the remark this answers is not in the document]
maximpertsov: yeah, not sure why this got in here - removing

Makefile
```diff
@@ -59,9 +59,14 @@ lint-go: tool-install
 cover: tool-install
 	PATH=$(PATH_WITH_TOOLS) ./etc/test.bash cover
 
-test: tool-install
+test: test-go test-web
+
+test-go: tool-install
 	PATH=$(PATH_WITH_TOOLS) ./etc/test.bash
 
+test-web:
```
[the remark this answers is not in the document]
maximpertsov: good call, let me see if I can just make this a dependency

rpc/examples/echo/tests/echo.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  ];
+
+  for (const [testid, expected] of table) {
+    const actual = await page
+      .getByTestId(testid)
+      .evaluate((elem) =>
+        Array.from(elem.querySelectorAll("div")).map((div) => div.innerText),
+      );
+    PW.expect(actual).toStrictEqual(expected);
+  }
```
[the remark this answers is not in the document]
maximpertsov: hm, this test is flaky now with firefox - we are gathering the results before they are ready I think. Maybe adding the extra testid as you suggested and waiting for those elements to appear is the way to go.

.github/workflows/test.yml
```diff
@@ -74,7 +74,10 @@ jobs:
             exit 1
         fi
 
-    - name: Test
+    - name: Test js library
+      run: sudo -u testbot bash -lc 'make test-web'
+
```
mcous, line 79: Forgot to mention: we should send the Playwright test results to a GH Actions artifact, like we do in app and RDK. This lets us view traces locally and diagnose test failures (and flakiness) more easily

rpc/examples/echo/tests/echo.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  ];
+
+  for (const [testID, expected] of table) {
+    await expect(page.getByTestId(testID).getByTestId("message")).toHaveCount(
+      expected.length,
+    );
+    const messages = page.getByTestId(testID).getByTestId("message");
+    const contents = await messages.allTextContents();
+
+    expect(contents).toStrictEqual(expected);
```
mcous, line 23: Locaters in Playwright are lazy and stateless. They're a description of an element search, and they don't evaluate to an actual element until used inside an `expect` call (or if you use a method like `await locator.allTextContents()`).

Accordingly, it makes sense to assign locators only once and re-use them. In general, I also find splitting locator creation away from calling `expect` makes Playwright tests much easier to read.

Finally, this `toHaveCount` check is redundant with `expect(contents).toStrictEqual(expected);` below. If count of elements is off, then `contents` will fail the call to `toStrictEqual`, because it will have different length

```suggestion
    const messages = page.getByTestId(testID).getByTestId("message");
    const contents = await messages.allTextContents();

    expect(contents).toStrictEqual(expected);
```

maximpertsov: Hm, the suggested implementation is flaky, at least for Firefox. It seems like `allTextContents()` might validly return nothing if we reach the `expect` call before the elements are loaded.

mcous: Oh oops, my bad, splitting out `allTextContents` from the `expect` will cause flakiness. You're right, we should check the count. We can then lean on `toContainText` to check contents.

This works for me locally:

```suggestion
    const messages = page.getByTestId(testID).getByTestId("message");

    await expect(messages).toHaveCount(expected.length)
    await expect(messages).toContainText(expected);
```

maximpertsov: sweet, confirmed that this is stable now

Makefile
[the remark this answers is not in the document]
maximpertsov: I'm torn... kinda want to just make a dependency, but it does make this step slower (+20s). I'll just note the one-time step in the readme

maximpertsov:
@mcous the CI run that includes Playwright tests and all review changes is here: https://github.com/viamrobotics/goutils/actions/runs/7091655187

Makefile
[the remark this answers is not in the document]
maximpertsov: https://github.com/viamrobotics/goutils/pull/221/commits/50543ef3b4df2e3aedf1c8e9790a05223c331469

rpc/examples/echo/tests/echo.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    ["multi-direct", ["h", "e", "l", "l", "o", "?"]],
+    // gRPC-web does not yet support bidirectional streaming so we expect to
+    // only receive a response to our first request.
+    ["bidi-direct", ["o", "n", "e"]],
+  ];
+
+  for (const [testID, expected] of table) {
+    const messages = page.getByTestId(testID).getByTestId("message");
+    // TODO: this test fails on purpose
+    await expect(messages).toHaveCount(expected.length + 1);
```
mcous, line 19: Revert before merge
[Commenting for safety, I know this is your intent]

mcous (approved):
✨

rpc/examples/echo/playwright.config.ts
```diff
@@ -0,0 +1,75 @@
...
+export default defineConfig({
+  testDir: "./tests",
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 1 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : 3,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: process.env.CI ? "dot" : "html",
```
mcous, line 20: You might need to change this to `html` to have a test report to upload as an artifact; `dot` might not save anything

maximpertsov: that seems to be the case - however, setting the report to `html` seems cause the step test to hang

maximpertsov: figured it out, I needed to configure the `html` report to not open: https://github.com/viamrobotics/goutils/pull/221/commits/97d4f6a63301fbeedfd4a153159f755724cb30a7 - curious why we don't have to do this in app tho

rpc/examples/echo/playwright.config.ts
```diff
@@ -17,7 +17,7 @@ export default defineConfig({
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : 3,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: process.env.CI ? "dot" : "html",
+  reporter: [["html", { open: "never" }]],
```
mcous, line 20: Weird, we definitely don't need to use this option in app. I wonder what's different here?

mcous: On the plus side, I find auto-open annoying personally, so I don't hate this!

PR #222, opened 2023-11-20
Bump @viamrobotics/rpc to 0.1.39

ethanlook:
Bumping the version here so we can get the typings issues that were fixed as a transient dependency in `@viamrobotics/sdk` and ultimately `app`.

See my PR for the downstream issue: https://github.com/viamrobotics/app/actions/runs/6895151896/job/18758485918?pr=2866


mcous (approved):
Reviewed the recent changes as they happened, so this bump should be safe 🤞

PR #223, opened 2023-11-29
Rsdk 5407 build goutils rpc with vite

purplenicole730:
Plan: Ship this and then test on TS SDK live (and then ship 0.2.1 if needed)


rpc/js/src/index.ts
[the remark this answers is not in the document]
micheal-parks: renamed to `main.ts` to match the TS SDK

rpc/js/tsconfig.json
```diff
@@ -4,11 +4,14 @@
     "target": "ES5",
     "useDefineForClassFields": true,
     "importHelpers": true,
-    "module": "commonjs",
+    "module": "ESNext",
     "moduleResolution": "Node",
     "verbatimModuleSyntax": false,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "dist",
```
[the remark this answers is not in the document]
purplenicole730: What would the value for `rootDir` be?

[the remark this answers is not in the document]
mcous: `src`

rpc/js/vite.config.ts
```diff
@@ -0,0 +1,42 @@
+/// <reference types="vitest" />
```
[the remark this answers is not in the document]
mcous: Comment still applies. Also, do we intend to unit tests with `vitest`? If so, it's currently missing from our `package.json`s.

If we do not intend to add `vitest` right now, then this line and the `test` field of the config should both be removed here and in other configs

rpc/js/vite.config.ts
```diff
@@ -0,0 +1,42 @@
...
+    commonjsOptions: {
+      transformMixedEsModules: true,
+      include: [
+        /exponential-backoff/u,
+        /google-protobuf/u,
+        /@improbable-eng\/grpc-web/u,
+        /gen\//u,
+      ],
+    },
+    minify: true,
```
mcous, line 23: I don't think this option is required; `vite` will [minify by default](https://vitejs.dev/config/build-options.html#build-minify)

rpc/js/vite.config.ts
```diff
@@ -0,0 +1,42 @@
...
+    target: 'esnext',
```
mcous, line 24: I feel a little weird about changing this value away from [its default](https://vitejs.dev/config/build-options.html#build-target); is it required?

rpc/js/tsconfig.json
```diff
@@ -4,11 +4,15 @@
...
+    "module": "ESNext",
+    "moduleResolution": "bundler",
     "verbatimModuleSyntax": false,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "dist",
     "lib": ["ESNext", "DOM", "DOM.iterable"]
   },
-  "include": ["src", ".*.cjs", "*.js"],
+  "types": ["vitest"],
```
mcous, line 15: This won't work unless `vitest` is added as a dependency to the project

rpc/js/.eslintignore
```diff
@@ -1,2 +1,5 @@
+.eslintrc.cjs
+.prettierrc.cjs
+dist/*
 src/gen
-src/*.d.ts
\ No newline at end of file
+vite.config.ts
```
mcous, line 5: This file should not be lint ignored

rpc/js/.npmignore
```diff
@@ -1,4 +1,3 @@
 etc
 node_modules
-tsconfig.json
-webpack.config.js
+tsconfig.json
```
mcous, line 3: Rather than maintain a `.npmignore` file, it's usually easier to specify an allowlist via the `files` array in package.json

rpc/js/package.json
```diff
@@ -23,18 +23,22 @@
     "ts-protoc-gen": "^0.14.0",
     "tslib": "^2.4.1",
     "typescript": "^5.2.2",
-    "vite": "^3.1.4",
-    "webpack": "^5.74.0",
-    "webpack-cli": "^4.10.0"
+    "vite": "^5.0.6"
```
mcous, line 26: We haven't yet migrated other projects to Vite v5. If we don't have any issues here, great! But it may be prudent to stick to Vite v4 if we run into weird stuff, especially given how much annoying CJS stuff we have to deal with in this project

rpc/js/package.json
```diff
@@ -23,18 +23,22 @@
...
-  "main": "dist/main.mjs",
-  "types": "src/index.d.ts",
   "scripts": {
     "prepare": "./etc/prepare.sh",
     "typecheck": "tsc --noEmit",
-    "lint": "eslint '.*.cjs' '**/*.{ts,js,cjs}' --quiet",
+    "lint": "eslint '**/*.{ts,js,cjs}' --quiet",
     "format": "prettier --write '**/*.{ts,js,cjs}'",
-    "build": "npx webpack",
+    "build": "vite build && tsc --emitDeclarationOnly",
```
mcous, line 41: Looks like `--emitDeclarationOnly` is redundant with the `tsconfig.json` setting

rpc/js/vite.config.ts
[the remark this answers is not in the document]
purplenicole730: Thanks for pointing that out. This was just copied and pasted from the TS SDK, but I don't see any unit tests with `vitest` in goutils

rpc/js/vite.config.ts
purplenicole730:
```nicolejung@Nicoles-MacBook-Air js % npm run build

> @viamrobotics/rpc@0.2.0 build
> vite build && tsc --emitDeclarationOnly

vite v5.0.6 building for production...
✓ 50 modules transformed.
dist/main.js  660.84 kB │ gzip: 86.49 kB
✓ built in 668ms
nicolejung@Nicoles-MacBook-Air js % npm run build

> @viamrobotics/rpc@0.2.0 build
> vite build && tsc --emitDeclarationOnly

vite v5.0.6 building for production...
✓ 50 modules transformed.
dist/main.js  659.53 kB │ gzip: 86.16 kB
✓ built in 640ms```

The first one was built without the `target` value, while the second one had `target` set to `esnext`. It seems it helped with both speed and space.

File: rpc/js/.eslintignore [hunk 10]
```diff
@@ -1,2 +1,4 @@
-src/gen
-src/*.d.ts
\ No newline at end of file
+.eslintrc.cjs
+.prettierrc.cjs
```
mcous, line 2: The linting configuration files need to be linted, don't ignore them here

rpc/js/vite.config.ts
mcous: It's expected that it would be slightly smaller with `esnext`, but I also think it will be less compatible. It may even _require_ further transpilation to actually be usable, which IMO would be undesirable in a published module.

- If we have a concrete reason for deviating from default, we should have a comment explaining why
- If we _don't_ have a concrete reason, we should remove the option to avoid encoding useless superstition

rpc/js/vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+  build: {
+    commonjsOptions: {
+      transformMixedEsModules: true,
+      include: [
+        /exponential-backoff/u,
+        /google-protobuf/u,
+        /@improbable-eng\/grpc-web/u,
+        /gen\//u,
+      ],
+    },
```
mcous, line 20: Can you add a comment explaining why this section is needed?

purplenicole730: @micheal-parks We copied and pasted this from the TS SDK. Do you have any insight as to why this is necessary?

rpc/js/vite.config.ts
micheal-parks: Agreed that the default should be fine here

rpc/js/vite.config.ts
micheal-parks: The libraries on the list have poor Vite compatibility and adding this config is necessary to transform them into ES modules. We should add a note that this can be removed when we migrate to protobuf-es or a code generating tool that has good support for ES modules.

maximpertsov: will defer to @micheal-parks to give more concrete details, but this option was necessary to [drastically reduce the bundle size of the typescript sdk](https://github.com/viamrobotics/viam-typescript-sdk/pull/41). I'm assuming we also need a version of this config in this package since we doing CommonJS > ESM conversion.

At minimum the `include` list is wrong, we can remove `/exponential-backoff/u,` since that package is only used in TS-SDK. I think the rest apply.

rpc/examples/echo/frontend/vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+export default defineConfig({
+  define: {
+    "process.env.NODE_ENV": '"production"',
+    __VERSION__: JSON.stringify(pkg.version),
+  },
+  build: {
+    commonjsOptions: {
+      transformMixedEsModules: true,
+      include: [
+        /exponential-backoff/u,
```
maximpertsov, line 15: per my comment https://github.com/viamrobotics/goutils/pull/223#discussion_r1427000432

```suggestion
```

rpc/js/vite.config.ts
micheal-parks: Oh yeah! Forgot that I did that lol, thanks for finding. Yeah, as far as I recall without transforming to ES modules those libraries are not at all tree shakable.

purplenicole730: Added a comment that says this: 
```
    // This config is necessary to transform libraries on the list into ES modules.
    // This can be removed if protobuf-es or a code generating tool that has good
    // support for ES modules is used.
```
Should I also include the bundle size reduction comment?

maximpertsov: I think that's sufficient

rpc/examples/echo/frontend/vite.config.ts
```diff
@@ -0,0 +1,35 @@
...
+  },
+  build: {
+    commonjsOptions: {
+      transformMixedEsModules: true,
+      include: [
+        /google-protobuf/u,
+        /@improbable-eng\/grpc-web/u,
+        /gen\//u,
+      ],
+    },
```
mcous, line 19: Copy over comment about CJS etc.

rpc/examples/echo/server/templates/index.html
```diff
@@ -21,6 +20,7 @@
       padding: 0.5rem;
     }
   </style>
+    <script src="./static/main.js"></script>
```
mcous, line 23: [Nit] indentation seems inconsistent here

rpc/examples/fileupload/frontend/vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+import pkg from "./package.json";
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  define: {
+    "process.env.NODE_ENV": '"production"',
+    __VERSION__: JSON.stringify(pkg.version),
+  },
+  build: {
+    commonjsOptions: {
```
mcous, line 12: CJS comment

rpc/examples/fileupload/frontend/vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+      transformMixedEsModules: true,
+      include: [
+        /exponential-backoff/u,
```
mcous, line 15: Is this the package that's not used here?

rpc/js/.eslintignore
```diff
@@ -1,2 +1,2 @@
-src/gen
-src/*.d.ts
\ No newline at end of file
+dist/*
```
mcous, line 1: [Nit] The `*` is unnecessary

```suggestion
dist
```

rpc/js/tsconfig.json
```diff
@@ -4,11 +4,15 @@
     "target": "ES5",
     "useDefineForClassFields": true,
     "importHelpers": true,
-    "module": "commonjs",
-    "moduleResolution": "Node",
+    "module": "ESNext",
+    "moduleResolution": "node",
     "verbatimModuleSyntax": false,
+    "declaration": true,
+    "emitDeclarationOnly": true,
```
mcous, line 11: [Nit] These options are only for the build, so should be moved to `tsconfig.build.json`

rpc/js/tsconfig.json
```diff
@@ -4,11 +4,15 @@
...
+    "rootDir": "src",
+    "outDir": "dist",
```
mcous, line 13: Same comment: these are only used for build, so should be moved to `tsconfig.build.json`

mcous (approved):
LGTM

PR #224, opened 2023-11-30
RSDK-5906: Upload CI test.json file as a github artifact.

dgottlieb:
Cribbed from [the rdk](https://github.com/viamrobotics/rdk/blob/e92959d049e72b0ed57c1dcb9a2328ee5160d802/.github/workflows/test.yml#L84-L90).

Claim that `test.json` is the file to upload:
* The [test command](https://github.com/viamrobotics/goutils/blob/39728ac2cfd339fc555eb547ca71e4ab59bf05bb/.github/workflows/test.yml#L77-L84) calls `make cover`
* [`make cover` calls `etc/test.bash`](https://github.com/viamrobotics/goutils/blob/39728ac2cfd339fc555eb547ca71e4ab59bf05bb/Makefile#L59-L60)
* [`test.bash` calls `gotestsum` with `--jsonfile json.log`](https://github.com/viamrobotics/goutils/blob/39728ac2cfd339fc555eb547ca71e4ab59bf05bb/etc/test.bash#L12)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,6 +83,14 @@ jobs:
         export ARTIFACT_GOOGLE_APPLICATION_CREDENTIALS=`pwd`/artifact_google_creds.json
         sudo -u testbot --preserve-env=ARTIFACT_GOOGLE_APPLICATION_CREDENTIALS,TEST_MONGODB_URI,MONGODB_TEST_OUTPUT_URI,GITHUB_SHA,GITHUB_RUN_ID,GITHUB_RUN_NUMBER,GITHUB_RUN_ATTEMPT,GITHUB_X_PR_BASE_SHA,GITHUB_X_PR_BASE_REF,GITHUB_X_HEAD_REF,GITHUB_X_HEAD_SHA,GITHUB_REPOSITORY bash -lc 'make cover'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -83,6 +83,14 @@ jobs:
...
 
+    - name: Upload test.json
+      if: always()
+      uses: actions/upload-artifact@v3
+      with:
+        name: test.json
+        path: json.log
+        retention-days: 30
+
     # this one runs as root for CAP_SETUID
```
dgottlieb, line 94: Not sure what to do about this. For now I'm fine if we just omit these results. Unless it becomes particularly noisy.

As far as ideas -- can I assume the same machine is running this? And that `json.log` already exists? Can I just change `run` to append to `json.log`. E.g:
```
    - name: test run-as-user
      env:
        TEST_SUBPROC_USER: testbot
      run: go test -v ./pexec -run TestManagedProcessStart >> json.log
```

abe-winter: that cap_setuid test is mine -- fine to omit, fine to `>>` append, your call

(steps in the same job run in the same machine)

abe-winter (approved):

PR #228, opened 2023-12-27
APP-2747 Add logging to send for debugging

vijayvuyyuru:
Hey yall, this PR is to add logging to the send method to determine what case it's hitting when it returns false. For now I just included the whole event with the intention of backing this out once we get to the root of this issue. I also added a log for how many answer channels there are as I realized in only logging on the warning, we don't know if only having 1 answer channel is normal or not.

### DIFF
--- a/rpc/wrtc_call_queue_mongodb.go
+++ b/rpc/wrtc_call_queue_mongodb.go
@@ -16,6 +16,7 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 	"go.uber.org/multierr"
+	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
@@ -300,7 +301,7 @@ type mongodbNewCallEventHandler struct {
 	receiveOnce sync.Once
 }
 
-func (newCall *mongodbNewCallEventHandler) Send(event mongodbCallEvent) bool {
+func (newCall *mongodbNewCallEventHandler) Send(event mongodbCallEvent, logger *zap.SugaredLogger) bool {
 	var sent bool
 	newCall.receiveOnce.Do(func() {
 		// should always work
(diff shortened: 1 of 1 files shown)


rpc/wrtc_call_queue_mongodb.go
```diff
@@ -607,10 +613,11 @@ func (queue *mongoDBWebRTCCallQueue) processNextSubscriptionEvent(next mongoutil
 				// though, we want to send the events as fast as possible as mentioned
 				// above and cannot block on the send to see if the receiver locked
 				// the document.
-				if answerChan.Send(event) {
+				if answerChan.Send(event, queue.logger) {
 					return
 				}
 			}
+			queue.logger.Infof("Number of answer channels: %d", len(answerChans))
```
benjirewis, line 620: Necessary? Looks like this is logged in the warn below.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -300,16 +301,21 @@ type mongodbNewCallEventHandler struct {
...
 		case newCall.eventChan <- event:
 			sent = true
 		default:
+			logger.Infow("Hit default select in send",
+				"Event", event.Call)
 		}
 	})
+	logger.Infow("returning from send",
+		"Event", event.Call,
+		"Sent", sent)
```
benjirewis, line 318: Yeah these seem good; do you care that this might flood logs? I imagine this two info logs will be hit a lot.

benjirewis (approved):

rpc/wrtc_call_queue_mongodb.go
benjirewis: Oops just saw your reference to this in the PR description, could you move this above the for loop maybe? If I understand correctly, the happy path would not get to this point (we'd `return` on L616 since we were able to send to some `answerChan`.

vijayvuyyuru: Oops you're right, will move then merge!

rpc/wrtc_call_queue_mongodb.go
vijayvuyyuru: I imagine these are temporary, I'm hoping in the next week or sooner I can back them out, so it might temporarily but I think that's fine.

PR #229, opened 2024-01-03
More comments in the signalling code

RoxyFarhad:


vijayvuyyuru (approved):
These changes look good and will definetly help in the future if someone else starts looking in here. I can publish a new release after the metric change is merged in! Nice work!

ohEmily:
[minor] could you add your name to your comments to follow Eric's convention? I know people can use git blame but sometimes these are harder to follow.
```
// Note(erd): 
```

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -422,14 +424,13 @@ func (queue *mongoDBWebRTCCallQueue) changeStreamManager() {
 		}
 	}()
 	var lastSeq uint64
-	var ranOnce bool
+	var ranOnce bool // this is only for the first time the changestream is setup
```
ohEmily, line 427: [minor] would it be worth renaming to `initialized`? or something else that documents that it only happens once.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -597,15 +602,21 @@ func (queue *mongoDBWebRTCCallQueue) processNextSubscriptionEvent(next mongoutil
 			if _, ok := queue.csTrackingHosts[callResp.Host]; !ok {
 				// no one connected to this operator is currently subscribed to insert
 				// events for this host; skip
+				// we do this because each server is listening to an event and each host only lives on one server
+				// it could be on another server
 				return
```
ohEmily, line 607: would it be worth adding a `debug` log here or would it be so insanely noisy that it's not useful?

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -614,10 +625,12 @@ func (queue *mongoDBWebRTCCallQueue) processNextSubscriptionEvent(next mongoutil
 				// though, we want to send the events as fast as possible as mentioned
 				// above and cannot block on the send to see if the receiver locked
 				// the document.
+				queue.logger.Infow("sending event", "event", event)
```
ohEmily, line 628: [q] wouldn't this log once per channel without including identifying information for the channel? Would it make sense to instead log the event in the log line above at line 619?

(Apologies if I'm misunderstanding here!)

rpc/wrtc_call_queue_mongodb.go
RoxyFarhad: i think it would be quite noisy and not useful

rpc/wrtc_call_queue_mongodb.go
RoxyFarhad: I actually just think we should remove this here, it will be noisy and it was mainly for me debugging locally :)

rpc/wrtc_call_queue_mongodb.go
RoxyFarhad: changed to isInitialized

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
ohEmily, line 1: (maybe committed by accident?)

rpc/wrtc_call_queue_mongodb.go
ohEmily: thanks!

ohEmily (approved):
LGTM! Wasn't sure exactly how you decided which comments to prefix with `note(roxy)` but it's a detail to add that; happy to defer to your decision tree. Thank you SO MUCH for marking this up and working to improve our signaling knowledge.

ohEmily (approved):

PR #230, opened 2024-01-05
APP-2747 Add hostname to metric

vijayvuyyuru:
Hey yall, this PR is to add the hostname to our metric for call answerers too busy. This will allow us to filter by host in the metric explorer so we can breakdown what robot are causing what errors more easily.


RoxyFarhad (approved):
thanks for this :)

PR #231, opened 2024-01-09
Move metadata initialization before usage

DTCurrie:
Fixes an issue where we were attempting to use a metadata object before it was initialized.

### DIFF
--- a/rpc/js/src/dial.ts
+++ b/rpc/js/src/dial.ts
@@ -120,6 +120,7 @@ async function makeAuthenticatedTransportFactory(
 ): Promise<grpc.TransportFactory> {
   let accessToken = '';
   const getExtraMetadata = async (): Promise<grpc.Metadata> => {
+    const md = new grpc.Metadata();
     // TODO(GOUT-10): handle expiration
     if (accessToken == '') {
       let thisAccessToken = '';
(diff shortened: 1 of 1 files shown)


rpc/js/src/dial.ts
```diff
@@ -204,7 +205,6 @@ async function makeAuthenticatedTransportFactory(
         accessToken = thisAccessToken;
       }
     }
-    const md = new grpc.Metadata();
     md.set('authorization', `Bearer ${accessToken}`);
```
mcous, line 208: Should this be moved up to? Having a little trouble following the code 🙈

DTCurrie: So I wasn't sure, because we change the value of `accessToken` before we set it here.

> Having a little trouble following the code 🙈

It is not easy to read, having a hard time untangling it myself.

DTCurrie: So re-reading the function and trying to clean it up a bit, it basically looks like there is a metadata object that we set an authenticated token on, that we then conditionally pass as the metadata to get an external auth token and use that as the metadata. So it is probably ok to just move that metadata object to the top since it gets passed around, and leave the set here since it should be in the right order of operations. It's really a mess to try to figure out.

njooma (approved):
Thank you!!

mcous (approved):

PR #234, opened 2024-01-11
RSDK-6151 parallel mDNS / WebRTC dials

maximpertsov:
Update dial logic to attempt mDNS and WebRTC connections in parallel.

# Testing

I ran a script that dials a robot 20 times using the current (serial) and newly implemented (parallel) dial logic. I ran this comparison under two network conditions:

* **hotspot -> viam**: Client and server on different networks, so mDNS will always fail.
* **viam -> viam**: Client and server on the same network, so mDNS may succeed.

Here are the results (in seconds):

| | Serial Hotspot->Viam | Parallel Hotspot->Viam | Serial Viam->Viam | Parallel Viam->Viam |
|------------|----------------------|------------------------|-------------------|---------------------|
| Avg        |                 5.28 |                   2.67 |              1.99 |                1.30 |
| Stdev      |                 1.30 |                   0.64 |              2.48 |                1.14 |

See [this spreadsheet](https://docs.google.com/spreadsheets/d/1m4uZfDQQoO7aIOB1DM1ZLcYMJ6QYirZAVtwTLaDw_rQ/edit#gid=1355373522) for more details.

# TODO

* [x] close unused connection if successful
* [x] run [this test](https://github.com/viamrobotics/goutils/pull/234#issuecomment-1894172191) from the same subnet to utilize mDNS
* [x] ~is the performance improvement worth the extra complexity?~ Yes, the performance profile (see Testing above) shows significant improvement, especially in cases where mDNS will always fail).
* [x] handle config mutations that might happen in parallel (either pass copies or lock as needed)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,14 +78,6 @@ jobs:
       run: sudo -u testbot bash -lc 'CI=true make test-web'
 
     - if: failure()
-      run: |
-        pwd
-        ls rpc
-        ls rpc/examples
-        ls rpc/examples/echo/
-        ls rpc/examples/echo/playwright-report
-
-    - if: failure()
       name: Upload js test report
       uses: actions/upload-artifact@v3
       with:
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -20,9 +20,6 @@ import (
 // via WebRTC if a signaling server is detected or provided. Otherwise it attempts to connect directly.
 // TODO(GOUT-7): figure out decent way to handle reconnect on connection termination.
 func Dial(ctx context.Context, address string, logger golog.Logger, opts ...DialOption) (ClientConn, error) {
-	start := time.Now()
-	defer func() { logger.Infof("dial time: %dms", time.Since(start).Milliseconds()) }()
-
 	var dOpts dialOptions
 	for _, opt := range opts {
 		opt.apply(&dOpts)
@@ -79,7 +76,7 @@ func dialInner(
 	}
 	if cached {
 		if dOpts.debug {
-			logger.Infow("connected (cached)", "address", address)
+			logger.Debugw("connected (cached)", "address", address)
 		}
 	}
 	return conn, nil
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -257,7 +257,7 @@ func dial(
 			logger.Debug("failed dial attempt, may still try direct", "error", err)
 		}
 	}
-	wg.Wait()
+
 	if conn != nil {
 		return conn, cached, nil
 	}
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -218,7 +218,7 @@ func dial(
 				// TODO(RSDK-6493): Investigate if we must `skipDirect` here.
 				dialCh <- dialResult{err: err, skipDirect: true}
 			case ctxParallel.Err() != nil:
-				dialCh <- dialResult{err: err, skipDirect: true}
+				dialCh <- dialResult{err: ctxParallel.Err(), skipDirect: true}
 			default:
 				dialCh <- dialResult{err: err}
 			}
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -93,6 +93,9 @@ func dial(
 	dOpts *dialOptions,
 	tryLocal bool,
 ) (ClientConn, bool, error) {
+	start := time.Now()
+	defer func() { logger.Infof("dial time: %dms", time.Since(start).Milliseconds()) }()
+
 	var isJustDomain bool
 	if strings.HasPrefix(address, "unix://") {
 		dOpts.mdnsOptions.Disable = true
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -135,10 +135,12 @@ func dial(
 	)
 	defer cancelParallel()
 	if !dOpts.mdnsOptions.Disable && tryLocal && isJustDomain {
+		dOptsCopy := *dOpts
+
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			conn, cached, err := dialMulticastDNS(ctxParallel, address, logger, dOpts)
+			conn, cached, err := dialMulticastDNS(ctxParallel, address, logger, &dOptsCopy)
 			switch {
 			case err != nil:
 				dialCh <- dialResult{err: err}
@@ -167,7 +169,6 @@ func dial(
 					dialCh <- dialResult{err: err, skipDirect: true}
 					return
 				}
-				// TODO: add lock or copy dOpts to safely mutate concurrently
 				fixupWebRTCOptions(dOpts, target, port)
 
 				// When connecting to an external signaler for WebRTC, we assume we can use the external auth's material.
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -142,7 +142,8 @@ func dial(
 			case err != nil:
 				dialCh <- dialResult{err: err}
 			case conn != nil:
-				// TODO(docs): how can we get a `nil` connection when there is no error?
+				// TODO(RSDK-6490): investigate why we can get a `nil` connection that is
+				// not accompanied with an error.
 				dialCh <- dialResult{conn: conn, cached: cached, usedMDNS: true}
 			default:
 				dialCh <- dialResult{err: errors.New("no connection")}
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -125,6 +125,10 @@ func dial(
 		isJustDomain = net.ParseIP(address) == nil
 	}
 
+	// RSDK-6151: We make concurrent dial attempts via mDNS and WebRTC, taking the first
+	// connection that succeeds. We then cancel the slower connection and wait for its
+	// coroutine to complete. If the slower connection succeeds before it can be
+	// cancelled then we explicitly close it to prevent a memory leak.
 	var (
 		wg                          sync.WaitGroup
 		dialCh                      = make(chan dialResult)
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -116,9 +116,9 @@ func dial(
 
 	connCh := make(chan dialSuccess, 2)
 	errCh := make(chan error, 2)
-	var dialAttempts int
+	var concurrentDials int
 	if !dOpts.mdnsOptions.Disable && tryLocal && isJustDomain {
-		dialAttempts++
+		concurrentDials++
 		go func() {
 			conn, cached, err := dialMulticastDNS(ctx, address, logger, dOpts)
 			if err != nil {
@@ -137,7 +137,7 @@ func dial(
 	}
 
 	if !dOpts.webrtcOpts.Disable {
-		dialAttempts++
+		concurrentDials++
 		go func() {
 			signalingAddress := dOpts.webrtcOpts.SignalingServerAddress
 			if signalingAddress == "" || dOpts.webrtcOpts.AllowAutoDetectAuthOptions {
@@ -209,12 +209,15 @@ func dial(
 		dialFails  int
 		dialErrors []error
 	)
-	for dialFails < dialAttempts {
+	for dialFails < concurrentDials {
 		select {
 		case success := <-connCh:
 			return success.conn, success.cached, nil
 		case err := <-errCh:
 			dialFails++
+			// a nil error means that a concurrent dial attempt failed, but it should not
+			// stop us from trying to establish a direct gRPC connection after all
+			// concurrent dial attempts fail.
 			if err != nil {
 				dialErrors = append(dialErrors, err)
 			}
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -98,7 +98,7 @@ type dialResult struct {
 	// connection errors
 	err error
 	// whether we should skip dialing gRPC directly as a fallback
-	exitEarly bool
+	skipDirect bool
 }
 
 func dial(
@@ -166,7 +166,7 @@ func dial(
 				target, port, err := getWebRTCTargetFromAddressWithDefaults(signalingAddress)
 				if err != nil {
 					// TODO(docs): why don't we try dialing directly after this error?
-					dialCh <- dialResult{err: err, exitEarly: true}
+					dialCh <- dialResult{err: err, skipDirect: true}
 					return
 				}
 				// TODO: add lock or copy dOpts to safely mutate concurrently
@@ -215,9 +215,9 @@ func dial(
 				dialCh <- dialResult{conn: conn, cached: cached}
 			case !errors.Is(err, ErrNoWebRTCSignaler):
 				// TODO(docs): why don't we try dialing directly after this error?
-				dialCh <- dialResult{err: err, exitEarly: true}
+				dialCh <- dialResult{err: err, skipDirect: true}
 			case ctxParallel.Err() != nil:
-				dialCh <- dialResult{err: err, exitEarly: true}
+				dialCh <- dialResult{err: err, skipDirect: true}
 			default:
 				dialCh <- dialResult{err: err}
 			}
@@ -248,7 +248,7 @@ func dial(
 			conn, cached = result.conn, result.cached
 			mu.Unlock()
 			cancelParallel()
-		case result.err != nil && result.exitEarly:
+		case result.err != nil && result.skipDirect:
 			logger.Debug("failed dial attempt, will not try direct", "error", err)
 			mu.Lock()
 			err = result.err
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -116,26 +116,26 @@ func dial(
 		isJustDomain = net.ParseIP(address) == nil
 	}
 
-	connCh := make(chan dialSuccess, 2)
-	errCh := make(chan error, 2)
-	var cancels []context.CancelFunc
+	var (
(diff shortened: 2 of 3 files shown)


rpc/dial.go
```diff
@@ -107,77 +118,118 @@ func dial(
...
+			if err.exit {
+				exit = true
 			}
-			return conn, cached, nil
-		}
-		if !errors.Is(err, ErrNoWebRTCSignaler) {
-			return nil, false, err
-		}
-		if ctx.Err() != nil {
-			return nil, false, ctx.Err()
```
maximpertsov, line 178: It's not clear to me why we exit early on these particular errors without making an attempt to dial directly through gRPC.

rpc/dial.go
```diff
@@ -99,85 +105,121 @@ func dial(
...
 	if !dOpts.webrtcOpts.Disable {
-		signalingAddress := dOpts.webrtcOpts.SignalingServerAddress
-		if signalingAddress == "" || dOpts.webrtcOpts.AllowAutoDetectAuthOptions {
-			if signalingAddress == "" {
-				// try WebRTC at same address
-				signalingAddress = address
-			}
-			target, port, err := getWebRTCTargetFromAddressWithDefaults(signalingAddress)
-			if err != nil {
-				return nil, false, err
```
maximpertsov, line 130: ditto https://github.com/viamrobotics/goutils/pull/234/files#r1449060104

cheukt:
what sort of testing did you do for this change? any notable speedups (I would assume ~1 s for non-LAN connections)?

maximpertsov:
> what sort of testing did you do for this change? any notable speedups (I would assume ~1 s for non-LAN connections)?

I only tested locally so far - let me try against a real robot.

ale7714:
Will this change only impact Viam server?

rpc/dial.go
cheukt: the context one makes sense, and tho the one above I'm not sure about - maybe we know for sure that if there is a webrtc signaler and we can't connect there's no chance of the connection succeeding?

rpc/dial.go
cheukt: ya not sure about this - do the errors from getWebRTCTargetFromAddressWithDefaults seem like they would break direct grpc as well?

.github/workflows/test.yml
```diff
@@ -77,14 +77,6 @@ jobs:
     - name: Test js library
       run: sudo -u testbot bash -lc 'CI=true make test-web'
 
-    - if: failure()
-      run: |
-        pwd
-        ls rpc
-        ls rpc/examples
-        ls rpc/examples/echo/
-        ls rpc/examples/echo/playwright-report
```
benjirewis, line 86: [q] Just didn't need this extra debug information?

maximpertsov: yeah, this was testing from another PR - should have been removed there

maximpertsov:
> > what sort of testing did you do for this change? any notable speedups (I would assume ~1 s for non-LAN connections)?
> 
> I only tested locally so far - let me try against a real robot.

I tried dialing a robot in the office 20 times from home the old serial logic and the new parallel logic, and the new parallel logic does seem to be faster on average with less variance:

**Dial time in seconds**
|       | Serial | Parallel |
|-------|--------|----------|
| Avg   |   5.66 |     3.18 |
| Stdev |   1.48 |     1.07 |

**Reference**
https://docs.google.com/spreadsheets/d/1m4uZfDQQoO7aIOB1DM1ZLcYMJ6QYirZAVtwTLaDw_rQ/edit#gid=0

benjirewis:
> I tried dialing a robot in the office 20 times from home the old serial logic and the new parallel logic, and the new parallel logic does seem to be faster on average with less variance:

Slick testing 🧑‍🔧 ! Ty for the spreadsheet, too. Any chance you tried/could try connecting to a robot on the same subnet? Curious what happens to connections that should leverage mDNS with this change.

rpc/dial.go
```diff
@@ -94,90 +101,137 @@ func dial(
...
+	var (
+		dialFails  int
+		dialErrors []error
+	)
+	for dialFails < len(cancels) {
+		select {
+		case success := <-connCh:
+			for i, cancel := range cancels {
+				if i != success.key {
+					cancel()
```
maximpertsov, line 218: looking at this again, I don't think this handles the case of the slower connection succeeding - my understanding is that `cancel` does nothing in that case, and the connection gets leaked, right?

maximpertsov:
> > I tried dialing a robot in the office 20 times from home the old serial logic and the new parallel logic, and the new parallel logic does seem to be faster on average with less variance:
> 
> Slick testing 🧑‍🔧 ! Ty for the spreadsheet, too. Any chance you tried/could try connecting to a robot on the same subnet? Curious what happens to connections that should leverage mDNS with this change.

I was testing this in the office earlier and it seemed like there wasn't a big difference in dial times between parallel and serial, but it's worth checking again.

maximpertsov:
> > > I tried dialing a robot in the office 20 times from home the old serial logic and the new parallel logic, and the new parallel logic does seem to be faster on average with less variance:
> > 
> > 
> > Slick testing 🧑‍🔧 ! Ty for the spreadsheet, too. Any chance you tried/could try connecting to a robot on the same subnet? Curious what happens to connections that should leverage mDNS with this change.
> 
> I was testing this in the office earlier and it seemed like there wasn't a big difference in dial times between parallel and serial, but it's worth checking again.

Same test, in the office, in seconds
|       | Serial | Parallel |
|-------|--------|----------|
| Avg   |   0.92 |     1.55 |
| Stdev |   0.70 |     1.86 |

second tab: https://docs.google.com/spreadsheets/d/1m4uZfDQQoO7aIOB1DM1ZLcYMJ6QYirZAVtwTLaDw_rQ/edit#gid=1318453722

rpc/dial.go
```diff
@@ -94,90 +101,137 @@ func dial(
...
+				fixupWebRTCOptions(dOpts, target, port)
+
+				// When connecting to an external signaler for WebRTC, we assume we can use the external auth's material.
+				// This path is also called by an mdns direct connection and ignores that case.
+				// This will skip all Authenticate/AuthenticateTo calls for the signaler.
+				if !dOpts.usingMDNS && dOpts.authMaterial == "" && dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial != "" {
+					logger.Debug("using signaling's external auth as auth material")
+					dOpts.authMaterial = dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial
+					dOpts.creds = Credentials{}
+				}
```
maximpertsov, line 163: TODO: `fixupWebRTCOptions` and the lines below mutate the dial options struct - need to either lock or pass copies since this operation could happen concurrently now.

rpc/dial.go
```diff
@@ -19,6 +20,9 @@ import (
 // via WebRTC if a signaling server is detected or provided. Otherwise it attempts to connect directly.
 // TODO(GOUT-7): figure out decent way to handle reconnect on connection termination.
 func Dial(ctx context.Context, address string, logger golog.Logger, opts ...DialOption) (ClientConn, error) {
+	start := time.Now()
+	defer func() { logger.Infof("dial time: %dms", time.Since(start).Milliseconds()) }()
+
```
maximpertsov, line 25:
```suggestion
```

added for testing purposes

rpc/dial.go
```diff
@@ -75,7 +79,7 @@ func dialInner(
 	}
 	if cached {
 		if dOpts.debug {
-			logger.Debugw("connected (cached)", "address", address)
+			logger.Infow("connected (cached)", "address", address)
```
maximpertsov, line 82:
```suggestion
			logger.Debugw("connected (cached)", "address", address)
```

changed for testing purposes

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			conn, cached, err := dialMulticastDNS(ctxParallel, address, logger, &dOptsCopy)
+			switch {
+			case err != nil:
+				dialCh <- dialResult{err: err}
+			case conn != nil:
+				// TODO(docs): how can we get a `nil` connection when there is no error?
+				dialCh <- dialResult{conn: conn, cached: cached, usedMDNS: true}
```
maximpertsov, line 146: from what I can tell, there are some code paths where mDNS does not find a host but there is no associated error. we really should throw an error in this case instead... but I'd prefer to do this on a follow-up PR

rpc/dial.go
maximpertsov: [done indirectly by copy dial options in the mDNS dial routine earlier](https://github.com/viamrobotics/goutils/pull/234/commits/9c86723c004e7af6626ed60517756ba530658d4d).

I really don't like how various stages of this dial pipeline modify options in place, but it might be too much to undo in this PR. I'm tempted to at least update all function signature to accept `dialOptions` values rather than pointers.

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
-					dOpts,
-					logger,
-				)
-			})
-		if err == nil {
-			if !cached {
-				logger.Debug("connected via WebRTC")
-			} else if dOpts.debug {
-				logger.Debug("connected via WebRTC (cached)")
+	go func() {
```
benjirewis, line 223: Took me a bit to understand the function of this goroutine; would you mind adding a comment above it elaborating a bit on your method? I see now that it's closely tied to the `for` loop below. My understanding is that: the `for` loop calls `cancelParallel` to stop the slower method of dialing, the goroutine `Wait`s for the slower method to "finish up" after presumably hitting a `context.Canceled`, and the goroutine closes the `dialCh` to allow the main function to continue running.

benjirewis (commented):
Cool 🧑‍🔧 ! Some questions on the concurrent stuff.

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
+						ctxParallel,
+						dOpts.webrtcOpts.SignalingServerAddress,
+						originalAddress,
+						dOpts,
+						logger,
+					)
+				})
+
+			switch {
+			case err == nil:
```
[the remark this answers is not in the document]
maximpertsov: I'm assuming no, but only because we were not explicitly checking for this case like we were in the mDNS dial logic

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
+		conn     ClientConn
+		cached   bool
+		err      error
+		usedMDNS bool
+		mu       sync.Mutex
+	)
+	for result := range dialCh {
+		switch {
+		case result.err == nil && result.conn != nil:
+			mu.Lock()
```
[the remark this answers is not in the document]
maximpertsov: I don't think we need this anymore - it might have been a relic from an earlier iteration - will confirm.

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
+			defer wg.Done()
+			signalingAddress := dOpts.webrtcOpts.SignalingServerAddress
+			if signalingAddress == "" || dOpts.webrtcOpts.AllowAutoDetectAuthOptions {
+				if signalingAddress == "" {
+					// try WebRTC at same address
+					signalingAddress = address
+				}
+				target, port, err := getWebRTCTargetFromAddressWithDefaults(signalingAddress)
+				if err != nil {
+					// TODO(docs): why don't we try dialing directly after this error?
```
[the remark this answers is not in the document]
maximpertsov: just noting that I don't understand why an error here warrants exiting early and not trying to dial directly: https://github.com/viamrobotics/goutils/pull/234/files#r1450884029

rpc/dial.go
maximpertsov: that's correct - I'll add a comment to explain this flow

rpc/dial.go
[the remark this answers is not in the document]
maximpertsov: L213 corresponds to this thread: https://github.com/viamrobotics/goutils/pull/234#discussion_r1450882536

rpc/dial.go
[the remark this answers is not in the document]
maximpertsov: yeah the mutex is totally unnecessary and I can't remember why I added it - thank you for catching and sorry for being sloppy here

rpc/dial.go
maximpertsov: how's this? https://github.com/viamrobotics/goutils/pull/234/commits/4036d5b7a6a1960f0aec20fc09147161e262f152 - let me know if you'd like me to clarify anything further.

thank you for pushing me to document this better!

rpc/dial.go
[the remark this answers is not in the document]
maximpertsov: I'll make a ticket for these as well

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
```
benjirewis, line 170:
```suggestion
					// TODO(RSDK-6493): Investigate if we must `skipDirect` here.
```

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
+			switch {
+			case err == nil:
+				if !cached {
+					logger.Debug("connected via WebRTC")
+				} else if dOpts.debug {
+					logger.Debug("connected via WebRTC (cached)")
+				}
+				dialCh <- dialResult{conn: conn, cached: cached}
+			case !errors.Is(err, ErrNoWebRTCSignaler):
+				// TODO(docs): why don't we try dialing directly after this error?
```
benjirewis, line 218:
```suggestion
				// TODO(RSDK-6493): Investigate if we must `skipDirect` here.
```

rpc/dial.go
benjirewis: That comment looks great thank you! Was just getting a little lost in the flow.

benjirewis (approved):
LGTM mod the `TODO`s + tests passing!

rpc/dial.go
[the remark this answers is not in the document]
benjirewis: Ah no worries; I definitely buy that it came from a previous version + was necessary then.

rpc/dial.go
maximpertsov: I made [a follow-up ticket](https://viam.atlassian.net/browse/RSDK-6493) to investigate or possibly refactor this

rpc/dial.go
maximpertsov: in any case, I made [a follow-up ticket](https://viam.atlassian.net/browse/RSDK-6493) to investigate or possibly refactor this

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
-		} else {
-			isJustDomain = net.ParseIP(address) == nil
-		}
+	case strings.ContainsRune(address, ':'):
+		isJustDomain = false
+	default:
+		isJustDomain = net.ParseIP(address) == nil
 	}
 
+	// RSDK-6151: We make concurrent dial attempts via mDNS and WebRTC, taking the first
```
cheukt, line 128: nit: don't need ticket number for if not TODO

rpc/dial.go
```diff
@@ -93,90 +108,161 @@ func dial(
...
-					dOpts,
-					logger,
-				)
-			})
-		if err == nil {
-			if !cached {
-				logger.Debug("connected via WebRTC")
-			} else if dOpts.debug {
-				logger.Debug("connected via WebRTC (cached)")
+	go func() {
```
cheukt, line 228:
```suggestion
        // Make sure the slower connection attempt is fully cancelled, or if the attempt succeeded,
        // close the slower connection.
	go func() {
```
can we add a comment here to reference its purpose in the overall function?

cheukt (approved):
LGTM with some minor doc nits, what debug logs does this function spew out now? do we only log at the end?

rpc/dial.go
maximpertsov: we've been linking some of our code comments back to JIRA tickets to give more context on things work the way they do, for example: https://github.com/viamrobotics/rdk/pull/3467/files#diff-4e03c7046f4cde2fb3cfabe60f410c7cc0380ec31028edec049f6a37cab65cd8R119

maximpertsov:
> LGTM with some minor doc nits, what debug logs does this function spew out now? do we only log at the end?

This was challenging due to the recursive nature of `dial` - even if we log at the end of the parallel attempts we still get multiple lines for each level of recursion.

I ended up removing all debug logs related to mDNS and instead report whether a WebRTC connection used mDNS since we store that information in dial options. Now we will only get two connection success lines if both the fast and slow connections succeed: https://github.com/viamrobotics/goutils/pull/234/commits/3726f0df0ff73c58994410cf200cccc0c022613a

Debug log if mDNS was used:
```
2024-01-30T18:25:27.851-0500    DEBUG   client  rpc/dial.go:209 connected via WebRTC    {"address": "10.1.X.XXX:XXXXX", "cached": false, "using mDNS": true}
```

Debug log if mDNS was not used:
```
2024-01-30T18:26:14.228-0500    DEBUG   client  rpc/dial.go:209 connected via WebRTC    {"address": "some-host.xxxxx.yyyy.cloud", "cached": false, "using mDNS": false}
```

rpc/dial.go
maximpertsov: On second thought, it's not necessary here

PR #240, opened 2024-01-31
RSDK-6041 - Add dial timeout to ts 

cheukt:
adding dial timeout lets us give up on failed connection attempts faster, shouldn't matter too much for fast connections, but 95% time is a lot faster (on app, which auto redials)

tested by linking the libraries together, so checked that the dialTimeout option actually works

data collected doesn't show the value of the change very well, but from previous anecdotal data, any connection that fails on the first try will take 20s, but with the new change it would be closer to 8-9s (since first failures usually take 15s or so)

| Attempt | With Change (# fails) | Without Change (# fails) |
|---------|-----------------------|--------------------------|
| 1       | 9.5 (1)               | 30.07                    |
| 2       | 4.61                  | 3                        |
| 3       | 1.591                 | 6.7                      |
| 4       | 1.282                 | 1.5                      |
| 5       | 1.216                 | 1.2                      |
| 6       | 2.234                 | 1.1                      |
| 7       | 1.465                 | 1.8                      |
| 8       | 10.7 (1)              | 5.9                      |
| 9       | 2.031                 | 1.59                     |
| 10      | 1.639                 | 1.4                      |

seems like we can actually make the default lower, but can see for now

### DIFF
--- a/rpc/js/src/dial.ts
+++ b/rpc/js/src/dial.ts
@@ -44,7 +44,7 @@ export interface DialOptions {
   // externalAuthAddress, externalAuthToEntity, webrtcOptions.signalingAccessToken
   accessToken?: string | undefined;
 
-  // set timeout in milliseconds for dialing. default is 5000.
+  // set timeout in milliseconds for dialing.
   dialTimeout?: number | undefined;
 }
 
@@ -589,12 +589,14 @@ export async function dialWebRTC(
 
     const cc = new ClientChannel(pc, dc);
 
-    // set timeout for dial attempt
-    setTimeout(() => {
-      if (!successful) {
-        cc.close();
-      }
-    }, opts.dialTimeout ?? 5000);
+    // set timeout for dial attempt if a timeout is specified
+    if (opts.dialTimeout) {
+      setTimeout(() => {
+        if (!successful) {
+          cc.close();
+        }
+      }, opts.dialTimeout);
+    }
 
     cc.ready
       .then(() => clientEndResolve())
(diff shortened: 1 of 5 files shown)


rpc/js/src/dial.ts
```diff
@@ -581,6 +588,17 @@ export async function dialWebRTC(
...
     const cc = new ClientChannel(pc, dc);
+
+    // set timeout for dial attempt
+    setTimeout(
+      () => {
+        if (!successful) {
+          cc.close();
+        }
+      },
+      opts?.dialTimeout ? opts?.dialTimeout : 5000
```
maximpertsov, line 599:
```suggestion
      opts.dialTimeout ?? 5000
```

maximpertsov:
I don't fully understand the chart in the description - how do more or less failures correlate with time savings on the connection? Also, what do numbers in parens mean? i.e. the first row:

| Attempt | With Change (# fails) | Without Change (# fails) |
|:-------:|:---------------------:|:------------------------:|
| 1       | 9.5 (1) <- ?             | 30.07                    |

cheukt:
the number is seconds to connect successfully, the hypothesis is that more failures is good since we want to retry once we know something isn't likely to work. But when I tested somehow it didn't fail as often for the control

maximpertsov:
> I don't fully understand the chart in the description - how do more or less failures correlate with time savings on the connection? Also, what do numbers in parens mean? i.e. the first row:
> Attempt 	With Change (# fails) 	Without Change (# fails)
> 1 	9.5 (1) <- ? 	30.07

Discussed IRL - the updated description now has the approximate time savings.

maximpertsov (approved):

maximpertsov (approved):

PR #243, opened 2024-02-13
RSDK-5562 Ignore sendDone errors in signaling answerer

benjirewis:
[RSDK-5562](https://viam.atlassian.net/browse/RSDK-5562) (one manifestation of the issue; I can make a separate ticket if desired)

This patch "fixes" the repro of immediate context cancelation from `client.New` I have and logs:
```
logger.go:130: 2024-02-13T09:52:42.292-0500 WARN    internal_signaler       rpc/wrtc_signaling_answerer.go:495 error ending signaling exchange from answer client       {"error": "EOF"}
```

cc @cheukt 

[RSDK-5562]: https://viam.atlassian.net/browse/RSDK-5562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -478,7 +478,12 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 		return err
 	}
 	if err := sendDone(); err != nil {
-		return err
+		// Errors from sendDone (such as EOF) are sometimes caused by the signaling
+		// server "ending" the exchange process earlier than the answerer due to
+		// the caller being able to establish a connection without all the
+		// answerer's ICE candidates (trickle ICE). Only Warn the error here to
+		// avoid accidentally Closing a healthy, established peer connection.
+		ans.logger.Warnw("error ending signaling exchange from answer client", "error", err)
 	}
 	successful = true
 	return nil


dgottlieb (approved):
This lgtm and the comment looks great.

We definitely need to document more of what we found. Probably as a primer in a wiki page.

rpc/wrtc_signaling_answerer.go
```diff
@@ -478,7 +478,12 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 		return err
 	}
 	if err := sendDone(); err != nil {
-		return err
+		// Errors from sendDone (such as EOF) are sometimes caused by the signaling
+		// server "ending" the exchange process earlier than the answerer due to
+		// the caller being able to establish a connection without all the
+		// answerer's ICE candidates (trickle ICE). Only Warn the error here to
+		// avoid accidentally Closing a healthy, established peer connection.
```
maximpertsov, line 485: [minor] unless you purposely capitalized to refer to functions

```suggestion
		// answerer's ICE candidates (trickle ICE). Only warn about the error here to
		// avoid accidentally closing a healthy, established peer connection.
```

rpc/wrtc_signaling_answerer.go
```diff
@@ -478,7 +478,12 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 		return err
 	}
 	if err := sendDone(); err != nil {
```
maximpertsov, line 480: should we be more conservative and only warn if there is an EOF? or are we pretty comfortable just warning on any error at this point?

maximpertsov (commented):
this is awesome. just a minor q and comment suggestion

rpc/wrtc_signaling_answerer.go
benjirewis: Ah I did mean to refer to `logger.Warn` and `peerConnection.Close`, so might leave the caps.

rpc/wrtc_signaling_answerer.go
dgottlieb: I think the more "conservative" behavior is to ignore all errors. As per the conversion [following this](https://viaminc.slack.com/archives/C06542APFRQ/p1707832372930119?thread_ts=1707763492.626749&cid=C06542APFRQ). The error handling here is an optimization that we demonstrate goes wrong. It's more conservative to not have that optimization.

I don't see any reason why the more conservative change* of only suppressing EOF has an advantage.

maximpertsov: cool, thanks for the discussion link! makes sense, I'm on board.

maximpertsov (approved):

PR #244, opened 2024-02-13
RSDK-6151: Add cause to parallel dial cancellation

maximpertsov:
Add cause to parallel dial cancellation.

* [x] Depends on: https://github.com/viamrobotics/goutils/pull/245

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module go.viam.com/utils
 
-go 1.18
+go 1.20
 
 require (
 	cloud.google.com/go/compute/metadata v0.2.3
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -130,9 +130,9 @@ func dial(
 	var (
 		wg                          sync.WaitGroup
 		dialCh                      = make(chan dialResult)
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module go.viam.com/utils
 
-go 1.18
+go 1.20
```
maximpertsov, line 3: @abe-winter I'm introducing a go 1.20 feature in this change - which CI image should I be building with to support this?

abe-winter: switch your base image from `canon` to `rdk-devenv`

maximpertsov: discussed offline, since this library doesn't need most of the dependencies provided by `rdk-devenv`, I ended up using a generic `ubuntu-latest` image with `setup-go/-js/-protoc` instead: https://github.com/viamrobotics/goutils/pull/245

benjirewis (approved):
LGTM, and I definitely like this pattern especially when there are multiple cancellation locations with different implications.

cheukt (approved):
LGTM

PR #248, opened 2024-02-23
APP-3914: Add function to pass options to mongoclient

jr22:
We'll want to initialize some tests with a mongo client that has specific options set. Creating a helper that allows that.


testutils/mongodb.go
```diff
@@ -143,3 +147,14 @@ func BackingMongoDBClient(tb testing.TB) *mongo.Client {
 	}
 	return client
 }
+
+// BackingMongoDBClientWithOptions returns a backing MongoDB client to use with the provided options.
+func BackingMongoDBClientWithOptions(tb testing.TB, baseOptions *options.ClientOptions) *mongo.Client {
+	tb.Helper()
+	client, err := backingMongoDBClientWithOptions(baseOptions)
+	if err != nil {
+		skipWithError(tb, err)
```
RoxyFarhad, line 156: for my own understanding: why do we do this?

testutils/mongodb.go
```diff
@@ -110,7 +114,7 @@ func backingMongoDBClient() (*mongo.Client, error) {
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
-	client, err := mongo.NewClient(options.Client().ApplyURI(mongoURI))
+	client, err := mongo.NewClient(baseOptions.ApplyURI(mongoURI))
```
RoxyFarhad, line 117: check that baseOptions is not nil? think this will panic otherwise?

testutils/mongodb.go
jr22: i think this is so that if this fails for one specific test, itll "skip" the test with the error mesasge included and continue running other tests

testutils/mongodb.go
jr22: updated! this is all just testutils stuff so wasn't too worried but probably better to add it

RoxyFarhad: oh yeah, I guess I was thinking that if some test instantiated this with a nil client it would panic. but yeah maybe we aren't even exposing that code path.

PR #249, opened 2024-02-29
Enable React Native

njooma:
This PR enables the RPC library to be used within React Native settings. The major changes required are:

* Allowing a configuration for a non-browser `grpc.TransportFactory`
  * The user can specify a new `TransportFactory` by setting the global variable `global.VIAM.GRPC_TRANSPORT_FACTORY`
  * This library will default to the cross-browser transport
* Polyfill `atob` and `btoa` because those aren't available outside of browser environments
* Change all the `on{EVENT}` functions to `addEventListener({EVENT})` to work with both browser and React Native

### DIFF
--- a/rpc/js/src/dial.ts
+++ b/rpc/js/src/dial.ts
@@ -1,5 +1,7 @@
 import { grpc } from '@improbable-eng/grpc-web';
+import { ReactNativeTransport } from '@improbable-eng/grpc-web-react-native-transport';
 import type { ProtobufMessage } from '@improbable-eng/grpc-web/dist/typings/message';
+import { RTCPeerConnection, RTCSessionDescription } from 'react-native-webrtc';
 import { ClientChannel } from './ClientChannel';
 import { ConnectionClosedError } from './errors';
 import { Code } from './gen/google/rpc/code_pb';
@@ -21,12 +23,14 @@ import {
   CallUpdateRequest,
   CallUpdateResponse,
   ICECandidate,
-  WebRTCConfig,
   OptionalWebRTCConfigRequest,
   OptionalWebRTCConfigResponse,
+  WebRTCConfig,
 } from './gen/proto/rpc/webrtc/v1/signaling_pb';
 import { SignalingService } from './gen/proto/rpc/webrtc/v1/signaling_pb_service';
 import { addSdpFields, newPeerConnectionForClient } from './peer';
+
+import { atob, btoa } from './polyfills';
 
 export interface DialOptions {
   authEntity?: string | undefined;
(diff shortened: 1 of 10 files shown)


rpc/js/src/dial.ts
```diff
@@ -458,44 +468,47 @@ export async function dialWebRTC(
     let exchangeDone = false;
     if (!webrtcOpts?.disableTrickleICE) {
       // set up offer
-      const offerDesc = await pc.createOffer();
+      const offerDesc = await pc.createOffer({});
```
maximpertsov, line 471: is this change significant to the logic or just a linting recommendation?

rpc/js/src/dial.ts
```diff
@@ -21,13 +21,16 @@ import {
...
   ICECandidate,
-  WebRTCConfig,
   OptionalWebRTCConfigRequest,
   OptionalWebRTCConfigResponse,
+  WebRTCConfig,
 } from './gen/proto/rpc/webrtc/v1/signaling_pb';
 import { SignalingService } from './gen/proto/rpc/webrtc/v1/signaling_pb_service';
 import { addSdpFields, newPeerConnectionForClient } from './peer';
 
+import { atob, btoa } from './polyfills';
```
maximpertsov, line 31: do we need to import these here? it doesn't look like we use until later

maximpertsov (commented):
nice! some minor questions for my understanding

rpc/js/src/dial.ts
njooma: It's significant to the logic -- the react-native version need the empty dict

rpc/js/src/dial.ts
njooma: We can't use the builtin versions of `atob` and `btoa` anymore because those are browser only and won't work in react native (and node in the future)

maximpertsov: understood, but what I meant with my comment is that these functions are not called in this file, so is it necessary to import them here? are they used under the hood somewhere? it looks like we only call them in `rpc/js/src/peer.ts`

njooma: We call them in this file too: https://github.com/viamrobotics/goutils/pull/249/files#diff-19b1e810c0b7e0d9e2b5bb4c4393faa56cd8c1d4b144c2605b9f1113a61810f9R534

purplenicole730 (approved):
LGTM, although `polyfills.ts` went over my head, so I can't speak on that.

rpc/js/src/dial.ts
maximpertsov: whoops, didn't expand enough... should just looked at the full file

maximpertsov (approved):

PR #250, opened 2024-03-04
Add github token

njooma:
The `proto-checks` workflow failed because of a rate limit exception (see [this run](https://github.com/viamrobotics/goutils/actions/runs/8143607497/job/22255828000))

This PR should hopefully fix it. Also migrates away from a self-hosted runner


Otterverse (approved):
LGTM!

PR #251, opened 2024-03-06
RSDK-1687 Better synchronize `Stop` and `processHeaders`

benjirewis:
RSDK-1687

Changes `wrtc_server.go` and `wrtc_server_stream.go` logic:
- Renames `mu` to `peerConnsMu`.
- Renames `activeBackgroundWorkers` to `processHeadersWorkers`.
- Adds a `processHeadersMu` to guard it
- `RLock`s the `processHeadersMu` when `Add`ing to `processHeadersWorkers`
- `Lock`s the `processHeadersMu` when `Wait`ing on `processHeadersWorkers`

The linked ticket describes a data race where `activeBackgroundWorkers.Add(1)` was racing with `activeBackgroundWorkers.Wait()`. This was happening because the `Add` was in a separate goroutine from the `Wait` due to the asynchronous nature of WebRTC "callbacks", and there was no synchronization of `Add`s and `Wait`s.

cc @viamrobotics/netcode .

### DIFF
--- a/rpc/wrtc_server_stream.go
+++ b/rpc/wrtc_server_stream.go
@@ -277,17 +277,19 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
 		}
 	}
 
-	s.ch.server.peerConnsMu.RLock()
+	s.ch.server.processHeadersMu.RLock()
 	s.ch.server.processHeadersWorkers.Add(1)
-	s.ch.server.peerConnsMu.RUnlock()
+	s.ch.server.processHeadersMu.RUnlock()
+
+	// Check if context has errored: underlying server may have been `Stop`ped,
+	// in which case we mark this processHeaders worker as `Done` and return.
+	if err := s.ch.server.ctx.Err(); err != nil {
+		s.ch.server.processHeadersWorkers.Done()
+		return
+	}
 
 	// take a ticket
 	select {
-	case <-s.ch.server.ctx.Done():
-		if err := s.closeWithSendError(status.FromContextError(s.ch.server.ctx.Err()).Err()); err != nil {
-			s.logger.Errorw("error closing", "error", err)
-		}
-		return
 	case s.ch.server.callTickets <- struct{}{}:
 	default:
 		if err := s.closeWithSendError(status.Error(codes.ResourceExhausted, "too many in-flight requests")); err != nil {
--- a/rpc/wrtc_server_stream.go
+++ b/rpc/wrtc_server_stream.go
@@ -293,12 +293,22 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
 	}
 
 	s.headersReceived = true
-	s.ch.server.activeBackgroundWorkers.Add(1)
+	s.ch.server.processHeadersLock.Lock()
+	// Check once more if underlying webrtcServer was `Stop`ped before lock was
+	// acquired (context was canceled).
+	if err := s.ch.server.ctx.Err(); err != nil {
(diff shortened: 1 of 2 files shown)


rpc/wrtc_server_stream.go
```diff
@@ -293,12 +293,22 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
 	}
 
 	s.headersReceived = true
-	s.ch.server.activeBackgroundWorkers.Add(1)
+	s.ch.server.processHeadersLock.Lock()
+	// Check once more if underlying webrtcServer was `Stop`ped before lock was
+	// acquired (context was canceled).
```
benjirewis, line 298: This is new logic: if we `Lock()` and the underlying server was closed (context was canceled), we can `closeWithSendError` and `return` as is currently done on L283. I'm not exactly sure what the behavior is here when peer connections are closed, but it seems like we were already fine calling `closeWithSendError` after the server's context had expired...

dgottlieb:
I'm pretty sure this change makes it such that we only run one grpc request at a time, which would be a dealbreaker.

IIUC, what we have is a real logic bug (not simply a data race). But I do think we need the wait group. But using it correctly requires more nuance.

Dave Cheney's comment was simplified down for that example. It doesn't apply here.

rpc/wrtc_server_stream.go
```diff
@@ -293,12 +297,11 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
 	}
 
 	s.headersReceived = true
-	s.ch.server.activeBackgroundWorkers.Add(1)
 	utils.PanicCapturingGo(func() {
 		defer func() {
+			s.ch.server.processHeadersWorkers.Done()
 			<-s.ch.server.callTickets // return a ticket
 		}()
-		defer s.ch.server.activeBackgroundWorkers.Done()
```
benjirewis, line 301: Just put the second defer in the first defer for clarity of ordering.

rpc/wrtc_server_stream.go
```diff
@@ -277,6 +277,10 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
 		}
 	}
 
+	s.ch.server.peerConnsMu.RLock()
+	s.ch.server.processHeadersWorkers.Add(1)
+	s.ch.server.peerConnsMu.RUnlock()
```
benjirewis, line 282: It's definitely possible my concurrent logic is not perfect here; would appreciate a couple pairs of eyes 😆 . The `RWMutex` + `WaitGroup` hurts my head, but I think it's probably the best way to synchronize this stuff.

rpc/wrtc_server_stream.go
```diff
@@ -277,13 +277,19 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
...
+	// Check if context has errored: underlying server may have been `Stop`ped,
+	// in which case we mark this processHeaders worker as `Done` and return.
+	if err := s.ch.server.ctx.Err(); err != nil {
+		s.ch.server.processHeadersWorkers.Done()
+		return
+	}
+
 	// take a ticket
 	select {
-	case <-s.ch.server.ctx.Done():
```
benjirewis, line 282: This `case` has been replaced by the `if` check above. I'm no longer `closeWithSendError`ing, as `Stop` should be closing all underlying peer connections. I can't quite remember why/if we thought this was a good/acceptable change, @dgottlieb; let me know if what I have seems incorrect.

dgottlieb: Looks good to me

dgottlieb (approved):

maximpertsov (approved):

PR #252, opened 2024-03-06
Correct logout path for fusionauth

michaellee1019:


ohEmily (approved):
LGTM! confirmed in the official [docs](https://fusionauth.io/docs/lifecycle/authenticate-users/oauth/endpoints) too.

PR #254, opened 2024-03-07
[APP-4184]: Add DisconnectWithReconnect for mongo shutdown simulation

maxhorowitz:
Motivation came from trying to manually tear down a mongo client in a testing suite and pull it back up after completion - the "pulling back up" failed because the disconnected mongo client was cached in memory. This PR adds handling for that case. Left inline comments / open questions to consider before going through with PR. Let me know what you think.

### DIFF
--- a/testutils/mongodb.go
+++ b/testutils/mongodb.go
@@ -97,8 +97,9 @@ var (
 
 // DisconnectWithReconnect closes sockets referenced by cachedBackingMongoDBClient
 // to simulate a db disconnection. The returned callback can be used to reinitialize
-// connection to the same db deployment. Use this function to validating behavior of
-// client shutdown scenarios.
+// connection to the same db deployment. Use this function to validate behavior of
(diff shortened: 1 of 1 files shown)


edaniels:
This change has the chance to break other tests that are running in parallel. How about in the code you're intending to use this in that you expose the backing MongoDB URI and make your own client? If you don't need the randomized namespaces support, you can construct a client yourself but if you need that behavior, maybe there could be a Clone or Standalone function that is specifically not shared and not cached. What do you think of that?

testutils/mongodb.go
```diff
@@ -95,6 +95,25 @@ var (
 	errCachedBackingMongoDBClient       error
 )
 
+// DisconnectWithReconnect closes sockets referenced by cachedBackingMongoDBClient
+// to simulate a db disconnection. The returned callback can be used to reinitialize
+// connection to the same db deployment. Use this function to validating behavior of
```
maxhorowitz, line 100: nit: `validate`

PR #256, opened 2024-03-08
[APP-3046] Post logout redirect URL update.

maxhorowitz:
Depending on auth provider, we need to populate post logout redirect URL with different query params.

### DIFF
--- a/web/auth0.go
+++ b/web/auth0.go
@@ -76,6 +76,7 @@ func InstallAuth0(
 	config AuthProviderConfig,
 	logger golog.Logger,
 ) (io.Closer, error) {
+	config.PostLogoutRedirectURL = "returnTo"
 	authProvider, err := installAuthProvider(
 		ctx,
 		sessions,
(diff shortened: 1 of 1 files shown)


web/auth0.go
```diff
@@ -106,10 +107,18 @@ func InstallFusionAuth(
...
 	logger golog.Logger,
 ) (io.Closer, error) {
+	overwrite := AuthProviderConfig{
+		Domain:                config.Domain,
+		ClientID:              config.ClientID,
+		Secret:                config.Secret,
+		BaseURL:               config.BaseURL,
+		PostLogoutRedirectURL: "post_logout_redirect_uri",
+		EnableTest:            config.EnableTest,
+	}
```
ohEmily, line 117: I think you can instead just do `config.PostLogoutRedirectURL = "post_logout_redirect_uri"`, no?

And then in `InstallAuth0`, add `config.PostLogoutRedirectURL = "returnTo"`.

Maybe I'm missing something?

ohEmily: Then you could remove this change if I understand correctly -- https://github.com/viamrobotics/app/pull/3982/files#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4R431-R433

maxhorowitz: yes

maxhorowitz: completely

ohEmily (approved):
LGTM, thank you!

PR #257, opened 2024-03-09
RSDK-6883 rpc: Replace RS256 with EdDSA as default signing algo

edaniels:
- And support multiple keys for verification and select one for signing

### DIFF
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/ed25519"
 	"crypto/rsa"
-
 	//nolint:gosec // using for fingerprint
 	"crypto/sha1"
 	"crypto/subtle"
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -6,8 +6,9 @@ import (
 	"crypto/rsa"
 
 	//nolint:gosec // using for fingerprint
-
+	"crypto/sha1"
 	"crypto/subtle"
+	"encoding/base64"
 	"errors"
 	"fmt"
 
@@ -117,7 +118,7 @@ func (p TokenVerificationKeyProviderFunc) Close(ctx context.Context) error {
 	return nil
 }
 
-// MakePublicKeyProvider returns a TokenVerificationKeyProvider that provides a public key for JWT verification.
+// MakeRSAPublicKeyProvider returns a TokenVerificationKeyProvider that provides a public key for JWT verification.
 func MakeRSAPublicKeyProvider(pubKey *rsa.PublicKey) TokenVerificationKeyProvider {
 	return TokenVerificationKeyProviderFunc(
 		func(ctx context.Context, token *jwt.Token) (interface{}, error) {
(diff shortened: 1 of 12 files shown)


rpc/auth.go
```diff
@@ -256,18 +270,6 @@ type Credentials struct {
 	Payload string          `json:"payload"`
 }
 
-// RSAPublicKeyThumbprint returns SHA1 of the public key's modulus Base64 URL encoded without padding.
-func RSAPublicKeyThumbprint(key *rsa.PublicKey) (string, error) {
```
maximpertsov, line 260: we should deprecate instead of removing - app is using this method.

maximpertsov (commented):
looks good generally - we need to update the two examples (echo + fileupload) to use `ed25519`

rpc/auth.go
edaniels: it's using it via goutils/rpc, not app

edaniels: ah right

edaniels:
@maximpertsov ready again!

maximpertsov (approved):
LGTM as far as how this affects the RDK server. [Added a reviewer from fleet](https://github.com/viamrobotics/goutils/pull/257#issuecomment-1995052863) to make sure there are no concerns app-side.

michaellee1019:
Sorry for the delay. Makes sense. Will we be running both RSA + EdDSA on viam's authentication backend for some time? I think we'll need to respect the already issued JWTs for some time. And then the plan is to switch to RDK to use EdDSA in the next release?

edaniels:
@michaellee1019 what this PR does is sign new signatures in EdDSA but continues to accept all valid signature types. That means pre-deploy you have RSA signed JWTs. Post-deploy you have a mix. Weeks later without any deploy due to JWT expiration, you will have all EdDSA

edaniels:
> @michaellee1019 what this PR does is sign new signatures in EdDSA but continues to accept all valid signature types. That means pre-deploy you have RSA signed JWTs. Post-deploy you have a mix. Weeks later without any deploy due to JWT expiration, you will have all EdDSA

Actually for app I think we can stick to RSA right now. I need to add back in support here. I was only considering RDK signed JWTs at the moment. I'll get to this this week.

michaellee1019:
> Actually for app I think we can stick to RSA right now. I need to add back in support here. I was only considering RDK signed JWTs at the moment. I'll get to this this week.

Ok cool. TY!

edaniels:
@maximpertsov @michaellee1019 need a re-review. I've now added support for multiple keys to be added and let the user select a key to use for signing if more than one has been provided. I realized it was silly to just support either/or and this wasn't much more code to add. This will allow the app server to utilize multiple verification keys and select a new signing key/method in the future which is good for security + flexibility!

Will add tests later this week and verify rdk/app acceptance.

edaniels:
Tested this works when upgrading RDK and App (requires a one line change due to breaking API). RDK doesn't really matter much since every server restart is already a new private key. App will continue to work as is but can be extended to support both types of keys in the future for a no-downtime swap to EdDSA (if desired).

PR #260, opened 2024-03-12
[RSDK-6030] add Status() to ManagedProcess

mattjperez:
This change allows managed processes to check if their process is still running.
~This method (and the underlying Signal call from golang stdlib) is not supported on windows.~
Correction: Signal Interrupt is not supported on windows

### DIFF
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -31,12 +31,11 @@ type ManagedProcess interface {
 
 	// Stop signals and waits for the process to stop. An error is returned if
 	// there's any system level issue stopping the process.
-	Stop() error
+  Stop() error
 
   // Status return nil when the process is both alive and owned.
   // If err is non-nil, process may be a) alive but not owned or b) dead.
-  // Method not supported/implemented on Windows
-	Status() error
+  Status() error 
 }
 
 // NewManagedProcess returns a new, unstarted, from the given configuration.
@@ -110,8 +109,7 @@ func (p *managedProcess) ID() string {
 func (p *managedProcess) Status() error {
 	p.mu.Lock()
   defer p.mu.Unlock()
-  err := p.cmd.Process.Signal(syscall.Signal(0))
-  return err
+  return p.cmd.Process.Signal(syscall.Signal(0))
 }
 
 func (p *managedProcess) Start(ctx context.Context) error {
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -32,6 +32,11 @@ type ManagedProcess interface {
 	// Stop signals and waits for the process to stop. An error is returned if
 	// there's any system level issue stopping the process.
 	Stop() error
+
+  // Status return nil when the process is both alive and owned.
+  // If err is non-nil, process may be a) alive but not owned or b) dead.
+  // Method not supported/implemented on Windows
+	Status() error
 }
 
 // NewManagedProcess returns a new, unstarted, from the given configuration.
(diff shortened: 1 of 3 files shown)


pexec/managed_process.go
```diff
@@ -102,6 +107,13 @@ func (p *managedProcess) ID() string {
 	return p.id
 }
 
+func (p *managedProcess) Status() error {
+	p.mu.Lock()
+  defer p.mu.Unlock()
+  err := p.cmd.Process.Signal(syscall.Signal(0))
```
benjirewis, line 113: What if we wrote a `status` method in each of `managed_process_unix.go` and `managed_process_windows.go` where the former does `p.cmd.Process.Signal(syscall.Signal(0))` and the latter just `return errors.New("status not available on windows, signals not supported")`?

pexec/managed_process_test.go
```diff
@@ -182,7 +182,13 @@ func TestManagedProcessStart(t *testing.T) {
 
 			<-watcher.Events
 
-			test.That(t, proc.Stop(), test.ShouldBeNil)
+      if runtime.GOOS != "windows" {
+        test.That(t, proc.Status(), test.ShouldBeNil)
+      }
+      test.That(t, proc.Stop(), test.ShouldBeNil)
+      if runtime.GOOS != "windows" {
+        test.That(t, proc.Status(), test.ShouldNotBeNil)
```
benjirewis, line 190: We should probably assert something about what these errors look like.

pexec/managed_process_test.go
```diff
@@ -699,3 +715,9 @@ func (fp *fakeProcess) Stop() error {
 	}
 	return nil
 }
+func (fp *fakeProcess) Status() error {
+	if fp.stopErr || fp.startErr {
+		return errors.New("dead")
```
benjirewis, line 720: 💀

benjirewis (commented):
Looking good! Just a few comments.

pexec/managed_process.go
mattjperez: Actually, I think I misread the golang docs about this. As this is just a signal and not an Interrupt, this should be compatible with windows as well. 

https://pkg.go.dev/os#Process.Signal

mattjperez: So we should be okay keeping it in the main managed process file, without the warning

pexec/managed_process_test.go
```diff
@@ -182,7 +182,9 @@ func TestManagedProcessStart(t *testing.T) {
 
 			<-watcher.Events
 
+			test.That(t, proc.Status(), test.ShouldBeNil)
 			test.That(t, proc.Stop(), test.ShouldBeNil)
+			test.That(t, proc.Status(), test.ShouldNotBeNil)
```
benjirewis, line 187: Can we assert anything about this error? What does it look like? I'm curious if we should be retuning an error at all if it's not a useful one, and whether `Status` should be named `Alive` and return a bool.

mattjperez: My bad, I thought I added those changes. Though it may have been in the related PR draft in rdk. 

On error it can return things like 
- operation not permitted (when not owned(
- process does not exist
- process already exited (as in the test in the RDK pr)

mattjperez: If you think the above errors don't add any value then yeah, I think Alive may be more intuitive

benjirewis: Cool those error messages are def helpful;`Status` and returning an error seems correct, but let's assert that the messages are what we expect in tests here 👍🏻 .

pexec/managed_process.go
benjirewis: Nice; SGTM!

PR #261, opened 2024-03-13
RSDK-6917: Have ClientConn objects expose access to underlying PeerConnection, if applicable.

dgottlieb:


rpc/dialer.go
```diff
@@ -68,6 +73,15 @@ type ClientConnAuthenticator interface {
 	Authenticate(ctx context.Context) (string, error)
 }
 
+// A GrpcOverHttpClientConn is grpc connection that is not backed by a `webrtc.PeerConnection`.
+type GrpcOverHttpClientConn struct {
```
dgottlieb, line 77: This is public because there's a place in the RDK that does a direct grpc dial that I need to instrument. I figured it'd be appropriate to reuse this. Even if the type is buried pretty deep.

benjirewis (approved):
LGTM but

> beat lint before lint beats me

The linter is still complaining 😆 . I think you're missing a `PeerConn` impl in some test file.

PR #262, opened 2024-03-13
RSDK-5534 Better synchronize answer workers

benjirewis:
RSDK-5534

Better synchronize active goroutines in `wrtc_signaling_answerer.go`. Very similar to change in https://github.com/viamrobotics/goutils/pull/251.

- Renames `activeBackgroundWorkers` to `answerWorkers`
- Renames `cancelBackgroundWorkers` to `cancelAnswerWorkers`
- Adds an `answerMu` `RWMutex` to be `RLock`ed when adding an answer worker and `Lock`ed when waiting on answer workers

The linked ticket describes a data race where `activeBackgroundWorkers.Add(1)` was racing with `activeBackgroundWorkers.Wait()`. This was happening because the `Add` was in a separate goroutine from the `Wait` due to the asynchronous nature of our answering process, and there was no synchronization of `Add`s and `Wait`s.

### DIFF
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -229,8 +229,8 @@ func (ans *webrtcSignalingAnswerer) Stop() {
 
 	ans.cancelAnswerWorkers()
 	ans.answerMu.Lock()
+	defer ans.answerMu.Unlock()
 	ans.answerWorkers.Wait()
-	ans.answerMu.Unlock()
 }
 
 // answer accepts a single call offer, responds with a corresponding SDP, and
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -29,15 +29,21 @@ const testDelayAnswererNegotiationVar = "TEST_DELAY_ANSWERER_NEGOTIATION"
 type webrtcSignalingAnswerer struct {
 	startStopMu sync.Mutex // startStopMu guards the Start and Stop methods so they do not happen concurrently.
 
-	address                 string
-	hosts                   []string
-	server                  *webrtcServer
-	dialOpts                []DialOption
-	webrtcConfig            webrtc.Configuration
-	activeBackgroundWorkers sync.WaitGroup
-	cancelBackgroundWorkers func()
-	closeCtx                context.Context
-	logger                  golog.Logger
+	address      string
+	hosts        []string
+	server       *webrtcServer
+	dialOpts     []DialOption
+	webrtcConfig webrtc.Configuration
+
+	// answererMu should be `Lock`ed in `Stop` to `Wait` on ongoing answer workers in startAnswerer/answer.
+	// answererMu should be `RLock`ed when starting a new answer worker (allow concurrent answer workers) to
+	// `Add` to answerWorkers.
+	answerMu            sync.RWMutex
+	answerWorkers       sync.WaitGroup
+	cancelAnswerWorkers func()
+
+	closeCtx context.Context
+	logger   golog.Logger
 }
 
 // newWebRTCSignalingAnswerer makes an answerer that will connect to and listen for calls at the given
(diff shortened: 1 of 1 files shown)


rpc/wrtc_signaling_answerer.go
```diff
@@ -29,15 +29,21 @@ const testDelayAnswererNegotiationVar = "TEST_DELAY_ANSWERER_NEGOTIATION"
...
+	address      string
+	hosts        []string
+	server       *webrtcServer
+	dialOpts     []DialOption
+	webrtcConfig webrtc.Configuration
+
+	// answererMu should be `Lock`ed in `Stop` to `Wait` on ongoing answer workers in startAnswerer/answer.
+	// answererMu should be `RLock`ed when starting a new answer worker (allow concurrent answer workers) to
+	// `Add` to answerWorkers.
+	answerMu            sync.RWMutex
```
maximpertsov, line 41: q: can this mutex replace `startStopMu`? or is it necessary to have separate mutexes for starting/stopping vs in-progress answers.

maximpertsov: > `Start` should never interfere with in progress answers

q:  just to clarify - do mean _should_ as in start and in-progress will never interfere w/ each other, or they might and it's a requirement that they don't? if the former than i think one mutex makes things simpler/clearer, if the latter then separate mutexes makes sense.

benjirewis: Update: `Start` and `startAnswerer` do indeed interfere with each other. If we were to use one mutex, `Start` would `Lock` it, call `startAnswerer`, and `startAnswerer` would `Rock()` the same lock to `Add` to the `WaitGroup` and deadlock the system 😢 . So, I think we need two muteness.

maximpertsov: gotcha, thanks for looking into this. makes sense!

maximpertsov (approved):

rpc/wrtc_signaling_answerer.go
```diff
@@ -145,7 +151,9 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 		}
 		return answerClient, nil
 	}
-	ans.activeBackgroundWorkers.Add(1)
+	ans.answerMu.RLock()
```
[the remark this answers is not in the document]
benjirewis: > This patch right now doesn't have that strong invariant. Which makes proving the patch, as-is, doesn't have a bug much more "fun" than it needs to be.

An excellent description of the issue here hahaha. You're totally right that this patch is not enforcing as _strong_ of an invariant as the last; let me add a context and see what you think.

rpc/wrtc_signaling_answerer.go
```diff
@@ -210,8 +227,10 @@ func (ans *webrtcSignalingAnswerer) Stop() {
 	ans.startStopMu.Lock()
 	defer ans.startStopMu.Unlock()
 
-	ans.cancelBackgroundWorkers()
-	ans.activeBackgroundWorkers.Wait()
+	ans.cancelAnswerWorkers()
+	ans.answerMu.Lock()
+	ans.answerWorkers.Wait()
+	ans.answerMu.Unlock()
```
maximpertsov, line 233: super minor - noticed this flow is ever-so-slightly different in the corresponding [caller PR](https://github.com/viamrobotics/goutils/pull/265/files#r1526670921). is that intentional? should we standardize?

benjirewis: Not intentional! I'll standardize to the `defer`.

rpc/wrtc_signaling_answerer.go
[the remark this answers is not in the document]
maximpertsov: > Because I claim we'll always hit the client == nil check on L160. Proving that is left to the reader.

heh, still working through this one... but thank you for outlining the bug we are vulnerable to if `client == nil` did not save us. 

I agree that implementing the invariant you describe would make it easier to reason that there is no such bad flow. I'm also okay leaving as is if implementing said invariant leads to significant rework.

dgottlieb (approved):

rpc/wrtc_signaling_answerer.go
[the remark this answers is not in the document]
benjirewis: I _think_ I enforced the stronger variant with the context checks I added.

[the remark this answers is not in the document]
dgottlieb: Yep, I agree the code succeeded in doing that.

PR #265, opened 2024-03-15
RSDK-5690 Better synchronize call workers and Close

benjirewis:
RSDK-5690

Better synchronize active goroutines in `wrtc_signaling_server.go`. Very similar to change in https://github.com/viamrobotics/goutils/pull/251 and https://github.com/viamrobotics/goutils/pull/262.

- Renames `activeBackgroundWorkers` to `callWorkers`
- Adds a `callMu` `RWMutex` to be `RLock`ed when adding a call worker and `Lock`ed when waiting on call workers

The linked ticket describes a data race where `activeBackgroundWorkers.Add(1)` was racing with `activeBackgroundWorkers.Wait()`. This was happening because the `Add` was in a separate goroutine from the `Wait` due to the asynchronous nature of our call process, and there was no synchronization of `Add`s and `Wait`s.

### DIFF
--- a/rpc/wrtc_signaling_server.go
+++ b/rpc/wrtc_signaling_server.go
@@ -35,10 +35,15 @@ type WebRTCSignalingServer struct {
 	webrtcConfigProvider WebRTCConfigProvider
 	forHosts             map[string]struct{}
 
-	activeBackgroundWorkers sync.WaitGroup
-	cancelCtx               context.Context
-	cancelFunc              func()
-	logger                  golog.Logger
+	// callMu should be `Lock`ed in `Close` to `Wait` on ongoing Calls (incoming
+	// callers). callMu should be `RLock`ed in processHeaders (allow concurrent
+	// processHeaders) to `Add` to processHeadersWorkers.
+	callMu      sync.RWMutex
+	callWorkers sync.WaitGroup
+
+	cancelCtx  context.Context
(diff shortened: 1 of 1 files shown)


rpc/wrtc_signaling_server.go
```diff
@@ -508,5 +523,7 @@ func (srv *WebRTCSignalingServer) OptionalWebRTCConfig(
 // Close cancels all active workers and waits to cleanly close all background workers.
 func (srv *WebRTCSignalingServer) Close() {
 	srv.cancelFunc()
-	srv.activeBackgroundWorkers.Wait()
+	srv.callMu.Lock()
+	defer srv.callMu.Unlock()
+	srv.callWorkers.Wait()
```
maximpertsov, line 528: super minor - i noticed we `Wait -> Unlock` without a `defer` in the corresponding [caller PR](https://github.com/viamrobotics/goutils/pull/262/files#diff-5514904bc577d9192b1f4805ade57d369a09c437b0bc2036986efa4be3b2bd0bR230-R233). any reason this one should be different (or vice versa)?

benjirewis: Not intentional! I'll standardize the other PR to a `defer` :) good catch.

maximpertsov (approved):

PR #266, opened 2024-03-15
[RSDK-6030] Status() rewrite to IsRunning()

mattjperez:


pexec/managed_process_test.go
```diff
@@ -184,7 +191,7 @@ func TestManagedProcessStart(t *testing.T) {
 
 			test.That(t, proc.Status(), test.ShouldBeNil)
 			test.That(t, proc.Stop(), test.ShouldBeNil)
-			test.That(t, proc.Status().Error(), test.ShouldContainSubstring, "process already finished")
```
mattjperez, line 187: this is what using `Signal(0)` returns on unix.

pexec/managed_process.go
```diff
@@ -109,7 +109,23 @@ func (p *managedProcess) ID() string {
 func (p *managedProcess) Status() error {
```
cheukt, line 109: this api seems unwieldy - if anyone wants to extract the exit code, they have to parse the error. I feel like an enum might make more sense (states for running, haven't started, exited, signaled, etc) if we want to keep `Status()` as the  method name. We would lose exit code, but not the most important for our purposes. Or just `isAlive() bool` to simplify it even more

mattjperez: great points, I agree. Taking this all the way to clear enums and the parsing/handling that would be necessary to map things correctly across architectures, OSs, etc.

I changed it to `IsRunning()`. This also covers the case where the `cmd` is not set for some reason.

benjirewis (approved):

pexec/managed_process.go
```diff
@@ -358,7 +373,6 @@ func (p *managedProcess) Stop() error {
 		p.mu.Unlock()
 		return nil
 	}
-	p.mu.Unlock()
```
benjirewis, line 361: Why has our section of `Lock`ed behavior expanded? And, if it's valid to expand it, I think you should just `defer p.mu.Unlock()` here instead of adding an `Unlock` to every return path.

mattjperez: I tried defer initially. To prevent a race condition flagged by CI, but it still didn't pass. I did this just to be explicit and see if there was a difference and now it passes with this.

mattjperez: Need to do further testing and see how I can pass CI with minimal changes. @maximpertsov is also tracking

cheukt (commented):
idea looks good, thanks for updating the interface! will hold off on approving until RDK pr is in and can evaluate at the same time

mattjperez:
Handing off to @maximpertsov

PR #267, opened 2024-03-16
Disable audience check.

dgottlieb:


rpc/server_auth.go
```diff
@@ -359,6 +359,7 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (context.Context, erro
 			break
 		}
 	}
+	audVerified = true
```
erh, line 362: if you have to push again anyway ,add a log if this is false and we flip it, and maybe log claims.RegistedClaims, and ss.authaudience, but minor

erh: if !audVerified {
   audVerified = true
   ss.logger.Infof("hack %v %v", claims.RegistedClaims, ss.authaudience)
}

erh (approved):

PR #268, opened 2024-03-16
stop swallowing error in logline

ohEmily:


rpc/server_auth.go
```diff
@@ -360,22 +360,44 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (context.Context, erro
 		}
 	}
 	if !audVerified {
+		// TODO cleanup at a later date -- urgent issue
 		audVerified = true
-		ss.logger.Infof("hack %v %v", claims.RegisteredClaims, ss.authAudience)
+		ss.logger.Infow("hack %v %v",
+			"registeredClaims", claims.RegisteredClaims,
+			"authAudience", ss.authAudience)
```
ohEmily, line 367: Screenshot 2024-03-16 at 11 20 02 AM

not sure how useful this line given the image above but not going to take it out in case it ends up being helpful

erh (approved):

rpc/server_auth.go
erh: I’d leave it all in until we fix and fully understand

PR #269, opened 2024-03-18
Remove bad return value

ohEmily:
Accidentally copy-pasted a return where we should not be returning, only logging.

Fixes [this issue](https://viaminc.slack.com/archives/C01Q5S8V6TB/p1710689886947519).

### DIFF
--- a/rpc/server_auth.go
+++ b/rpc/server_auth.go
@@ -374,12 +374,11 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (context.Context, erro
 				"error", err,
 				"empty expected audience", len(ss.authAudience) == 0,
 				"expected audience list", audienceList)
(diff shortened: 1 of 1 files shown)


rpc/server_auth.go
```diff
@@ -374,12 +374,11 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (context.Context, erro
 				"error", err,
 				"empty expected audience", len(ss.authAudience) == 0,
 				"expected audience list", audienceList)
-			//return nil, status.Error(codes.Unauthenticated, "invalid audience")
```
ohEmily, line 377: This was a mistake. Eliot tried commenting it out but it seems like the build didn't like that.

michaellee1019 (approved):

PR #270, opened 2024-03-18
RSDK-6490 Simplify mDNS dial logic to return either an error or a valid connection

maximpertsov:
Remove logic branches from mDNS dial that return a `nil` connection without an error

### DIFF
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -138,15 +138,10 @@ func dial(
 		go func(dOpts dialOptions) {
 			defer wg.Done()
 			conn, cached, err := dialMulticastDNS(ctxParallel, address, logger, dOpts)
-			switch {
-			case err != nil:
+			if err != nil {
 				dialCh <- dialResult{err: err}
-			case conn != nil:
-				// TODO(RSDK-6490): investigate why we can get a `nil` connection that is
-				// not accompanied with an error.
+			} else {
 				dialCh <- dialResult{conn: conn, cached: cached}
-			default:
-				dialCh <- dialResult{err: errors.New("no connection")}
 			}
 		}(dOpts)
 	}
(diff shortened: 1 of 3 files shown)


rpc/dial.go
```diff
@@ -278,42 +273,45 @@ func dial(
...
 			}
 		}
+	}
+	if ctx.Err() != nil {
 		return nil, ctx.Err()
 	}
-	// lookup for candidates for backward compatibility
-	entry, err := candidateLookup(ctx, candidates)
-	if err != nil || entry == nil {
+	return nil, errors.New("mDNS query failed to find a candidate")
```
maximpertsov, line 304: This extra guard might be unnecessary - it's not clear to me if it's practically possible to not find an entry (or find a `nil` entry) without a context error/completion.

rpc/server_auth.go
```diff
@@ -374,7 +374,8 @@ func (ss *simpleServer) ensureAuthed(ctx context.Context) (context.Context, erro
 				"error", err,
 				"empty expected audience", len(ss.authAudience) == 0,
 				"expected audience list", audienceList)
-			//return nil, status.Error(codes.Unauthenticated, "invalid audience")
+			// TODO: re-enable once auth migration is complete
+			// return nil, status.Error(codes.Unauthenticated, "invalid audience")
```
maximpertsov, line 378: flyby: had to add a space after `//` to avoid a linter crash/deletion. added a TODO to clarify why we left this code in as a comment.

maximpertsov: ignore this - the issue was resolved in `main`

rpc/dial.go
```diff
@@ -278,42 +273,45 @@ func dial(
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		// entries gets closed after lookupCtx expires or there is a real entry
+		case entry := <-entries:
+			if entry != nil {
+				return entry, nil
 			}
 		}
+	}
+	if ctx.Err() != nil {
```
benjirewis, line 301: Did you have a reason to add this extra context check? I don't think it will hurt, but curious if you were consciously thinking about a case where after examining candidates the context has errored?

rpc/dial.go
```diff
@@ -329,7 +327,7 @@ func dialMulticastDNS(
 
 	// IPv6 with scope does not work with grpc-go which we would want here.
 	if !(hasGRPC || hasWebRTC) || len(entry.AddrIPv4) == 0 {
-		return nil, false, nil
+		return nil, false, errors.New("mDNS query failed to find an IPv4 candidate")
```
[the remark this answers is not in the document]
benjirewis: I talked with Ethan a bit offline; rust-utils has a slightly different architecture (in particular, I don't think parallel mDNS/WebRTC dials have been implemented yet) for how it returns mDNS connections, but [in the analogous piece of code](https://github.com/viamrobotics/rust-utils/blob/main/src/rpc/dial.rs#L364-L371), it returns `None` and doesn't really report an error. I think returning an error here is ultimately just more, helpful information, but if we do this, we should probably try to get the same behavior in rust.

rpc/dial.go
maximpertsov: [commented here](https://github.com/viamrobotics/goutils/pull/270/files#r1528461649)

mainly out of caution - it guards against the following scenario:

* select <-entries (meaning an entry is found before context is done)
* entry is nil (L296) suggests this is a possible
* return ctx.Err(), but it's still nil, so we return `nil, nil`

maybe there's some logical reason why this can't happen, but it's not clear to me that it can't happen.

benjirewis: Sorry missed your other comment for some reason; cool, it seems fine to include this extra check IMO.

rpc/dial.go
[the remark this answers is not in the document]
maximpertsov: gotcha, thanks for that data point... I agree we should make these behave the same.

my general understanding is that the rest of the dial logic in goutils does not support ipv6 addresses, so we might as well return an error (and maybe export that error) - otherwise downstream code becomes more complicated since it has to handle "no error but also no connection cases".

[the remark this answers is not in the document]
maximpertsov: actually, I guess the error description isn't quite comprehensive - we also return `nil, nil` when the candidate we find has neither grpc nor webrtc as a text field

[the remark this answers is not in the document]
cheukt: ya, think we should update the error to say `failed to find an IPv4 candidate or the candidate we found has neither grpc nor webrtc as a text field`

cheukt (approved):
generally looks good to me

rpc/dial.go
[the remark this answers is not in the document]
maximpertsov: added a custom error for this case: https://github.com/viamrobotics/goutils/pull/270/commits/f2020141d3d02b6365cdd7a8762089e0a23ccbb1

PR #273, opened 2024-03-22
RSDK-7062: Make webrtc negotiation related helpers public.

dgottlieb:


nicksanford (approved):

rpc/wrtc_peer.go
```diff
@@ -188,7 +188,7 @@ func newPeerConnectionForServer(
 	}
 	dataChannel.OnError(initialDataChannelOnError(peerConn, logger))
 
-	if _, err = configureForRenegotiation(peerConn, logger); err != nil {
+	if _, err = ConfigureForRenegotiation(peerConn, logger); err != nil {
```
dgottlieb, line 191: Interesting that lsp-rename caught these but not the encode/decode

dgottlieb: Oh wait, it did. I just didn't save the file :1st_place_medal:

PR #276, opened 2024-04-02
RSDK-6130 Improve error message if CWD of process does not exist

maximpertsov:
If current working directory (CWD) is provided, make sure it exists before starting the managed process. If it doesn't or there is some other error related to looking up the CWD, return a more descriptive error.

### DIFF
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -128,7 +128,7 @@ func (p *managedProcess) Start(ctx context.Context) error {
 	}
 
 	if p.cwd != "" {
-		if _, err := os.Stat(p.cwd); err != nil {
+		if _, err := os.Lstat(p.cwd); err != nil {
 			return fmt.Errorf("error with current working directory: %w", err)
 		}
 	}
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -128,7 +128,7 @@ func (p *managedProcess) Start(ctx context.Context) error {
 	}
 
 	if _, err := os.Stat(p.cwd); err != nil {
-		return err
+		return fmt.Errorf("error with current working directory: %w", err)
 	}
 
 	if _, err := exec.LookPath(p.name); err != nil {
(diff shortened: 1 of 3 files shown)


pexec/managed_process.go
```diff
@@ -127,6 +127,12 @@ func (p *managedProcess) Start(ctx context.Context) error {
 	default:
 	}
 
+	if p.cwd != "" {
+		if _, err := os.Stat(p.cwd); err != nil {
```
dgottlieb, line 131: I think we want this to be `Lstat` rather than stat. Such that an existing, but broken symlink generates an error. But this is just in the name of better error messaging, so definitely not necesssary.

pexec/managed_process.go
```diff
@@ -127,6 +127,12 @@ func (p *managedProcess) Start(ctx context.Context) error {
...
+			return fmt.Errorf("error with current working directory: %w", err)
```
dgottlieb, line 132: Given the finicky nature of relative paths, I'd expect this error message to have some more context:
* The process [working directory](https://pkg.go.dev/os#Getwd) (or the error string that method can return)
* The `p.cwd` itself

So if there is an error, a human can determine what path we attempted to check for existence.

pexec/managed_process.go
maximpertsov: it's a small enough change, happy to make a flyby improvement

pexec/managed_process.go
maximpertsov: nevermind, it's obviously helpful in the case relative paths

so now it's a matter of naming everything in a way that's not confusing, since we have a "present" cwd and "future" cwd :upside_down_face:

maximpertsov: https://github.com/viamrobotics/goutils/pull/276/commits/5a5d3907291531729bb7fa653df6438792aef6cf

maximpertsov:
failed my own unit test - TDD, obviously :smirk:

dgottlieb (approved):

PR #277, opened 2024-04-02
RSDK-6493 clarify early exit vs direct dial fallback

maximpertsov:


rpc/dial.go
```diff
@@ -236,7 +258,10 @@ func dial(
 			if conn != nil {
 				errClose := conn.Close()
 				if errClose != nil {
-					logger.Warnw("unable to close redundant connection", "error", err)
+					logger.Warnw(
+						"unable to close redundant connection",
+						"error", errClose,
+					)
```
maximpertsov, line 264: flyby: was logging the wrong error here :disappointed:

rpc/dial.go
```diff
@@ -207,12 +214,27 @@ func dial(
...
+				}
 				dialCh <- dialResult{err: ctxParallel.Err(), skipDirect: true}
+			case errors.Is(err, ErrNoWebRTCSignaler):
+				// There is no WebRTC signaler, so dialing WebRTC is expected to fail.
+				// We should still try to dial directly as a fallback.
+				dialCh <- dialResult{err: err}
+			case isUnauthenticatedError(err):
+				// We are our not authenticated to dial the host, so we should not
+				// attempt to dial directly either.
+				dialCh <- dialResult{err: err, skipDirect: true}
```
maximpertsov, line 233: Some tests break without this check, but they contain scenarios where webrtc dial requires authentication, but direct dial does not.

rpc/dial.go
```diff
@@ -259,9 +279,11 @@ func dial(
...
 		logger.Debugw("trying direct", "address", address)
 	}
-	conn, cached, err = dialDirectGRPC(ctx, address, dOpts, logger)
-	if err != nil {
-		return nil, false, err
+
+	var directErr error
+	conn, cached, directErr = dialDirectGRPC(ctx, address, dOpts, logger)
+	if directErr != nil {
+		return nil, false, multierr.Combine(directErr, nonFatalErr)
```
maximpertsov, line 286: flyby: include error(s) from initial dial attempt if we fail to dial directly as a fallback

cheukt (approved):

PR #280, opened 2024-04-15
RSDK-7010 Invalidate JWT access token upon Unauthenticated error

benjirewis:
RSDK-7010

Adds a unary client interceptor to clear the `accessToken` set on `perRPCJWTCredentials` in the event of an error containing the substring `"Unauthenticated"`. Clearing the `accessToken` causes an attempt to reauthenticate on the next RPC request. Adds a test for this behavior.

### DIFF
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -3,6 +3,7 @@ package rpc
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"go.opencensus.io/trace"
 	"google.golang.org/grpc"
@@ -44,4 +45,20 @@ func contextWithSpanMetadata(ctx context.Context) context.Context {
 		"trace-options", fmt.Sprint(span.SpanContext().TraceOptions),
 	)
 	return ctx
+}
+
+// UnaryClientReauthInterceptor clears the access token stored on creds in the
+// event of an "Unauthenticated" gRPC error to force re-auth.
+func UnaryClientReauthInterceptor(creds *perRPCJWTCredentials) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{},
+		cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption,
+	) error {
+		err := invoker(ctx, method, req, reply, cc, opts...)
+		if err != nil && strings.Contains(err.Error(), "Unauthenticated") {
+			if creds != nil {
+				creds.accessToken = ""
+			}
+		}
+		return err
+	}
 }
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -60,6 +60,7 @@ func UnaryClientReauthInterceptor(creds *perRPCJWTCredentials) grpc.UnaryClientI
 			if creds != nil {
 				creds.accessToken = ""
 			}
+		default:
 		}
 		return err
 	}
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -48,9 +48,10 @@ func contextWithSpanMetadata(ctx context.Context) context.Context {
 	return ctx
 }
 
-// UnaryClientReauthInterceptor clears the access token stored on creds in the
-// event of an "Unauthenticated" or "PermissionDenied" gRPC error to force re-auth.
-func UnaryClientReauthInterceptor(creds *perRPCJWTCredentials) grpc.UnaryClientInterceptor {
+// UnaryClientInvalidAuthInterceptor clears the access token stored on creds in
+// the event of an "Unauthenticated" or "PermissionDenied" gRPC error to force
+// re-auth.
+func UnaryClientInvalidAuthInterceptor(creds *perRPCJWTCredentials) grpc.UnaryClientInterceptor {
 	return func(ctx context.Context, method string, req, reply interface{},
 		cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption,
 	) error {
--- a/rpc/client_interceptors.go
+++ b/rpc/client_interceptors.go
@@ -3,11 +3,12 @@ package rpc
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"go.opencensus.io/trace"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 )
 
 // UnaryClientTracingInterceptor adds the current Span's metadata to the context.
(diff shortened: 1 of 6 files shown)


rpc/client_interceptors.go
```diff
@@ -45,3 +46,19 @@ func contextWithSpanMetadata(ctx context.Context) context.Context {
...
 }
+
+// UnaryClientReauthInterceptor clears the access token stored on creds in the
+// event of an "Unauthenticated" gRPC error to force re-auth.
+func UnaryClientReauthInterceptor(creds *perRPCJWTCredentials) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{},
+		cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption,
+	) error {
+		err := invoker(ctx, method, req, reply, cc, opts...)
+		if err != nil && strings.Contains(err.Error(), "Unauthenticated") {
```
benjirewis, line 57: I don't _love_ just checking for the string "Unauthenticated", and I'd rather check for the gRPC error codes 7 and 12. I tried casting to a gRPC status locally but had some trouble. Will keep looking into it, but let me know if you're familiar with how to extract the gRPC error code from a golang `error`.

rpc/dialer.go
```diff
@@ -290,17 +290,13 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts dialOptions, logg
...
-	unaryInterceptor := grpc_middleware.ChainUnaryClient(unaryInterceptors...)
-	dialOpts = append(dialOpts, grpc.WithUnaryInterceptor(unaryInterceptor))
 
 	var streamInterceptors []grpc.StreamClientInterceptor
 	streamInterceptors = append(streamInterceptors, grpc_zap.StreamClientInterceptor(grpcLogger))
 	streamInterceptors = append(streamInterceptors, StreamClientTracingInterceptor())
 	if dOpts.streamInterceptor != nil {
 		streamInterceptors = append(streamInterceptors, dOpts.streamInterceptor)
 	}
-	streamInterceptor := grpc_middleware.ChainStreamClient(streamInterceptors...)
```
benjirewis, line 302: I also moved `streamInterceptor`s creation to below the auth logic to not weirdly have stream interceptor creation here and unary interceptor creation somewhere else.

rpc/dialer.go
```diff
@@ -335,8 +331,17 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts dialOptions, logg
 			connPtr = &rpcCreds.conn
 		}
 		dialOpts = append(dialOpts, grpc.WithPerRPCCredentials(rpcCreds))
+		unaryInterceptors = append(unaryInterceptors, UnaryClientReauthInterceptor(rpcCreds))
+		// ReauthInterceptors will not work for streaming calls; we can only
+		// intercept the creation of a stream, and the ensuring of authentication
+		// server-side happens per RPC request (per usage of the stream).
```
benjirewis, line 337: Discussed this with @cheukt offline; I can't find a great way to clear the `accessToken` _per usage_ of the gRPC stream. This means that if `streamClient.RecvMsg` returns an unauthenticated error, all subsequent calls to `RecvMsg` will return the same error.

rpc/server_options.go
```diff
@@ -358,6 +362,15 @@ func WithAuthHandler(forType CredentialsType, handler AuthHandler) ServerOption
 	})
 }
 
+// WithEnsureAuthedHandler returns a ServerOptions which adds custom logic for
+// the ensuring of authentication on each incoming request.
+func WithEnsureAuthedHander(eah func(ctx context.Context) (context.Context, error)) ServerOption {
```
benjirewis, line 367: Had to add this to inject into `ensureAuthed`/examine its behavior. If we're nervous about making a method like this public, I can _probably_ find another way.

rpc/client_interceptors.go
benjirewis: That'll do it! Thank you 🖖🏻 .

maximpertsov: Yeah we have some examples of this in tests... we do something like usually:

```go
       gStatus, ok := status.FromError(err)
       return ok && gStatus.Code() == codes.Unauthenticated
```

rpc/client_interceptors.go
```diff
@@ -45,3 +47,21 @@ func contextWithSpanMetadata(ctx context.Context) context.Context {
 	)
 	return ctx
 }
+
+// UnaryClientReauthInterceptor clears the access token stored on creds in the
```
cheukt, line 51: ReauthInterceptor implies that it actively reauths for the caller, which this doesn't, so I'd like a different name - trying to think of other still succinct names but not succeeding.

maybe `ClearInvalidAuthInterceptor`? or even just `AuthInterceptor`?

rpc/server_auth.go
```diff
@@ -284,6 +284,11 @@ func (ss *simpleServer) tryAuth(ctx context.Context) (context.Context, error) {
 }
 
 func (ss *simpleServer) ensureAuthed(ctx context.Context) (context.Context, error) {
+	// Use handler if set (only used for testing).
```
cheukt, line 287: if this really should only be used during testing, should we add a guard around this? or maybe the comment should be `(should only be used for testing)`

cheukt (changes requested):
changes look good, but do want to see a different name. I think the way testing is done is ok, though perhaps we can add some guards to make sure it really is only used in testing or loosen the comments a bit

rpc/client_interceptors.go
benjirewis: Yeah good point; we're not doing any active re-authing here. Switched to `UnaryClientInvalidAuthInterceptor`.

rpc/server_auth.go
benjirewis: Extended the comment on `WithEnsureAuthedHandler` to say "should only be used for testing", and threw an error below in the case that we're in a testing environment `if !testing.Testing()`.

rpc/client_interceptors.go
benjirewis: Gotchaa thanks for the context; I did something similar but used `status.Code`:
```go
if c := status.Code(err); c == codes.Unauthenticated || c == codes.PermissionDenied &&                                                                            
       creds != nil {                                                                                                                                                  
       creds.accessToken = ""                                                                                                                                          
}       
```

rpc/dialer_test.go
```diff
@@ -542,3 +542,78 @@ func TestWithStreamInterceptor(t *testing.T) {
 	)
 	test.That(t, interceptedCount, test.ShouldEqual, 1)
 }
+
+func TestReauth(t *testing.T) {
```
cheukt, line 546: nit: TestInvalidAuth

cheukt (approved):

rpc/server_auth.go
maximpertsov: [opt] we might want to consider either sticking this method into `testutils` or creating a separate test version of this package (something akin to https://pkg.go.dev/net/http/httptest). This prevents `testing` from being included in the final build of `rpc`.

That said, this might be annoying to do given that this method references a bunch of private methods, so I won't block on this.

benjirewis: Oof yeah you're right we'll include `testing` in the build of `rpc`. How would you both feel about only having the comment "should only be used for testing" and no actual error from `ensureAuthed` within a testing environment?

maximpertsov (approved):

PR #281, opened 2024-04-22
RSDK-7381: add more numato product IDs

mcvella:
This simply adds two more tested numato product IDs (8 channel GPIO and 1 channel relay with 4 channel GPIO)


biotinker (approved):
LGTM

PR #283, opened 2024-04-29
RSDK-7311 Fix MDB call queue `UpdateOne`

benjirewis:
[RSDK-7311](https://viam.atlassian.net/browse/RSDK-7311)

Adds an extra filter to the MDB call queue's `UpdateOne` that ensures the given `calls` document has not already been answered.

Signaling answerers run two goroutines (up to `defaultMaxAnswerers`) that both open bidi [`Answer`](https://github.com/viamrobotics/goutils/blob/main/proto/rpc/webrtc/v1/signaling.proto#L41) streams against the signaling server. The signaling server waits on `RecvOffer` from its call queue implementation (`wrtc_call_queue_mongodb.go` in prod) for both of these calls. It could be the case (and seemingly often was in my testing) that 

1. The first `RecvOffer` would run `FindOneAndUpdate` to mark the `calls` document as `answered` and send an `AnswerRequest_Init` back to one of the signaling answerer goroutines.
2. The second `RecvOffer` would run `FindOneAndUpdate`, find that the document had already been marked `answered`, receive an event from the change stream channel (`events`), run `UpdateOne` _without_ checking `answered`, and send an identical `AnswerRequest_Init` back to the other signaling answerer goroutine
3. Both signaling answerer goroutines would begin handling sending answerer candidate updates to the signaling server for the same connection establishment attempt

This resulted in
- Duplicate answerer candidates for a single connection establishment attempt (see RSDK-7311)
- Potentially "rogue" updates to incorrect connection establishment documents
- The usage of an answerer goroutine that could otherwise be used to handle a different connection establishment attempt
    - I believe this one resulted in successive connection establishment attempts causing longer connection establishment times

It's not going to be possible to write unit tests for this change. But, I tested locally with my own version of `app`: while the double `AnswerRequest_Init` is far less common with a local version of `app`, it still occasionally happens (~20/100 times). With the fix here, the issue seems to disappear (0/100).

cc @viamrobotics/netcode + @vijayvuyyuru 

[RSDK-7311]: https://viam.atlassian.net/browse/RSDK-7311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rpc/wrtc_call_queue_mongodb.go
+++ b/rpc/wrtc_call_queue_mongodb.go
@@ -1119,6 +1119,7 @@ func (queue *mongoDBWebRTCCallQueue) RecvOffer(ctx context.Context, hosts []stri
 						recvOfferCtx,
 						bson.D{
 							{webrtcCallIDField, callReq.ID},
+							{webrtcCallAnsweredField, false},
 						},
 						bson.D{
 							{"$set", bson.D{


edaniels (approved):
LGTM if in manual testing the repro goes away

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1119,6 +1119,7 @@ func (queue *mongoDBWebRTCCallQueue) RecvOffer(ctx context.Context, hosts []stri
 						recvOfferCtx,
 						bson.D{
 							{webrtcCallIDField, callReq.ID},
+							{webrtcCallAnsweredField, false},
```
dgottlieb, line 1122: I'd still like to know why we expect this second attempt to "do good". For example, if the first query failed because the "[caller error field](https://github.com/viamrobotics/goutils/pull/283/files#diff-c7273d63cbd81a307511aef73b2baa73baa9de7e566898c4ce7237f3ed192de0R1085)" had a value -- is that a case where this second chance will yield a better outcome?

dgottlieb (approved):

rpc/wrtc_call_queue_mongodb.go
edaniels: It's less a second chance and more so the case that the code is now on the "train" of collection events. Each one that comes in gives the code a shot to claim the call request. That is, it's a race among the other answerers for that host to claim the document/call.

PR #284, opened 2024-05-03
APP-3746: Redirect to login when no auth cookie present

michaellee1019:
When there is no cookie present, redirect back to login. The application should then request authentication flow again. In most cases the user will already be logged in and see nothing. This just add another redirect. In other cases the user will need to login again if the auth provider cannot validate the state and cookies.


ohEmily (approved):
Ah, thanks so much for getting to this!

PR #286, opened 2024-05-07
RSDK-7514: Add "on close" synchronization handle for PeerConnection renegotiation channels.

dgottlieb:


nicksanford:
@dgottlieb 
1. you have test failures
2. could you please confirm that this change solves the problem we would like it to solve in RDK before we merge this?

dgottlieb:
> you have test failures

Thankfully just as simple as compile errors

> could you please confirm that this change solves the problem we would like it to solve in RDK before we merge this?

Yep, I have a fix for that test. But it's a bit of a game to run the RDK CI without this in.

nicksanford (approved):
LGTM! I also tested this change in RDK with the repro steps you described here: https://viam.atlassian.net/browse/RSDK-7514.

PR #288, opened 2024-05-17
Statz: Add a counter wrapper with 5 labels

stevebriskin:


ohEmily (approved):

ohEmily:
surprised we can't use [variadic functions](https://gobyexample.com/variadic-functions) here and potentially a generic return type

PR #291, opened 2024-06-12
RSDK-7678 - Server streaming does not obey context cancels over WebRTC

cheukt:
actual code change is tiny - risk should be low as well since reset stream is only ever called when context is cancelled and in that case, the request should be cancelled anyway

### DIFF
--- a/rpc/wrtc_client_channel_test.go
+++ b/rpc/wrtc_client_channel_test.go
@@ -606,7 +606,6 @@ func TestClientStreamCancel(t *testing.T) {
 	// 2. After the initial request, the client closes its send side of the stream (replicating what happens in a server-side streaming scenario).
 	// 2. 3 messages are sent from the server. After the 3rd message, the client stream is cancelled.
 	// 3. The server should receive a RST_STREAM message and cancel the server context.
-
 	testutils.SkipUnlessInternet(t)
 	logger := golog.NewTestLogger(t)
 	pc1, pc2, dc1, dc2 := setupWebRTCPeers(t)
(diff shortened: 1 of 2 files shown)


rpc/wrtc_client_channel_test.go
```diff
@@ -599,3 +599,88 @@ func TestWebRTCClientChannelCanStopStreamRecvMsg(t *testing.T) {
...
+						}
+						stream.SendMsg(pongStatus.Proto())
+
+						// Using channels is not enough ensure that the client reset
+						// went through in time.
+						// Sleep here to let client process messages and send reset.
+						time.Sleep(10 * time.Millisecond)
+					}
+					// Failure as this means the context was never cancelled.
+					t.Fail()
```
maximpertsov, line 644: [minor]

```suggestion
					t.FailNow()
```

cheukt: FailNow does not work if called in a goroutine (which this is)

maximpertsov: ah, good to know

rpc/wrtc_client_channel_test.go
```diff
@@ -599,3 +599,89 @@ func TestWebRTCClientChannelCanStopStreamRecvMsg(t *testing.T) {
...
+
+	err = clientStream.CloseSend()
+	test.That(t, err, test.ShouldBeNil)
+
+	for i := 0; i < 3; i++ {
+		var respStatus pbstatus.Status
+		err = clientStream.RecvMsg(&respStatus)
+		test.That(t, err, test.ShouldBeNil)
+	}
+	cancelStream()
```
maximpertsov, line 685: do we also need to `defer` cancel this context earlier in case a prior assertion fails?

maximpertsov (approved):

rpc/wrtc_client_channel_test.go
cheukt: should be ok since we defer close the server channel and client channels separately

dgottlieb (approved):

PR #293, opened 2024-06-17
RSDK-7280 Report more ICE timings

benjirewis:
RSDK-7280

Logs:
- When/which local ICE candidates are gathered
- When/which remote ICE candidates are received
- How long from start of `dialWebRTC` to ICE connection state of connected
     - Number of caller updates in that time
     - Average duration of caller updates in that time
     - Max duration of caller updates in that time

### DIFF
--- a/rpc/wrtc_base_channel.go
+++ b/rpc/wrtc_base_channel.go
@@ -38,6 +38,7 @@ func newBaseChannel(
 	peerConn *webrtc.PeerConnection,
 	dataChannel *webrtc.DataChannel,
 	onPeerDone func(),
+	onICEConnected func(),
 	logger golog.Logger,
 ) *webrtcBaseChannel {
 	ctx, cancel := context.WithCancel(ctx)
@@ -108,8 +109,13 @@ func newBaseChannel(
 					"conn_state", connectionState.String(),
 				)
 				doPeerDone()
+			case webrtc.ICEConnectionStateConnected:
+				if onICEConnected != nil {
+					onICEConnected()
+				}
+				fallthrough
 			case webrtc.ICEConnectionStateChecking, webrtc.ICEConnectionStateCompleted,
-				webrtc.ICEConnectionStateConnected, webrtc.ICEConnectionStateNew:
+				webrtc.ICEConnectionStateNew:
 				fallthrough
 			default:
 				candPair, hasCandPair := webrtcPeerConnCandPair(peerConn)
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -162,9 +162,8 @@ func dialWebRTC(
 	)
 	onICEConnected := func() {
 		// Delay by up to 5s to allow more caller updates/better stats.
-		timer := time.NewTimer(5 * time.Second)
 		select {
-		case <-timer.C:
+		case <-time.After(5 * time.Second):
 		case <-ctx.Done():
 		}
 
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/edaniels/golog"
 	"github.com/pion/webrtc/v3"
@@ -102,6 +103,8 @@ func dialWebRTC(
 	dOpts dialOptions,
 	logger golog.Logger,
 ) (ch *webrtcClientChannel, err error) {
+	dialStart := time.Now()
+
 	logger = logger.Named("webrtc")
 	dialCtx, timeoutCancel := context.WithTimeout(ctx, getDefaultOfferDeadline())
 	defer timeoutCancel()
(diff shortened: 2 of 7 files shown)


rpc/wrtc_client.go
```diff
@@ -152,6 +155,15 @@ func dialWebRTC(
 		}
 	}()
 
+	var callUpdates int
+	var maxCallUpdateDuration, totalCallUpdateDuration time.Duration
+	onICEConnected := func() {
+		averageCallUpdateDuration := totalCallUpdateDuration / time.Duration(callUpdates)
+		// TODO: Potentially report these stats to sentry/some central location at some point.
+		logger.Debugw("ICE connected", "time_since_dial_start", time.Since(dialStart), "num_call_updates",
+			callUpdates, "average_duration", averageCallUpdateDuration, "max_call_update_duration", maxCallUpdateDuration)
```
benjirewis, line 164: Our wrtc wrappers + Pion never seem to reach a state of ICE ["completed"](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/iceConnectionState#completed), only "connected". In the flutter SDK, we only report caller updates after the ICE process has been completed, which lets us gather stats on something like 10 caller updates. Here in go, we usually only see 2 or 3 until reaching connected. Still probably worth logging, though.

rpc/wrtc_base_channel.go
```diff
@@ -108,8 +109,13 @@ func newBaseChannel(
 					"conn_state", connectionState.String(),
 				)
 				doPeerDone()
+			case webrtc.ICEConnectionStateConnected:
+				if onICEConnected != nil {
+					onICEConnected()
```
benjirewis, line 114: I had to do some funky `onICEConnected` callback logic here, as defining an `OnICEConnectionStateChange` handler after `newBaseChannel` ran would've overwritten the existing one (see the `connStateChanged` anonymous function this line is in). Only one specified handler is allowed.

maximpertsov: it took me a second to understand where the overwrite was happening: https://github.com/viamrobotics/goutils/blob/377d0dcf55f18e19ffb46d53363b5a02a8e6aa71/rpc/wrtc_base_channel.go#L134

makes sense now

rpc/wrtc_client.go
maximpertsov: do we never reach "connected" because we use trickle ice?

rpc/wrtc_client.go
```diff
@@ -152,6 +155,18 @@ func dialWebRTC(
 		}
 	}()
 
+	var statsMu sync.Mutex
+	var callUpdates int
+	var maxCallUpdateDuration, totalCallUpdateDuration time.Duration
```
maximpertsov, line 160: minor

```suggestion
	var (
		statsMu sync.Mutex
		callUpdates int
		maxCallUpdateDuration, totalCallUpdateDuration time.Duration
	)
```

maximpertsov (approved):
cha

rpc/wrtc_client.go
dgottlieb: Why isn't the former preferable? This looks like more lines without really affecting the readability.

maximpertsov: For me it helps to highlight the variables as a group since they are all related to stats, but I definitely don't feel strongly about this. Totally cool leaving as is.

rpc/wrtc_client.go
benjirewis: Discussed a bit offline; it's probably not because of trickle ICE. The flutter SDK/`flutter-webrtc` is able to declare a state of ICE completed when trickle ICE is being used. It's probably because our golang WebRTC clients never declare the ICE exchange as "complete" (not sure what this looks like in code). Per @dgottlieb's suggestion, I'm going to add a delay of 5s after we reach ICE connected before reporting these stats in the hope that the stats will include information from more caller updates.

rpc/wrtc_client.go
benjirewis: Grouped!

rpc/wrtc_client.go
benjirewis: I can't just `time.Sleep(5 * time.Second)`, as I think tests would try to log after completion if I did. `select`ed on a 5s timer and the `dialWebRTC` context instead; might be the best we can do unless others have ideas.

rpc/wrtc_client.go
```diff
@@ -152,6 +155,27 @@ func dialWebRTC(
...
+	var (
+		statsMu                                        sync.Mutex
+		callUpdates                                    int
+		maxCallUpdateDuration, totalCallUpdateDuration time.Duration
+	)
+	onICEConnected := func() {
+		// Delay by up to 5s to allow more caller updates/better stats.
+		timer := time.NewTimer(5 * time.Second)
+		select {
+		case <-timer.C:
```
maximpertsov, line 167: q: we only need to delay once right? 

```suggestion
		select {
		case <-time.After(5 * time.Second):
```

PR #298, opened 2024-06-26
RSDK-6659: wait on running goroutines before concluding test

puranamp:
Description:
We had observed a datarace in the test `TestWebRTCAnswererImmediateStop` that resulted in an attempt to write logs after completion of the test. To fix this, we wait on the running goroutines to complete before closing the test.

### DIFF
--- a/rpc/wrtc_signaling_test.go
+++ b/rpc/wrtc_signaling_test.go
@@ -190,8 +190,6 @@ func TestWebRTCAnswererImmediateStop(t *testing.T) {
 	var wg sync.WaitGroup
 	wg.Add(2)
 
-	ch := make(chan struct{})
-
 	go func() {
 		defer wg.Done()
 		answerer.Start()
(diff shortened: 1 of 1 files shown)


maximpertsov (approved):

rpc/wrtc_signaling_test.go
```diff
@@ -186,6 +187,16 @@ func TestWebRTCAnswererImmediateStop(t *testing.T) {
...
+
+	go func() {
+		defer wg.Done()
+		answerer.Start()
+	}()
+	go func() {
+		defer wg.Done()
+		answerer.Stop()
+	}()
+	wg.Wait()
```
dgottlieb, line 201: @maximpertsov, I felt the original timeout of 30 seconds was well intentioned, but then we lose the stack traces of what's happening if the test were to fail. And without those stacks we'd never understand why* `Wait` didn't return. I don't think there's any real value in trying to protect CI from spinning for 10 minutes before failing, so I urged Pranav to strip that bit out.

maximpertsov: fair point, that works for me

maximpertsov (approved):

dgottlieb (approved):

PR #299, opened 2024-06-28
Avoid division-by-zero in stats reporting

benjirewis:
Do not report stats if no call updates were captured (avoid divide-by-zero).

### DIFF
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -169,6 +169,9 @@ func dialWebRTC(
 
 		statsMu.Lock()
 		defer statsMu.Unlock()
+		if callUpdates == 0 {
+			return
+		}
 		averageCallUpdateDuration := totalCallUpdateDuration / time.Duration(callUpdates)
 		// TODO: Potentially report these stats to sentry/some central location at some point.
 		logger.Debugw("ICE connected", "time_since_dial_start", time.Since(dialStart), "num_call_updates",


rpc/wrtc_client.go
```diff
@@ -169,6 +169,9 @@ func dialWebRTC(
 
 		statsMu.Lock()
 		defer statsMu.Unlock()
+		if callUpdates == 0 {
```
dgottlieb, line 172: I suppose we're surprised that `onICEConnected` can be called if we never incremented `callUpdates`?

I'm just looking at the debug log line and wondering if we would still like to output that. This patch would (intentionally) be suppressing the log line.

I know I saw this panic get printed when running the rdk/cli tests. Logs are blown away now (probably will get them again soon), but not sure exactly which test I saw it in if you wanted to repro.

benjirewis: > I suppose we're surprised that onICEConnected can be called if we never incremented callUpdates?

Yeah I've also found that to be the case in typescript (and now you've seen it in Go). In typescript, I found that we received one remote ICE candidate, completed no `CallUpdateRequest`s, and almost immediately declared ICE [connected](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/iceConnectionState#connected). Since Go/TS are the controlling agents from what I can tell, I guess that means that the single remote ICE candidate was "valid" from the perspective of the caller (probably due to immediate, successful connectivity checks) and a usable connection was created.

Also, we have [another log](https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_base_channel.go#L126-L130) that will print when we've switched to ICE connected; I think this one would just be for stats.

dgottlieb (approved):

PR #301, opened 2024-06-28
RSDK-7935 - Fix header datarace

cheukt:
if we call SetHeader and SendHeader at the same time, there's a chance at a datarace. Adding the RLock removes that possibility. The grpc-go impl has specific [locks](https://github.com/grpc/grpc-go/blob/f199062ef31ddda54152e1ca5e3d15fb63903dc3/internal/transport/transport.go#L469) for the header, but not sure if it's necessary. The main call into writeHeader already holds the RLock anyway (https://github.com/viamrobotics/goutils/blob/59474fccb894a0593029792c977ffb9bd173a4aa/rpc/wrtc_server_stream.go#L169).

Able to reproduce with a test that ran SetHeader and SendHeader side-by-side, and confirms that the race went away with this fix. Did not add that test into this PR because it's flaky

### DIFF
--- a/rpc/wrtc_server_stream.go
+++ b/rpc/wrtc_server_stream.go
@@ -392,6 +392,7 @@ func (s *webrtcServerStream) closeWithSendError(err error) (writeErr error) {
 }
 
 func (s *webrtcServerStream) writeHeaders() error {
+	// Grab the RLock to prevent headers from being set while written.
 	s.mu.RLock()
 	if !s.headersWritten.CompareAndSwap(false, true) {
 		s.mu.RUnlock()
--- a/rpc/wrtc_server_stream.go
+++ b/rpc/wrtc_server_stream.go
@@ -392,12 +392,11 @@ func (s *webrtcServerStream) closeWithSendError(err error) (writeErr error) {
 }
 
 func (s *webrtcServerStream) writeHeaders() error {
-	// Grab the RLock to prevent headers from being set while written.
-	s.mu.RLock()
 	if !s.headersWritten.CompareAndSwap(false, true) {
-		s.mu.RUnlock()
 		return nil
 	}
+	// Grab the RLock to prevent headers from being set while written.
+	s.mu.RLock()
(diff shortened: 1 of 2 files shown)


dgottlieb:
Even with a lock to remove the data race, it seems incorrect for two goroutines to be accessing the same message object. Did you uncover the circumstances in which the two are racing?

cheukt:
We are indeed accessing the same message object, but not sure about the best way to mitigate.

How the race happens - the server got a reset stream and heartbeat at around the same time. Possible that the heartbeat was sent from the client and then the stream got closed (client closed or program ended). The race is that the routine handling the reset stream is trying to read headers to be written to the client, and the routine that is handling the heartbeat wants to add the op id to the header. We’ve seen cases where the header does get sent and then the write comes in, and we’d then get the unchecked error being logged saying we’re trying to write headers after they’ve already been sent. Both error cases happen because we’re handling two requests on the same stream at around the same time on different go routines.

grpc-go I believe has a similar issue, https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_server.go#L829 is where the reset stream is handled and based on user reports, we see similar issues in https://viam.atlassian.net/browse/RSDK-6985. Ultimately, the issue should only appear when the stream is about to be closed anyway, so not as important of an issue. Separately, I do think we can filter those errors out from being logged if it can be reliably reproduced that they only show up when a stream/request is getting closed down.

rpc/wrtc_server_stream.go
```diff
@@ -392,10 +392,14 @@ func (s *webrtcServerStream) closeWithSendError(err error) (writeErr error) {
 }
 
 func (s *webrtcServerStream) writeHeaders() error {
+	// Grab the RLock to prevent headers from being set while written.
+	s.mu.RLock()
```
dgottlieb, line 396: Let's save ourselves the `RUnlock` on 398 and instead just grab the lock prior to `metadataToProto(s.header)`.

dgottlieb (approved):

PR #302, opened 2024-07-01
RSDK-7637 - rpc: Set OnNegotiationNeeded before any PeerConnection sub-resources are created

edaniels:


rpc/wrtc_peer.go
```diff
@@ -236,6 +233,16 @@ func newPeerConnectionForServer(
 	return peerConn, dataChannel, nil
 }
 
+// PeerRole identifies which role of a Client/Server relationship a peer is assuming.
+type PeerRole bool
```
edaniels, line 237: I thought this was a helpful addition to `ConfigureForRenegotiation` since we differentiate the behavior, but happy to remove it if it's too much fluff.

rpc/wrtc_peer.go
```diff
@@ -244,12 +251,84 @@ func newPeerConnectionForServer(
...
+
+	// OnNegotiationNeeded is webrtc callback for when a PeerConnection is mutated in a way such
+	// that its local description should change. Such as when a video track is added that should be
+	// streamed to the peer.
+	//
+	// Dan: The existing `OnNegotiationNeeded` algorithm is suitable when one side initiates all of
+	// the renegotiations. But it is not obvious that algorithm is suitable for when both sides can
+	// race on renegotiating. For now we "uninstall" the `OnNegotiationNeeded` callback and only
+	// allow the "server" to start a renegotiation.
+	if role == PeerRoleServer {
```
dgottlieb, line 287: I'm not observing where we set up an `OnNegotiationNeeded` handler for clients. I was expecting to still see the empty one. My understanding was that it was necessary to install a handler to ensure the renegotiation loop ends up in the right state. Was that intentional?

rpc/wrtc_peer.go
```diff
@@ -105,37 +115,20 @@ func newPeerConnectionForClient(
 		Ordered:    &ordered,
 	})
 	if err != nil {
-		return peerConn, nil, err
+		return nil, nil, err
```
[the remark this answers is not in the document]
edaniels: So there's this at line 92 where will close everything down if not successful in setting up the client peer which will ultimately close everything down (data channels too with the other PR)

```golang
	var successful bool
	defer func() {
		if !successful {
			err = multierr.Combine(err, peerConn.Close())
		}
	}()
```

rpc/wrtc_peer.go
edaniels: The empty one doesn't do anything so yeah, it's not needed. All setting it does is ultimately have it be called if negotiation is needed. I think we'd have seen a test fail here if it were needed.

rpc/wrtc_peer.go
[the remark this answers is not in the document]
dgottlieb: Ah, yes that, thanks for the reminder.

rpc/wrtc_peer.go
dgottlieb: Oh, is it not needed anymore because of [this](https://github.com/pion/webrtc/commit/b3856ffcfcd0eb8ae610bc72d407433ed86b39e4) patch? Before we had this [defer block](https://github.com/pion/webrtc/blob/952fbe068c0fc24dfde2c0788899827747ff12f0/peerconnection.go#L326-L334) and its placement relative to the "nil on-negotiation needed" check was important for preventing the problem.

edaniels: Even that I don't think matters. The client is never the one cause negotiation needed, so that part of the state machine ends up not mattering. The state for negotiation, like being stable, happens in set description.

edaniels: But if ever in the future the client would say add a track or a data channel on demand from its side, it would need the handler, otherwise it would miss negotiation needed events and not ever be stable.

dgottlieb (approved):

PR #303, opened 2024-07-02
RSDK-7925 - update webrtc to v3.2.44 for blocking pc.Close

edaniels:
Pulls in:
https://github.com/pion/ice/pull/706
https://github.com/pion/webrtc/pull/2798


rpc/wrtc_base_channel.go
```diff
@@ -75,15 +75,7 @@ func newBaseChannel(
 		}
 	}
 	connStateChanged := func(connectionState webrtc.ICEConnectionState) {
-		ch.mu.Lock()
```
edaniels, line 78: To be honest, I'm not sure why I ever wrote this part of the code in. It's deleted because now with an ICE connection closing down due to pc.Close, it may fire off its connection state changing which in turn calls this, causing a deadlock. I don't see why we should keep it!

dgottlieb: I suspect before with `peerConn.Close` not blocking, we could get this race/leak:
```
| connStateChanged                                | webrtcBaseChannel.Close                             |
|-------------------------------------------------+-----------------------------------------------------|
|                                                 | closeWithReason                                     |
|                                                 | - Lock                                              |
|                                                 | - closed = true                                     |
|                                                 | - PeerConn.Close                                    |
|                                                 | - Unlock                                            |
| activeBackgroundWorkers.Add(1) (0 workers -> 1) | activeBackgroundWorkers.Wait (0 workers -> returns) |
| Spawn Goroutine                                 |                                                     |
```

And now I suspect with close blocking -- it's guaranteed that either:
* We'll spin up the onStateChange callback + goroutine before* Close gets called or
* Close returning guarantees `connStateChanged` can no longer be invoked.

dgottlieb (approved):

rpc/wrtc_base_channel.go
edaniels: Yeah I think you're spot on

PR #304, opened 2024-07-08
RSDK-7936 - grpc: set keepalive to handle network interface changes

edaniels:
So on macOS, changing network interfaces does not appear to be an issue, but it is on Linux. As James pointed out, when you switch, it can take a very long time to have the gRPC connection start using the new network interface. This issue is further compounded by the cached gRPC dialer that rdk uses for app.viam.com based connections. Adding a keep alive solves the issue (by default it's not used) since once the keep alive is sent and enough are not ACKd, the http2 client transport is closed, the internal client load balancer sees this, and makes a new connection using a suitable interface that has a valid gateway.

### DIFF
--- a/rpc/dialer.go
+++ b/rpc/dialer.go
@@ -8,6 +8,7 @@ import (
 	"hash/fnv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/edaniels/golog"
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
@@ -21,6 +22,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
@@ -241,6 +243,10 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts dialOptions, logg
 	dialOpts := []grpc.DialOption{
 		grpc.WithBlock(),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(MaxMessageSize)),
+		grpc.WithKeepaliveParams(keepalive.ClientParameters{
+			Time:                15 * time.Second, // a little extra buffer to try to avoid ENHANCE_YOUR_CALM
+			PermitWithoutStream: true,
+		}),
 	}
 	if dOpts.insecure {
 		dialOpts = append(dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))


edaniels:
cc @ale7714 as you're the POC for this work

edaniels:
@abe-winter, when switching interfaces, the WebRTC signaling code gets "too many connection attempts; please wait a bit and try again". I think this is related to load balancing in GCP where the load balancer still sees an active `answer` stream, even though it's not longer in use. This has probably been an existing an issue but this is making it appear. I don't think I can do anything from my end to alleviate that, so be advised.

ale7714 (commented):
This makes sense and lgtm. I'll let @Otterverse test the change. FYI He's OOO today.

rpc/dialer.go
```diff
@@ -241,6 +243,10 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts dialOptions, logg
 	dialOpts := []grpc.DialOption{
 		grpc.WithBlock(),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(MaxMessageSize)),
+		grpc.WithKeepaliveParams(keepalive.ClientParameters{
+			Time:                15 * time.Second, // a little extra buffer to try to avoid ENHANCE_YOUR_CALM
+			PermitWithoutStream: true,
```
abe-winter, line 248: I think this means the device will send a packet every 15 seconds

These are HTTP2 pings per [grpc-go readme](https://github.com/grpc/grpc-go/blob/bb49a8868a2a48de9321a5ad6f0ab4ccfd805f77/Documentation/keepalive.md?plain=1#L3), so this shouldn't create extra log noise

the ClientParameters struct has Time and Timeout, so I think your config means the actual timeout is 35 seconds (15 + 20)?

also, I'm assuming RDK is used to having a transport that never goes away. now that this can close, does RDK reconnect cleanly with backoff? (i.e. not a try-fail flood).

abe-winter (commented):
see comment about reconnect backoff, and also IOU load balancer research, but otherwise looks good

rpc/dialer.go
edaniels: 35 seconds is correct and yeah it reconnects as you'd expect it to. There used to be keep alive but we removed it

edaniels: The RDK won't/can't see a disconnect in this way; it never has. The underlying load balancer does the connection logic

abe-winter: > underlying load balancer

ah you're talking about grpc client-side load balancing like in [grpc docs here](https://grpc.io/blog/grpc-load-balancing/#proxy-or-client-side), my brain kept hearing 'load balancer' and thinking server

reconnect logic lives in the grpc library, makes sense https://github.com/grpc/grpc-go/blob/bb49a8868a2a48de9321a5ad6f0ab4ccfd805f77/clientconn.go#L584-L587

abe-winter: commit removing old one https://github.com/viamrobotics/goutils/commit/ce35a14fc60f8a303d7479e61c6051dca2854688

edaniels: Yeah I don't buy that Google GCP load balancers for grpc are sending any keep alive params that are helpful for us. Otherwise we'd see faster connection reconnects

edaniels: So with 35 seconds it feels harmless. Happy to adjust it though

abe-winter: oh yeah no opinion on what the default should be, just wrapping my head around this part of the codebase

abe-winter: > sending any keep alive params that are helpful for us

^ what do you mean by this? does the cloud side of these connections send down config when a connection is established which the grpc client then uses to configure tcp keepalive?

edaniels: > > sending any keep alive params that are helpful for us
> 
> ^ what do you mean by this? does the cloud side of these connections normally send down config when a connection is established which the grpc client then uses to configure tcp keepalive?

I misspoke, my bad. Both sides are just supposed to agree on the values, but GCP doesn't publish what those are. I believe I removed the keep alives because we were seeing a lot of GOAWAYS, which the grpc library handles by backing off future keep alives.

abe-winter: also can you dump repro instructions somewhere for 'too many attempts' error when switching interfaces? I'll diff how this behaves with LB vs directly to the backend service, from there can experiment with LB config changes

edaniels: Yeah, just run a blank config from app.viam.com and switch between net ifaces being up and down that have internet access. In my test, I used nmcli to switch wifi networks.

abe-winter:
awesome thanks

Otterverse (approved):
LGTM!

I built custom agent and viam-server binaries against this branch and verified it drops invalid connections within 20 seconds or so once the connection goes away, sometimes before I can even connect the new network. New connections are established within seconds (e.g. the next retry) using the new network/IP as well, as soon as it comes up.

PR #306, opened 2024-07-09
RSDK-8139 - rpc: only take RLock after writeHeaders in webrtcServerStream.SendMsg

edaniels:
I noticed this failure in https://github.com/viamrobotics/goutils/actions/runs/9860875871/job/27228052504?pr=305. There's a chance that headers are not written by the time SendMsg is called which results in a RLock being called twice.


dgottlieb:
Can you slap this ticket name onto the commit message? https://viam.atlassian.net/browse/RSDK-8139

dgottlieb (approved):

PR #308, opened 2024-07-16
RSDK-8205: Add New RDK Compatible Logging Interface

danielbotros:
### Description
RSDK-8205: Allow RDK loggers to be used in goutils
* Adds a new interface so that both RDK loggers and Zap loggers can be used in goutils without the need to convert before passing into goutils
* Does so by defining a new `ZapCompatibleLogger` interface and replaces `golog.Logger` type requirements with this new `ZapCompatibleLogger`
* `zeroconf` functions only accept `golog.logger` / `*zap.SugaredLogger`, so to get around this loggers are `Desugar()`ed and then `Sugar()`ed to force the zap conversion. 
* RDK side of changes (WIP) found [here](https://github.com/viamrobotics/rdk/pull/4207)


rpc/dial.go
```diff
@@ -294,9 +294,9 @@ func dial(
 	return conn, cached, nil
 }
 
-func lookupMDNSCandidate(ctx context.Context, address string, logger golog.Logger) (*zeroconf.ServiceEntry, error) {
+func lookupMDNSCandidate(ctx context.Context, address string, logger utils.ZapCompatibleLogger) (*zeroconf.ServiceEntry, error) {
 	candidates := []string{address, strings.ReplaceAll(address, ".", "-")}
-	resolver, err := zeroconf.NewResolver(logger, zeroconf.SelectIPRecordType(zeroconf.IPv4))
+	resolver, err := zeroconf.NewResolver(logger.(*zap.SugaredLogger), zeroconf.SelectIPRecordType(zeroconf.IPv4))
```
danielbotros, line 299: this is just a raw type conversion for now since we aren't passing in RDK loggers to goutils yet, but it seems like we will end up needing some kind of zap conversion to be able to pass loggers into `zeroconf` functions

dgottlieb:
Being paranoid: Have you checked that [app](https://github.com/viamrobotics/app) with pointed at this change continues to compile?

danielbotros:
> Being paranoid: Have you checked that [app](https://github.com/viamrobotics/app) pointed at these change continues to compile?

Ran app + test suite. Everything seems to work as intended and test suite passes 👍, could not find any regression in app functionality

rpc/dial.go
```diff
@@ -294,9 +295,9 @@ func dial(
 	return conn, cached, nil
 }
 
-func lookupMDNSCandidate(ctx context.Context, address string, logger golog.Logger) (*zeroconf.ServiceEntry, error) {
+func lookupMDNSCandidate(ctx context.Context, address string, logger utils.ZapCompatibleLogger) (*zeroconf.ServiceEntry, error) {
 	candidates := []string{address, strings.ReplaceAll(address, ".", "-")}
-	resolver, err := zeroconf.NewResolver(logger, zeroconf.SelectIPRecordType(zeroconf.IPv4))
+	resolver, err := zeroconf.NewResolver(logger.Desugar().Sugar(), zeroconf.SelectIPRecordType(zeroconf.IPv4))
```
benjirewis, line 300: Calling out this `Desugar` + `Sugar` pattern here for discussion. Sounds like we need this to go from an RDK `impl` logger to zap specifically when passing to `zeroconf` API.

The zeroconf library is our fork (well, Eric's, strictly speaking), and we actually added these `logger` arguments at some point as the main fork of `zeroconf` used the standard library's log method and printed a bunch of stuff uncontrollably to STDERR.

There are two options in my head. We can:
- `Desugar` and `Sugar` when passing to zeroconf
- Refactor `zeroconf` to accept either a copy of `utils.ZapCompatibleLogger` or directly reference it (the latter would create an annoying circular dependency)

I think the first option (what you have) is fine, but happy to hear arguments to the contrary.

runtime.go
```diff
@@ -18,7 +18,7 @@ import (
 // ContextualMain calls a main entry point function with a cancellable
 // context via SIGTERM. This should be called once per process so as
 // to not clobber the signals from Notify.
-func ContextualMain[L ILogger](main func(ctx context.Context, args []string, logger L) error, logger L) {
+func ContextualMain[L ZapCompatibleLogger](main func(ctx context.Context, args []string, logger L) error, logger L) {
```
benjirewis, line 21: I _think_ this is a somewhat breaking change, so if we do this, we may want to announce it as such. You're modifying the generic constraint here to be more intense (require more methods than `ILogger`). Authors of golang modules, in particular, often use `ContextualMain` to wrap their main function and get automatic signal handling. If they were passing in their own bare-bones logger type, their code may no longer compile.

benjirewis (commented):
Nice! Some comments for discussion.

rpc/dial.go
dgottlieb: I think this is super clever and am happy to go with it. If we had a bunch more cases of using desugar->sugar, we'd maybe need to step back and take a better look. But given we only needed to overcome a small number of calls into zeroconf, this is ideal IMO.

I do think this would be surprising for a random code reader if they saw this. I'd add a comment -- maybe referencing the ticket for additional context. Just so they can understand that the motivation is to massage the type system and nothing more.

runtime.go
dgottlieb: Good catch @benjirewis. I had no intention of reading all of the substitutions.

I'd agree with going back to `ILogger` for these contextual main cases (but keeping the changes for "slow goroutine" methods in this file at ZapCompatibleLogger). We don't need the contextual main loggers to be hooked into the registry system. All this stuff actually gets kicked off before we've created a registry.

rpc/dial.go
maximpertsov: I agree with @dgottlieb - I'm fine with the current approach if we have a comment explaining why this is necessary.

runtime.go
danielbotros: Good call out, seems like it does not need to be a `ZapCompatibleLogger` and wasn't not sure if this was even necessary. Victim of the search and replace lol. Will revert!

benjirewis (approved):
Nice! LGTM assuming RDK and app are able to build locally for you with these changes.

maximpertsov (approved):

dgottlieb (approved):

puranamp (approved):

PR #309, opened 2024-07-16
RSDK-8215 - Decrease amount of GOAWAY errors

cheukt:
Adding the server side params eliminates the GOAWAY errors coming from the internal signaling server (tested by running a local config in RDK for 20 mins), TBD as to whether it'll stop errors coming from cloud connections (so far, I've seen reduced rates, but usually some pop up after 8 mins). There's a possibility that after pushing this change to app, this log will go away as well, but possible that app load balancer may continue to issue these errors.

We could also look into filtering that specific log out, since there's no real issue as far as we can tell

### DIFF
--- a/rpc/const.go
+++ b/rpc/const.go
@@ -1,4 +1,11 @@
 package rpc
 
-// MaxMessageSize is the maximum size a gRPC message can be.
-var MaxMessageSize = 1 << 25
+import "time"
+
+var (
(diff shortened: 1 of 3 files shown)


abe-winter (approved):

abe-winter:
in theory could add a DialOption to turn it off, but I don't feel strongly; not a bad default

PR #311, opened 2024-07-22
RSDK-8320: Allow RDK Loggers to Call Sublogger()

danielbotros:
### Description
RSDK-8320: Allow RDK loggers to call sublogger when creating new loggers in goutils
* We are unable to maintain any pointers to loggers created by `Named()`. Now that RDK loggers can be passed into goutils without any conversion, we can create and register new loggers in `goutils` by calling `Sublogger()`.
* `goutils` still needs to be compatible with the Zap logging interface because of `app`. In order to allow RDK loggers to call `Sublogger()` a `utils.Sublogger()` function is added in this PR which uses reflection to either call `Sublogger()` if it is an RDK logger, or `Named()` if it is a Zap logger. 
* This way, when RDK loggers are passed into `goutils`, `Sublogger()` is called registering these new loggers. 
* Replaces all instances of `Named()` in `goutils` with `utils.Sublogger()`
* We still don't have the ability to control a few log lines owned by loggers called with `With()`, but it seems like a minimal surface. 
### Testing
* Ran app + RDK and respective test suite against these changes to ensure nothing breaks.
* Print logger registry to confirm these loggers are being registered.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -51,7 +51,16 @@ type ZapCompatibleLogger interface {
 // This function uses reflection to dynamically create a sublogger from the provided logger by
 // calling its `Sublogger` method if it is an RDK logger, or its `Named` method if it is a Zap logger.
 // If neither method is available, it logs a debug message and returns the original logger.
-func Sublogger(inp ZapCompatibleLogger, subname string) ZapCompatibleLogger {
+func Sublogger(inp ZapCompatibleLogger, subname string) (loggerRet ZapCompatibleLogger) {
+	loggerRet = inp
+
+	defer func() {
+		if r := recover(); r != nil {
+			inp.Debugf("panic occurred while creating sublogger: %v, returning self", r)
+		}
+
+	}()
+
 	typ := reflect.TypeOf(inp)
 	sublogger, ok := typ.MethodByName("Sublogger")
 	if !ok {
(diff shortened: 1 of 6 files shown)


logger.go
```diff
@@ -45,3 +46,28 @@ type ZapCompatibleLogger interface {
...
+	sublogger, ok := typ.MethodByName("Sublogger")
+	if !ok {
+		sublogger, ok = typ.MethodByName("Named")
+		if !ok {
+			inp.Debugf("could not create sublogger from logger of type %s, returning self", typ.String())
+			return inp
+		}
+	}
+
+	ret := sublogger.Func.Call([]reflect.Value{reflect.ValueOf(inp), reflect.ValueOf(subname)})
```
dgottlieb, line 65: This panics if the function inputs aren't what we expect. I'm not sure how defensive we want to be here.

If we want to just `defer recover` and `log + return inp`, that's good enough. I don't realistically expect this error case to happen.

If we want to try to catch this before* calling the `sublogger.Func`, I think this starts with the `subloggerMethod.Type.NumIn()` call. Presumably we expect that to be `2` (the receiver argument + the sublogger name). And `subloggerMethod.Type.In(idx)` to check the types of the inputs.

benjirewis: Good catch; I'm also fine with `defer` + `recover`.

benjirewis (commented):
Can we add a test for the new `Sublogger` function in a `logger_test.go`? Could assert that both a `golog.Logger` and some `ZapCompatibleLogger`-fulfilling-struct-that-also-implements-`Sublogger` cause no error?

logger.go
danielbotros: `defer` + `recover` also good with me and safer

benjirewis (approved):
Looks good! Thanks 🧑‍🔧 .

dgottlieb (approved):

puranamp (approved):

maximpertsov (approved):

PR #313, opened 2024-07-23
RSDK-8288: Make caller and answerer less sensitive to errors from signaling server

danielbotros:
### Description
RSDK-8288: Make caller and answerer less sensitive to errors from signaling server
* Signaling caller and answerer are made less sensitive to signaling server errors by swallowing either `io.EOF` errors or grpc errors of the type `code = Internal desc = unexpected EOF` and logging it. 
* This is necessary since after one of the answerer/caller calls `sendDone()` and severs its connection to the signaling server, it is possible that the other has not had the chance to "finish" and timing can cause it to get a EOF error from its connection to the signaling server, causing the peer connection to be unecessarily closed.

### DIFF
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -362,7 +362,7 @@ func dialWebRTC(
 	utils.PanicCapturingGoWithCallback(func() {
 		if err := exchangeCandidates(); err != nil {
 			if haveInit && filterEOF(err, logger) == nil {
-				logger.Warnf("caller swallowed err: %v while exchanging ICE candidates", err)
+				logger.Warn("caller swallowed EOF err while exchanging ICE candidates")
 			} else {
 				sendErr(err)
 			}
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -181,7 +181,7 @@ func dialWebRTC(
 	exchangeCtx, exchangeCancel := context.WithCancel(signalCtx)
 	defer exchangeCancel()
 
-	// atomic bool representing whether initial sdp exchange has occured
+	// atomic bool representing whether initial sdp exchange has occurred
 	var haveInit atomic.Bool
 	haveInit.Store(false)
 
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -182,12 +182,12 @@ func dialWebRTC(
 	defer exchangeCancel()
 
 	// atomic "bool" representing whether initial sdp exchange has occured
-	var haveInit int32
-	atomic.StoreInt32(&haveInit, 0)
+	var haveInit atomic.Bool
+	haveInit.Store(false)
 
 	errCh := make(chan error)
 	sendErr := func(err error) {
-		if atomic.LoadInt32(&haveInit) == 1 {
+		if haveInit.Load() {
 			err = filterEOF(err, logger)
 		}
 		if s, ok := status.FromError(err); ok && strings.Contains(s.Message(), noActiveOfferStr) {
@@ -328,10 +328,10 @@ func dialWebRTC(
 			}
 			switch s := callResp.Stage.(type) {
 			case *webrtcpb.CallResponse_Init:
-				if atomic.LoadInt32(&haveInit) == 1 {
+				if haveInit.Load() {
 					return errors.New("got init stage more than once")
 				}
-				atomic.StoreInt32(&haveInit, 1)
+				haveInit.Store(true)
 				uuid = callResp.Uuid
 				answer := webrtc.SessionDescription{}
 				if err := DecodeSDP(s.Init.Sdp, &answer); err != nil {
@@ -348,7 +348,7 @@ func dialWebRTC(
 					return sendDone()
 				}
 			case *webrtcpb.CallResponse_Update:
-				if atomic.LoadInt32(&haveInit) == 0 {
+				if !haveInit.Load() {
 					return errors.New("got update stage before init stage")
 				}
 				if callResp.Uuid != uuid {
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -181,7 +181,7 @@ func dialWebRTC(
 	exchangeCtx, exchangeCancel := context.WithCancel(signalCtx)
 	defer exchangeCancel()
 
-	// atomic "bool" representing whether initial sdp exchange has occured
+	// atomic bool representing whether initial sdp exchange has occured
 	var haveInit atomic.Bool
 	haveInit.Store(false)
 
--- a/rpc/wrtc_client.go
+++ b/rpc/wrtc_client.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
@@ -182,12 +181,11 @@ func dialWebRTC(
 	defer exchangeCancel()
 
 	// atomic bool representing whether initial sdp exchange has occurred
-	var haveInit atomic.Bool
-	haveInit.Store(false)
+	haveInit := true
 
 	errCh := make(chan error)
 	sendErr := func(err error) {
-		if haveInit.Load() && isEOF(err) {
+		if haveInit && isEOF(err) {
 			logger.Warnf("caller swallowing err %v", err)
 			return
 		}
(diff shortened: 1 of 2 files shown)


rpc/wrtc_client.go
```diff
@@ -203,7 +203,7 @@ func dialWebRTC(
 				},
 			})
 		})
-		return err
+		return filterEOF(err, logger)
```
danielbotros, line 206: this is in `sendDone()` where I imagine that the caller / answerer tries to `sendDone()` but the signaling server has already severed the connection leading to an EOF

rpc/wrtc_signaling_answerer.go
```diff
@@ -509,7 +509,11 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 		utils.PanicCapturingGoWithCallback(func() {
 			defer close(done)
 			if err := exchangeCandidates(); err != nil {
-				sendErr(err)
+				if filterEOF(err, ans.logger) == nil {
```
danielbotros, line 512: there is no `haveInit` here because since the first thing the answerer does is receive a `webrtcpb.AnswerRequest_Init`, so `haveInit` will be  implicitly true if we ever make it this far

rpc/wrtc_client.go
```diff
@@ -433,3 +438,12 @@ func dialSignalingServer(
 	conn, _, err := dialDirectGRPC(ctx, signalingServer, dOpts, logger)
 	return conn, err
 }
+
+func filterEOF(err error, logger utils.ZapCompatibleLogger) error {
+	s, isGRPCErr := status.FromError(err)
+	if errors.Is(err, io.EOF) || (isGRPCErr && (s.Code() == codes.Internal && strings.Contains(s.Message(), "EOF"))) {
```
danielbotros, line 444: was not sure if this should be `io.EOF` errors since that is also what I see in the flaky test logs or grpc EOF errors

rpc/wrtc_signaling_answerer.go
danielbotros: also, this could just be simplified to `sendErr(filterEOF(err))` but I thought the log line indicating EOF during ICE candidate exchange would be useful

benjirewis (changes requested):
Nice! Can we be even more aggressive and swallow (+ log) all `EOF`s **from within** `sendErr` in `wrtc_client.go` (when `haveInit == true`) and `sendErr` in `wrtc_signaling_answerer.go`?

rpc/wrtc_client.go
```diff
@@ -180,8 +181,15 @@ func dialWebRTC(
 	exchangeCtx, exchangeCancel := context.WithCancel(signalCtx)
 	defer exchangeCancel()
 
+	// atomic "bool" representing whether initial sdp exchange has occured
+	var haveInit int32
```
benjirewis, line 185: Why not https://pkg.go.dev/sync/atomic#Bool?

danielbotros: somehow missed this thanks for pointing out

rpc/wrtc_client.go
benjirewis: What you have looks good to me; I'm not sure gRPC EOF's actually wrap `io.EOF`, so your code here is probably necessary.

rpc/wrtc_signaling_answerer.go
```diff
@@ -331,12 +331,12 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
...
 
-	errCh := make(chan interface{})
+	errCh := make(chan error)
 	defer exchangeCancel()
-	sendErr := func(err interface{}) {
+	sendErr := func(err error) {
 		select {
 		case <-exchangeCtx.Done():
-		case errCh <- err:
+		case errCh <- filterEOF(err, ans.logger):
```
benjirewis, line 339: Won't this send `nil` to `errCh`? Do we want that?

danielbotros: yeah probably not, I think this would close the server / client channel, channels are still a bit foreign to me

rpc/wrtc_client.go
```diff
@@ -206,7 +213,11 @@ func dialWebRTC(
 		return err
 	}
 
+	// this channel blocks goroutines spawned for each ICE candidate in OnIceCandidate from sending a CallUpdateRequest
+	// to the signaling server until a CallResponse_Init is recieved, which in turn causes the channel to be closed and
```
dgottlieb, line 217: received

rpc/wrtc_client.go
```diff
@@ -180,8 +180,15 @@ func dialWebRTC(
 	exchangeCtx, exchangeCancel := context.WithCancel(signalCtx)
 	defer exchangeCancel()
 
+	// bool representing whether initial sdp exchange has occurred
+	haveInit := true
```
dgottlieb, line 184: I don't think this variable is doing anything useful right now

rpc/wrtc_client.go
```diff
@@ -303,7 +314,6 @@ func dialWebRTC(
 	clientCh := newWebRTCClientChannel(peerConn, dataChannel, onICEConnected, logger, dOpts.unaryInterceptor, dOpts.streamInterceptor)
 
 	exchangeCandidates := func() error {
-		haveInit := false
```
dgottlieb, line 306: I can't comment off-diff, but [I think this "swallow EOF"s isn't right](https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_client.go#L315-L321). I think that should be:
```
			callResp, err := callClient.Recv()
			if err != nil {
				if !haveInit && !errors.Is(err, io.EOF) {
					return err
				}
				return nil
			}
```

benjirewis: Can we just remove the swallowing from that section and let `sendErr` handle the swallowing?

rpc/wrtc_signaling_answerer.go
```diff
@@ -331,9 +331,13 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
...
 
-	errCh := make(chan interface{})
+	errCh := make(chan error)
 	defer exchangeCancel()
-	sendErr := func(err interface{}) {
+	sendErr := func(err error) {
+		if isEOF(err) {
+			ans.logger.Warnf("answerer swallowing err %v", err)
+			return
+		}
```
benjirewis, line 340: Can we put another warn log line after this block showing the error with `%v` and `%T` (type)? If an error besides `EOF` slips into here, we want to know what it was so we can decide if we want to add it to our "allowlist" (currently we're just allowing `EOF`).

rpc/wrtc_client.go
```diff
@@ -180,8 +180,15 @@ func dialWebRTC(
...
 
+	// bool representing whether initial sdp exchange has occurred
+	haveInit := true
+
 	errCh := make(chan error)
 	sendErr := func(err error) {
+		if haveInit && isEOF(err) {
+			logger.Warnf("caller swallowing err %v", err)
+			return
+		}
```
benjirewis, line 191: See my other comment in the answerer; can we log the error even if we're not swallowing it?

rpc/wrtc_client.go
benjirewis: `haveInit` is useful for returning WebRTC client errors when two init stages are received or an update comes in before an init. I agree, though, that it seems like the two of you have found that `sendErr` won't actually happen until an init stage has been received? So, it seems like we don't need to move it up here, and there's no need to check its value in `sendErr`?

benjirewis (changes requested):

rpc/wrtc_client.go
dgottlieb: I think `haveInit` is useful yes -- sorry if I was being cryptic. IIUC the code in this PR initializes `haveInit` to true and nothing ever changes that.

benjirewis: Ah true; if we _were_ to keep it here, it should probably start out as false.

rpc/wrtc_client.go
danielbotros: yeah we can just return err and the goroutine that calls `exchangeCandidates()` will `sendErr`

dgottlieb (approved):
pending Benji's comments

benjirewis (approved):
Nice; LGTM! Thanks 🧑‍🔧 .

PR #316, opened 2024-07-26
RSDK-8377: Allow RDK to maintain track of Loggers that call With()

danielbotros:
### Description
RSDK-8377: Allow RDK to maintain track of Loggers that call With()
* Adds `LogWith()` util function that calls `With()` on `zap.SugaredLoggers` and `WithField()` on RDK loggers using reflection so that we do not lose track of RDK loggers that call `With()` in `goutils`.
* Change all instances of `With()` to `utils.LogWith()`
### Testing
* A few tests to ensure that the correct one of `With()` / `WithField()` is called depending on the type of logger.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -24,7 +24,6 @@ type ILogger interface {
 // ZapCompatibleLogger is a basic logging interface for golog.Logger (alias for *zap.SugaredLogger) and RDK loggers.
 type ZapCompatibleLogger interface {
 	Desugar() *zap.Logger
-	With(args ...interface{}) *zap.SugaredLogger
 
 	Debug(args ...interface{})
 	Debugf(template string, args ...interface{})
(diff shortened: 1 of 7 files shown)


logger.go
```diff
@@ -79,3 +78,45 @@ func Sublogger(inp ZapCompatibleLogger, subname string) (loggerRet ZapCompatible
...
+func LogWith(inp ZapCompatibleLogger, args ...interface{}) (loggerRet ZapCompatibleLogger) {
+	loggerRet = inp
+
+	defer func() {
+		if r := recover(); r != nil {
+			inp.Debugf("panic occurred while adding fields to logger: %v, returning self", r)
+		}
+	}()
+
+	reflectArgs := make([]reflect.Value, len(args)+1)
```
dgottlieb, line 95: I looked at this after the RDK patch -- and now I see we're using reflection again! Makes sense given `With` had that same "type A is an interface that generates more As" property.

logger.go
```diff
@@ -79,3 +78,45 @@ func Sublogger(inp ZapCompatibleLogger, subname string) (loggerRet ZapCompatible
...
+	reflectArgs[0] = reflect.ValueOf(inp)
```
dgottlieb, line 96: I think a small word about the convention here would be helpful. Specifically:
```
When using reflection to call receiver methods, the first argument must be the object. The remaining arguments are the actual function parameters.
```

Apologies if that critique have been applied to the prior CR and I'm being inconsistent.

logger.go
```diff
@@ -79,3 +78,45 @@ func Sublogger(inp ZapCompatibleLogger, subname string) (loggerRet ZapCompatible
...
+// LogWith adds fields for logging to a given ZapCompatibleLogger instance.
+// This function uses reflection to dynamically add fields to the provided logger by
+// calling its `WithFields` method if it is an RDK logger, or its `With` method if it is a Zap logger.
+// If neither method is available, it logs a debug message and returns the original logger.
+func LogWith(inp ZapCompatibleLogger, args ...interface{}) (loggerRet ZapCompatibleLogger) {
+	loggerRet = inp
+
+	defer func() {
+		if r := recover(); r != nil {
+			inp.Debugf("panic occurred while adding fields to logger: %v, returning self", r)
```
dgottlieb, line 91: The use of a named return value for defer+recover is perfectly idiomatic, but it's not used super often.

So I think a comment here saying this case will return `loggerRet` that was initialized to `inp` would help a reader. And maybe adding it's intentionally never re-assigned before calling functions can panic would be good.

logger.go
```diff
@@ -79,3 +78,45 @@ func Sublogger(inp ZapCompatibleLogger, subname string) (loggerRet ZapCompatible
 
 	return loggerRet
 }
+
+// LogWith adds fields for logging to a given ZapCompatibleLogger instance.
+// This function uses reflection to dynamically add fields to the provided logger by
+// calling its `WithFields` method if it is an RDK logger, or its `With` method if it is a Zap logger.
+// If neither method is available, it logs a debug message and returns the original logger.
+func LogWith(inp ZapCompatibleLogger, args ...interface{}) (loggerRet ZapCompatibleLogger) {
```
benjirewis, line 86: [nit] I find the name of this function _slightly_ confusing. It's more adding the fields to `inp`, while `LogWith` for some reason, to me, implies there's some "global" logger we're now "logging with". Could we call this `AddFieldsToLogger`? Open to counterarguments/other naming suggestions, too.

danielbotros: I like `AddFieldsToLogger` enough, I kept the word With to be consistent with the functions it calls, but definitely not necessary

dgottlieb (approved):

benjirewis (approved):
LGTM!

PR #318, opened 2024-08-06
RSDK-8404: Add logging around incoming GRPC requests + Refactor AddFieldsToLogger function to account for new logger type

danielbotros:
### Description
RSDK-8404: Add logging around incoming GRPC requests.

Also, RDK logger `WithFields()` now returns a new logger instead of a modifying the current logger, so this function needed a small refactor to account for that

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -607,7 +607,7 @@ func NewServer(logger utils.ZapCompatibleLogger, opts ...ServerOption) (Server, 
 				server.webrtcServer,
 				sOpts.webrtcOpts.ExternalSignalingDialOpts,
 				config,
-				utils.Sublogger(logger, "external_signaler"),
+				utils.Sublogger(logger, "signaler.external"),
 				true, // logStats == true
 			))
 		} else {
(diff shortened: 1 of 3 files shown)


rpc/wrtc_server_stream.go
```diff
@@ -263,6 +263,7 @@ func isContextCanceled(err error) bool {
 
 func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
 	s.logger = utils.AddFieldsToLogger(s.logger, "method", headers.Method)
+	s.logger.Debug("incoming grpc request")
```
benjirewis, line 266:
```suggestion
	s.logger.Debugw("Incoming gRCP request", "method", headers.Method)
```

danielbotros: since we add the method as a field in the line above, the suggested change will log the method name twice in the same log line.

benjirewis: Oops and I mispelled `grpc`. Ignore this comment.

benjirewis (approved):

rpc/server.go
```diff
@@ -651,7 +651,7 @@ func NewServer(logger utils.ZapCompatibleLogger, opts ...ServerOption) (Server,
 				server.webrtcServer,
 				answererDialOpts,
 				config,
-				utils.Sublogger(logger, "internal_signaler"),
+				utils.Sublogger(logger, "signaler.external"),
```
benjirewis, line 654:
```suggestion
				utils.Sublogger(logger, "signaler.internal"),
```

benjirewis (changes requested):

rpc/server.go
danielbotros: oops

benjirewis (approved):

PR #319, opened 2024-08-08
RSDK-8374: Open data channels earlier in client and answerer

puranamp:


benjirewis:
Could you verify locally that this gets rid of the hang we were reproing? Let me know if you need help getting the repro running.

benjirewis (approved):
LGTM (see my edited comment about not reproing).

dgottlieb (approved):

PR #320, opened 2024-08-09
RSDK-8508: Log all connection failure errors.

dgottlieb:


rpc/wrtc_signaling_answerer.go
```diff
@@ -194,19 +194,39 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
...
 			if err == nil {
-				err = ans.answer(client)
+				// `ans.answer` will send the initial message to the signaling server that says it
+				// is ready to accept connections. Then it waits, typically for a long time, for a
+				// caller to show up. Which is when the signaling server will send a
+				// response. `ans.answer` then follows with gathering ICE candidates and learning of
+				// the caller's ICE candidates to create a working WebRTC PeerConnection. If
+				// successful, the `PeerConnection` + `webrtcServerChannel` will be registered and
+				// available for the `webrtcServer`.
+				receivedInitRequest, err = ans.answer(client)
```
dgottlieb, line 208: I wasn't thinking too hard about options here. I just took the dumbest patch possible that didn't touch the scope/responsibility of this `ans.answer` function call.

rpc/wrtc_signaling_answerer.go
```diff
@@ -247,16 +267,18 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 		}
 	}()
 
+	receivedInitRequest = false
```
dgottlieb, line 270: The whole patch is really just this and the next 2 lines. Only `client.Recv` erroring should result in a "return false" and every other return case should have "receivedInitRequest = true".

rpc/wrtc_signaling_answerer.go
```diff
@@ -333,7 +355,7 @@ func (ans *webrtcSignalingAnswerer) answer(client webrtcpb.SignalingService_Answ
 			ans.logger.Warnf("answerer swallowing err %v", err)
 			return
 		}
-		ans.logger.Warnf("caller received err %v of type %T", err, err)
+		ans.logger.Warnf("answerer received err %v of type %T", err, err)
```
dgottlieb, line 358: This was a drive by. I think* answerer was the correct noun here.

rpc/wrtc_signaling_answerer.go
```diff
@@ -194,19 +194,39 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
...
-			// errors represent that an operation has failed, but can be safely retried over the
-			// existing connection.
-			if checkExceptionalError(err) != nil {
-				ans.logger.Errorw("error answering", "error", err)
+
+			switch {
+			case err == nil:
+			case receivedInitRequest && err != nil:
+				// We received an error while trying to connect to a caller/peer.
+				ans.logger.Errorw("error connecting", "error", err)
```
benjirewis, line 215: A little confused about these cases. If we've received the init request, doesn't that mean we had an `error answering` and _not_ an `error connecting`?

rpc/wrtc_signaling_answerer.go
benjirewis: Yep; thanks for the typo fix.

rpc/wrtc_signaling_answerer.go
dgottlieb: I think you and I have different ideas what connecting means. Will try different words

benjirewis (approved):
Yeah your new messages match my understanding (I interpreted the bare `connecting` differently).

PR #322, opened 2024-08-13
RSDK-8525: New answerer workers adding to a waitgroup must atomicly perform their `closeCtx` read and `wg.Add` call.

dgottlieb:


dgottlieb:
I'm going to merge this and star the PR on RDK. Can abort that PR if there's some risk we haven't identified yet

PR #324, opened 2024-08-14
RSDK-8529: Fix signaling server background worker accounting.

dgottlieb:


rpc/wrtc_signaling_server.go
```diff
@@ -470,9 +467,21 @@ func (srv *WebRTCSignalingServer) Answer(server webrtcpb.SignalingService_Answer
...
+		return srv.cancelCtx.Err()
+	}
+	srv.callWorkers.Add(1)
+	srv.callMu.RUnlock()
+
 	callerErrCh := make(chan error, 1)
 	utils.PanicCapturingGo(func() {
-		defer close(callerErrCh)
+		defer func() {
+			close(callerErrCh)
```
dgottlieb, line 482: This one was the bug in the test

rpc/wrtc_signaling_server.go
```diff
@@ -164,16 +164,13 @@ func (srv *WebRTCSignalingServer) Call(req *webrtcpb.CallRequest, server webrtcp
...
-		// we assume the number of goroutines is bounded by the gRPC server invoking this method.
-		srv.callWorkers.Add(1)
-		srv.callMu.RUnlock()
-
-		// Check if cancelCtx has errored: underlying server may have been
-		// `Close`ed, in which case we mark this call worker as `Done` and return.
+		// Atomically check if the cancelCtx was canceled, and if not, add to the `callWorkers`.
 		if err := srv.cancelCtx.Err(); err != nil {
-			srv.callWorkers.Done()
+			srv.callMu.RUnlock()
```
dgottlieb, line 169: Based on my internal model, this wasn't technically a bug (and why that is bothers me that I don't have a great way of describing the meat of the issue). But figured it was worth making it all look the same.

rpc/wrtc_signaling_server.go
benjirewis: Sorry which test? The data race you saw yesterday? We weren't keeping track of the `callerLoop` as a worker and it was trying to log after signal server close?

benjirewis (commented):
Seems good to me; just a question.

rpc/wrtc_signaling_server.go
dgottlieb: Correct, https://viam.atlassian.net/browse/RSDK-8529 was a "data race" because of logging after the test ended. And that was due this goroutine going unaccounted for.

benjirewis (approved):
Thanks!

PR #326, opened 2024-08-14
RSDK-8189: Bump webrtc for PeerConnection.GracefulClose hang fix.

dgottlieb:


benjirewis (approved):

dgottlieb:
So the prior test run found a hang we hadn't seen before. We believe this hang was due to another recent change to prevent leaks.

So I've bumped to 3.99.7 now and giving this a run. 3.99.7 reverted a fix for leaked goroutines. Thus, I believe this means, with the current patch in this PR, `GracefulClose` can still leak goroutines. Notably, I think the "operations queue" can create a data channel + start a read loop after peerconnection.close has finished waiting for background workers.

edaniels:
> So the prior test run found a hang we hadn't seen before. We believe this hang was [due to another recent change to prevent leaks](https://github.com/viamrobotics/webrtc/commit/887b10f2641dd140b62550671c151d8ff42fb10f).
> 
> So I've bumped to 3.99.7 now and giving this a run. 3.99.7 [reverted](https://github.com/viamrobotics/webrtc/commit/a49914b2536bb50f57904385a2d016e267e92fa2) the above fix for leaked goroutines. Thus, I believe this means, with the current patch in this PR, `GracefulClose` can still leak goroutines. Notably, I think the "operations queue" can create a data channel + start a read loop after peerconnection.close has finished waiting for background workers.
> 
> But I was relatively convinced these new data channel read loops would be short lived. So hopefully our timeout CI failures are gone and we just get goleak errors instead.

I actually don't think that can happen. Because of https://github.com/pion/sctp/commit/18c401513a010e9989689ae4af54d154a44d24ca, if the operation that adds a data channel during closing happens, it will either be rejected or be added to the list of data channels to close once we get to that part of the GracefulClose code.

What could leak are other goroutines that do something that are yet to abide by internal consistency rules.

dgottlieb:
Ah, nice. Very happy to hear that less bad stuff might happen than what I'm predicting!

edaniels:
Going to need https://github.com/pion/sctp/pull/350 as well to really support my statement 🙃

PR #332, opened 2024-08-19
DATA-3010 - allow method to access *grpc.ClientConn's GetState method from an rpc.ClientConn

nicksanford:
[Ticket](https://viam.atlassian.net/browse/DATA-3010)

1. Add `GRPCConn() GRPCClientConnInterface` method to the `rpc.ClientConn` interface
2. Add `rpc.GRPCClientConnInterface` type which allows other types to implement to the subset of the methods of `*rpc.ClientConn` viam needs, starting with just `GetState() connectivity.State` to allow checking whether or not the underlying GRPC connection is healthy.
3. Add `rpc.GrpcOverHTTPClientConn.GRPCConn()` returning the underlying `*rpc.ClientConn`
4. Add `webrtcClientChannel.GRPCConn()` returning nil (as it is a webrtc based connection)

### DIFF
--- a/rpc/dialer.go
+++ b/rpc/dialer.go
@@ -19,6 +19,7 @@ import (
 	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/keepalive"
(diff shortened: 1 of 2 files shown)


rpc/dialer.go
```diff
@@ -79,6 +83,17 @@ type GrpcOverHTTPClientConn struct {
 	*grpc.ClientConn
 }
 
+// GRPCClientConnInterface is the allow list of methods
+// we expect to be able to call on a *grpc.ClientConn.
+type GRPCClientConnInterface interface {
```
nicksanford, line 88: I decided to return an interface so that we can control what subset of the grpc.ClientConn methods we want to sign up to support for callers of the goutils rpc package.

It also allows us to mock out the connectivity.State in tests which is useful.

cheukt (approved):
LGTM

PR #333, opened 2024-08-23
[RSDK-8570] Add StoppableWorkers from RDK

penguinland:
Having it in RDK also bit me again today: during 20% Happy Hour, I tried using it in the `logging` package, but introduced a circular dependency (logging would import utils, which imports logging). I haven't needed to tweak the implementation in months; I think this is ready for goutils!

The linter and compiler are happy. I haven't tried this copy out more than that, but the first commit is a byte-for-byte copy of the widely-used version in RDK, and then the second commit is adapting it for goutils.

Benji thought we could get rid of the interface/struct split, and I'm open to such a change, though don't see how to do it myself. but in the meantime, here's the start.

### DIFF
--- a/stoppable_workers.go
+++ b/stoppable_workers.go
@@ -2,10 +2,14 @@ package utils
 
 import (
 	"context"
+	"errors"
 	"sync"
 )
 
-// StoppableWorkers is a collection of goroutines that can be stopped at a later time.
+var StoppableWorkersAlreadyStopped = errors.New("cannot add worker: already stopped")
+
+// StoppableWorkers is a collection of goroutines that can be stopped at a
+// later time.
 type StoppableWorkers struct {
 	mu         sync.RWMutex
 	ctx        context.Context
@@ -14,27 +18,25 @@ type StoppableWorkers struct {
 	workers sync.WaitGroup
 }
 
-// NewStoppableWorkers creates a new StoppableWorkers instance.
-func NewStoppableWorkers() *StoppableWorkers {
-	ctx, cancelFunc := context.WithCancel(context.Background())
+// NewStoppableWorkers creates a new StoppableWorkers instance. The instance's
+// context will derived from passed in context.
+func NewStoppableWorkers(ctx context.Context) *StoppableWorkers {
+	ctx, cancelFunc := context.WithCancel(ctx)
 	return &StoppableWorkers{ctx: ctx, cancelFunc: cancelFunc}
 }
 
-// AddWorker starts up a goroutine for the passed-in function. If you call this
-// after calling Stop, it will return immediately without starting a new
-// goroutine. Workers:
+// Add starts up a goroutine for the passed-in function. Workers:
 //
-//   - MUST accept a `context.Context` and return nothing.
-//   - MUST respond appropriately to errors on that context.
-//   - MUST NOT add more workers to the `StoppableWorkers` group to which they
-//     belong.
+//   - MUST respond appropriately to errors on the context parameter.
+//   - MUST NOT add more workers to the `StoppableWorkers` group to which
+//     they belong.
 //
 // Any `panic`s from workers will be `recover`ed and logged.
-func (sw *StoppableWorkers) AddWorker(worker func(context.Context)) {
+func (sw *StoppableWorkers) Add(worker func(context.Context)) error {
 	// Read-lock to allow concurrent worker addition. The Stop method will write-lock.
 	sw.mu.RLock()
 	if sw.ctx.Err() != nil {
-		return
+		return StoppableWorkersAlreadyStopped
 	}
 	sw.workers.Add(1)
 	sw.mu.RUnlock()
@@ -43,12 +45,16 @@ func (sw *StoppableWorkers) AddWorker(worker func(context.Context)) {
 		defer sw.workers.Done()
 		worker(sw.ctx)
 	})
+	return nil
 }
 
-// Stop shuts down all the goroutines we started up.
+// Stop idempotently shuts down all the goroutines we started up.
 func (sw *StoppableWorkers) Stop() {
 	sw.mu.Lock()
 	defer sw.mu.Unlock()
+	if sw.ctx.Err() != nil {
+		return
+	}
 
 	sw.cancelFunc()
 	sw.workers.Wait()
--- a/stoppable_workers.go
+++ b/stoppable_workers.go
@@ -19,7 +19,7 @@ type StoppableWorkers struct {
 }
 
 // NewStoppableWorkers creates a new StoppableWorkers instance. The instance's
-// context will derived from passed in context.
+// context will be derived from passed in context.
 func NewStoppableWorkers(ctx context.Context) *StoppableWorkers {
 	ctx, cancelFunc := context.WithCancel(ctx)
 	return &StoppableWorkers{ctx: ctx, cancelFunc: cancelFunc}
--- a/stoppable_workers.go
+++ b/stoppable_workers.go
@@ -3,13 +3,7 @@ package utils
 import (
 	"context"
 	"sync"
-
-	goutils "go.viam.com/utils"
 )
-
-// TODO: When this struct is widely used and feature complete, move this to goutils instead of
-// here. Until then, we cannot use this in any package imported by utils (e.g., the logging
-// package) without introducing a circular import dependency.
 
 // StoppableWorkers is a collection of goroutines that can be stopped at a later time.
 type StoppableWorkers interface {
@@ -54,7 +48,7 @@ func (sw *stoppableWorkersImpl) AddWorkers(funcs ...func(context.Context)) {
 		// the loop before the goroutine starts up, it starts this function instead of the next
 		// one. For details, see https://go.dev/blog/loopvar-preview
 		f := f
-		goutils.PanicCapturingGo(func() {
+		PanicCapturingGo(func() {
 			defer sw.activeBackgroundWorkers.Done()
 			f(sw.cancelCtx)
 		})
--- a/stoppable_workers.go
+++ b/stoppable_workers.go
@@ -6,66 +6,50 @@ import (
 )
 
 // StoppableWorkers is a collection of goroutines that can be stopped at a later time.
-type StoppableWorkers interface {
-	AddWorkers(...func(context.Context))
-	Stop()
-	Context() context.Context
+type StoppableWorkers struct {
+	mu         sync.RWMutex
+	ctx        context.Context
+	cancelFunc func()
+
+	workers sync.WaitGroup
 }
 
-// stoppableWorkersImpl is the implementation of StoppableWorkers. The linter will complain if you
-// try to make a copy of something that contains a sync.WaitGroup (and returning a value at the end
-// of NewStoppableWorkers() would make a copy of it), so we do everything through the
-// StoppableWorkers interface to avoid making copies (since interfaces do everything by pointer).
-type stoppableWorkersImpl struct {
-	mu                      sync.Mutex
-	cancelCtx               context.Context
-	cancelFunc              func()
-	activeBackgroundWorkers sync.WaitGroup
+// NewStoppableWorkers creates a new StoppableWorkers instance.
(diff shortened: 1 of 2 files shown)


stoppable_workers.go
```diff
@@ -0,0 +1,71 @@
...
+type StoppableWorkers interface {
+	AddWorkers(...func(context.Context))
+	Stop()
+	Context() context.Context
+}
+
+// stoppableWorkersImpl is the implementation of StoppableWorkers. The linter will complain if you
+// try to make a copy of something that contains a sync.WaitGroup (and returning a value at the end
+// of NewStoppableWorkers() would make a copy of it), so we do everything through the
+// StoppableWorkers interface to avoid making copies (since interfaces do everything by pointer).
```
benjirewis: We do want to return a pointer from the constructor, but we can just have `StoppableWorkers` be a struct and return `*StoppableWorkers` instead of returning an interface. Removed the interface definition.

stoppable_workers.go
```diff
@@ -0,0 +1,71 @@
...
+type stoppableWorkersImpl struct {
+	mu                      sync.Mutex
```
benjirewis: As mentioned offline, NetCode has been using `sync.RWMutex`es for groups of goroutines. We use these instead of `sync.Mutex` so that additions of new workers do not lock with each other in any way. We want as-fast-as-possible worker addition; so _read-locking_ only around context-checking and waitgroup-addition makes the most sense to us. `Stop` is the thing that should write-lock and "stop the world".

stoppable_workers.go
```diff
@@ -0,0 +1,71 @@
...
+	cancelCtx               context.Context
+	cancelFunc              func()
+	activeBackgroundWorkers sync.WaitGroup
+}
+
+// NewStoppableWorkers runs the functions in separate goroutines. They can be stopped later.
+func NewStoppableWorkers(funcs ...func(context.Context)) StoppableWorkers {
```
benjirewis: I changed the design here a bit and did not accept any workers in the constructor. I also changed `AddWorkers` to `Add` and had it accept only a single worker. I'd rather the pattern be that developers create a `StoppableWorkers` instance they store on _their_ struct, and then add workers _one at a time_. Adding multiple workers at once opens us up to more questions like "what if a `Stop` comes in between the addition of the second and third worker?" NetCode has found this logic to be exceedingly fragile, so trying to stay as simple as possible here.

stoppable_workers.go
```diff
@@ -0,0 +1,71 @@
...
+func (sw *stoppableWorkersImpl) Stop() {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+
+	sw.cancelFunc()
+	sw.activeBackgroundWorkers.Wait()
+}
+
+// Context gets the context the workers are checking on. Using this function is expected to be
+// rare: usually you shouldn't need to interact with the context directly.
```
benjirewis: I removed the `Context` method for exactly the reason you've pointed out here. I would rather completely disallow accessing the context directly unless there's some circumstance for which you need it in driver logic.

stoppable_workers.go
```diff
@@ -0,0 +1,64 @@
...
+	mu         sync.RWMutex
+	ctx        context.Context
+	cancelFunc func()
+
+	workers sync.WaitGroup
+}
+
+// NewStoppableWorkers creates a new StoppableWorkers instance. The instance's
+// context will be derived from passed in context.
+func NewStoppableWorkers(ctx context.Context) *StoppableWorkers {
```
benjirewis, line 25: I think `NewStoppableWorkers` should not always derive from `context.Background()` (hence the context parameter here). See the `nested workers` sub test I wrote; NetCode has objects that manage workers, and some of those workers _have their own_ workers. We want to allow establishing a context hierarchy such that a `StoppableWorkers` instance can be `Stop`ped and, in doing so, stop the workers of any of its workers `StoppableWorkers` instances.

benjirewis (commented):
Thanks so much for helping move this to goutils @penguinland ! I've revised the code a bit, added some tests, and left some GH comments justifying my changes. Curious what your thoughts are. I'm also bringing in other members of NetCode to take a look.

cheukt:
I do like the changes @benjirewis made, but wondering if that'll cause issues when replacing it in RDK @penguinland

stoppable_workers.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Add starts up a goroutine for the passed-in function. Workers:
+//
+//   - MUST respond appropriately to errors on the context parameter.
+//   - MUST NOT add more workers to the `StoppableWorkers` group to which
+//     they belong.
+//
+// Any `panic`s from workers will be `recover`ed and logged.
+func (sw *StoppableWorkers) Add(worker func(context.Context)) error {
```
randhid, line 37: I remember Alan had to make it take a variardic input to determine the additional workers at runtime; in one commit, you can find them in the StoppableWorkers' file commit history.

stoppable_workers.go
```diff
@@ -0,0 +1,64 @@
...
+	mu         sync.RWMutex
+	ctx        context.Context
+	cancelFunc func()
+
+	workers sync.WaitGroup
+}
+
+// NewStoppableWorkers creates a new StoppableWorkers instance. The instance's
+// context will be derived from passed in context.
+func NewStoppableWorkers(ctx context.Context) *StoppableWorkers {
```
randhid, line 25: I like the pattern of having each stoppable worker make its own context; it made sense since the components with multiple workers could all form their contexts and be self-contained in construction and close. Want to make another creator function that gives you your behaviour instead?

stoppable_workers.go
```diff
@@ -0,0 +1,64 @@
...
+func (sw *StoppableWorkers) Stop() {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+	if sw.ctx.Err() != nil {
+		return
+	}
+
+	sw.cancelFunc()
+	sw.workers.Wait()
+}
```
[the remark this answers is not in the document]
benjirewis: Yep.

stoppable_workers_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+func normalWorker(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(10 * time.Millisecond):
+		}
+	}
+}
```
randhid, line 64: Could we add tests where the worker is doing something a bit more active too? We've got workers reading and writing from serial/i2c/spi protocols, workers returning immediately after math.

I'm also worried about the done context argument being passed it, it was an excellent safety rail for resource construction and not passing in a context with a cancel or timeout you don't know about.

randhid (commented):
Benji, I will trade you StoppableWorkers for getting rid of reconfigure.

But serious q - we have many functions for spawning off goroutines that wrap others `ManagedGo`, `PanicCapturingGo`, why not add a few here for flexibility for the netcode team?

While you're not a bucketeer, could you test this works with a viam rover? We have two components using StoppableWorkers on there.

cc @JohnN193 @martha-johnston @oliviamiller @susmitaSanyal for visibility.
Places this is used: https://github.com/search?q=repo%3Aviamrobotics%2Frdk+StoppableWorkers&type=code

stoppable_workers.go
benjirewis: Yep can do.

stoppable_workers_test.go
benjirewis: > Could we add tests where the worker is doing something a bit more active too?

Yep!

> I'm also worried about the done context argument being passed it

Slightly confused what you're referencing here. That workers get passed the `StoppableWorkers` context?

randhid: hmm, maybe also a test that adds a few workers

stoppable_workers.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// ErrStoppableWorkersAlreadyStopped is returned by Add when the StoppableWorkers
+// instance has already been stopped.
+var ErrStoppableWorkersAlreadyStopped = errors.New("cannot add worker: already stopped")
+
+// StoppableWorkers is a collection of goroutines that can be stopped at a
+// later time.
+type StoppableWorkers struct {
+	mu         sync.RWMutex
```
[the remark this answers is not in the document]
randhid: Add comment in code to explain.

stoppable_workers_test.go
randhid: > Slightly confused what you're referencing here. That workers get passed the StoppableWorkers context?

No, that's fine and expected; I was also worried about the context being passed into `NewStoppableWorkers`, but you're making another creator function so I won't be anymore.

stoppable_workers.go
randhid: I think he made it variardic so that you could pass in many functions to use the same context,  but I guess this works the same now that I think about it. It'll just have to be called in a loop.

benjirewis (commented):
Put up a [draft PR](https://github.com/viamrobotics/rdk/pull/4325) for `rdk`. That's how I imagine we'd switch to this version of `StoppableWorkers`. Bucketeers should let me know what you think.

randhid:
@benjirewis let me know when you're ready to test

benjirewis:
@JohnN193 , @randhid and I tested offline with real hardware; an RDK with [these edits](https://github.com/viamrobotics/rdk/pull/4325) seems to work just fine. I've added better concurrent testing, too (actual communication between goroutines). Ready for a final look from you guys @cheukt + @maximpertsov .

maximpertsov:
@benjirewis @cheukt I'm removing myself from this PR since there are a few reviewers already and I don't think I can get to it in a reasonable time frame - please let me know if you need my input on anything specific.

cheukt (approved):
LGTM!

randhid (approved):
LGTM! Not blocking merge, but we talked about adding more testing cases, is that being done in a follow up commit?

benjirewis:
> but we talked about adding more testing cases, is that being done in a follow up commit?

Spoke offline. Current tests seem fine for now, and we can add more as encounter weird usage cases.

stoppable_workers.go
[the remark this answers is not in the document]
penguinland: Yeah, `Context()` was not in the original plan, and I begrudgingly added it because there were a bunch of places I wanted to use `StoppableWorkers` that needed it.

stoppable_workers.go
```diff
@@ -0,0 +1,88 @@
...
+	return &StoppableWorkers{ctx: ctx, cancelFunc: cancelFunc}
+}
+
+// NewBackgroundStoppableWorkers creates a new StoppableWorkers instance. The
+// instance's context will be derived from `context.Background()`. The passed
+// in workers will be `Add`ed. Workers:
+//
+//   - MUST respond appropriately to errors on the context parameter.
+//   - MUST NOT add more workers to the `StoppableWorkers` group to which
+//     they belong.
```
penguinland, line 34: Why not? I don't see any obvious problems with a worker spawning another worker. There's no deadlock in here: the first worker's goroutine would block until the StoppableWorkers' mutex was unlocked, then spawn the new goroutine afterward.

benjirewis: What if:
1. A worker is `Add`ed
2. `Stop` is called
      a. Mutex is `Lock`ed
      b. Context is canceled
      b. `Wait` is called
3. The worker adds another subworker _before it has a chance to check its context_
5. That second worker gets caught on `RLock`ing the locked mutex
6. The first worker never marks itself as `Done()` as the worker it spawns gets caught on `RLock`
7. `Wait` hangs forever

I realize this may seem like a farfetched scenario, but NetCode has encountered it in our code, and I have found it dangerous to have "subworkers" accessing the same concurrency primitives that their parents do. We can avoid complex situations like the one above by ensuring that workers with subworkers have their _own_ `StoppableWorkers` instance.

penguinland: Ah, very good insight! Thanks for explaining.

benjirewis: No prob 😎 . Thanks for writing all the original code here, too.

PR #348, opened 2024-09-06
RSDK-8480 - Actually pass server impl into auth stream handler

cheukt:
this has been an existing bug for at least a few years - we get the server panic but as long as we close down properly the error actually gets lost as streamClients don't error even if the server side definitely did, the error only shows up after we try to receive the message.

### DIFF
--- a/rpc/server_auth.go
+++ b/rpc/server_auth.go
@@ -216,7 +216,7 @@ func (ss *simpleServer) authStreamInterceptor(
 			return err
 		}
 		serverStream = ctxWrappedServerStream{serverStream, nextCtx}
-		return handler(nextCtx, serverStream)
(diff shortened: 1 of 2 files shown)


rpc/server_auth.go
```diff
@@ -216,7 +216,7 @@ func (ss *simpleServer) authStreamInterceptor(
 			return err
 		}
 		serverStream = ctxWrappedServerStream{serverStream, nextCtx}
-		return handler(nextCtx, serverStream)
```
dgottlieb, line 219: > It seems the `srv` is actually a `valueCtx`.

Well this would explain that!

dgottlieb (approved):

PR #353, opened 2024-09-17
RSDK-8802: Add StoppableWorkers.StartTimer/NextTick.

dgottlieb:
I like the idea of having a `NextTick` for my worker. But I think there are some oddities with this so floating it around first.

We should chat about what our options are here as a next step.

### DIFF
--- a/stoppable_workers.go
+++ b/stoppable_workers.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"context"
 	"sync"
+	"time"
 )
 
 // StoppableWorkers is a collection of goroutines that can be stopped at a
@@ -14,6 +15,7 @@ type StoppableWorkers struct {
 	mu         sync.RWMutex
 	ctx        context.Context
 	cancelFunc func()
+	timer      *time.Timer
 
 	workers sync.WaitGroup
 }
@@ -66,6 +68,28 @@ func (sw *StoppableWorkers) Add(worker func(context.Context)) {
 		defer sw.workers.Done()
 		worker(sw.ctx)
 	})
+}
+
+// StartTimer is an optional call that creates a timer that can be used in conjunction with
+// `NextTick`.
+func (sw *StoppableWorkers) StartTimer(duration time.Duration) {
+	sw.timer = time.NewTimer(duration)
+}
+
+// NextTick blocks until the timer ticks or the StoppableWorkers has been canceled. It returns true
+// if the timer has ticked and false if the context is canceled. Such that one can write a loop to
+// do work as such:
+//
+//	for sw.NextTick() {
+//	  doWork()
+//	}
+func (sw *StoppableWorkers) NextTick() bool {
+	select {
+	case <-sw.timer.C:
+		return true
+	case <-sw.ctx.Done():
+		return false
+	}
 }
 
 // Stop idempotently shuts down all the goroutines we started up.
--- a/stoppable_workers.go
+++ b/stoppable_workers.go
@@ -15,7 +15,6 @@ type StoppableWorkers struct {
 	mu         sync.RWMutex
 	ctx        context.Context
 	cancelFunc func()
-	timer      *time.Timer
 
 	workers sync.WaitGroup
 }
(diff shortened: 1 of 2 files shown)


stoppable_workers.go
```diff
@@ -68,6 +70,28 @@ func (sw *StoppableWorkers) Add(worker func(context.Context)) {
...
+// NextTick blocks until the timer ticks or the StoppableWorkers has been canceled. It returns true
+// if the timer has ticked and false if the context is canceled. Such that one can write a loop to
+// do work as such:
+//
+//	for sw.NextTick() {
+//	  doWork()
+//	}
+func (sw *StoppableWorkers) NextTick() bool {
+	select {
+	case <-sw.timer.C:
```
benjirewis, line 88: I would guess `NextTick()` before `StartTimer` could panic due to accessing `.C`.

stoppable_workers.go
```diff
@@ -68,6 +70,28 @@ func (sw *StoppableWorkers) Add(worker func(context.Context)) {
...
+// `NextTick`.
+func (sw *StoppableWorkers) StartTimer(duration time.Duration) {
+	sw.timer = time.NewTimer(duration)
+}
+
+// NextTick blocks until the timer ticks or the StoppableWorkers has been canceled. It returns true
+// if the timer has ticked and false if the context is canceled. Such that one can write a loop to
+// do work as such:
+//
+//	for sw.NextTick() {
```
benjirewis, line 83: Hmm I like the ergonomics of this example. But:

You likely already alluded to this in our offline discussion, but my main concern is about multiple goroutines calling `NextTick()` concurrently. IIUC only one can `<-` a tick of the timer, so all concurrent `NextTick()` calls are subject to missing ticks `<-`ed by other goroutines. If you wanted multiple goroutines to be able to check ticks, you would need some "fan-out" pattern that would require more than one channel (or, hackily, relying on the fact that `close`ing a channel "notifies" all receivers).

Did you plan to limit usage of this to only one goroutine? If so, we'd probably need to document that and, preferably, somehow disallow calling `NextTick()` when another goroutine is already waiting for a tick. If you wanted more of "fan-out" pattern then would be happy to discuss alternate designs here.

benjirewis (commented):
Nice! Some thoughts.

stoppable_workers_test.go
```diff
@@ -1,4 +1,4 @@
-package utils_test
```
dgottlieb, line 1: Sorry -- I wrote my test calling `NewStoppableWorkerWithTicker` and saw red when getting the compile error. I didn't need to do this, but it was already done.

I don't think* there's a concern with using the non `_test` package here.

stoppable_workers.go
```diff
@@ -44,16 +45,49 @@ func NewBackgroundStoppableWorkers(workers ...func(context.Context)) *StoppableW
...
+	sw := &StoppableWorkers{ctx: ctx, cancelFunc: cancelFunc}
+	sw.workers.Add(1)
+	PanicCapturingGo(func() {
+		defer sw.workers.Done()
+
+		timer := time.NewTicker(tickRate)
+		defer timer.Stop()
+		for {
+			select {
+			case <-ctx.Done():
```
dgottlieb, line 62: I normally prefer to _not_ see this extra "select" statement. But I left it in here because otherwise I'd have to justify why my test assertion needs to check for ">= 1" instead of exactly 1.

And in the process of justifying it I felt pretty petty for avoiding writing this code.

That said, I'd still prefer to not see these. Most use cases do not need them at all. But using tickers with a small timeout that kicks off jobs is one of the exceptions.

stoppable_workers.go
dgottlieb: Talked offline -- I have no need for multiple goroutines being managed by a ticker (or set of tickers) event. @benjirewis and I decided to make a special constructor that would leverage the type system to prevent those usages.

stoppable_workers.go
dgottlieb: Correct -- also talked offline and also leveraging the type system to prevent this programmer error.

stoppable_workers.go
```diff
@@ -70,13 +104,16 @@ func (sw *StoppableWorkers) Add(worker func(context.Context)) {
 
 // Stop idempotently shuts down all the goroutines we started up.
 func (sw *StoppableWorkers) Stop() {
+	// Call `cancelFunc` with the write lock that competes with "readers" that can add workers. This
+	// guarantees `Add` worker calls that start a goroutine have incremented the `workers` WaitGroup
+	// prior to `Stop` calling `Wait`.
 	sw.mu.Lock()
-	defer sw.mu.Unlock()
```
benjirewis, line 74: @penguinland I had mentioned before that a worker could not add another worker to the group as it would result in a deadlock. That was because I wrote the `Stop` code slightly incorrectly. Dan fixed it here by making sure `Wait` is not called while holding the mutex.

stoppable_workers.go
benjirewis: I buy that this is an exception. I agree that this pattern is generally confusing and usually unnecessary, though.

benjirewis (approved):
Looks good to me! Thanks for going the constructor route.

stoppable_workers.go
```diff
@@ -70,13 +104,16 @@ func (sw *StoppableWorkers) Add(worker func(context.Context)) {
...
 	if sw.ctx.Err() != nil {
 		return
```
penguinland, line 112: If we hit this return statement, we never unlock the mutex, and the program could deadlock. I think you're right that in this specific function, it's better to manually unlock the mutex and not use a `defer` statement, but usually I prefer using `defer` because it makes these things much easier to reason about.

benjirewis: Oof yeah @dgottlieb I think we're missing an `Unlock` here. You're definitely right that `defer` normally handles this stuff, but we really wanted that `Wait` to be outside of a lock. Thanks for taking a look post-hoc @penguinland !

penguinland: Linking things up: I made https://github.com/viamrobotics/goutils/pull/361 to fix this.

PR #354, opened 2024-09-17
Migrate to connect-es

edaniels:
Hey all, we talked about doing this migration in #frontend-friends. I figured it would be best to review this code in this repo first, get it merged, and then move it over to the TS SDK. If we somehow need some patches in the meantime, we can just branch off (or worst case revert).

Connect supports pretty much everything we need (node, react native, browser, and more). We can continue the model of having a `global.VIAM` for setting transports and other platform specific needs.

Before:
```shell
✓ 51 modules transformed.
dist/main.js  663.78 kB │ gzip: 87.59 kB
✓ built in 678ms
```

After:
```shell
✓ 133 modules transformed.
dist/main.js  130.79 kB │ gzip: 31.52 kB
✓ built in 356ms
```


rpc/examples/echo/node-client/src/index.ts
```diff
@@ -0,0 +1,108 @@
+import type { PartialMessage } from "@bufbuild/protobuf";
+import { ConnectError, createPromiseClient, PromiseClient } from "@connectrpc/connect";
+import { createWritableIterable } from "@connectrpc/connect/protocol";
+import { dialDirect, dialWebRTC } from "@viamrobotics/rpc";
+import { EchoService } from "./gen/proto/rpc/examples/echo/v1/echo_connect.js";
+import { EchoBiDiRequest, EchoMultipleRequest, EchoRequest } from "./gen/proto/rpc/examples/echo/v1/echo_pb.js";
+
+import { createGrpcTransport } from '@connectrpc/connect-node';
```
edaniels, line 8: @njooma this has webrtc "just work" but as far as the TS SDK, I feel like this is kind of ugly and warrants its own node sdk package, unless viam is okay with everyone needing to do this in node. there could be a helper that we may be able to get away with using dynamic imports (that you await)

proto/rpc/examples/fileupload/v1/fileupload.proto
```diff
@@ -16,7 +16,5 @@ message UploadFileResponse {
 }
 
 service FileUploadService {
-  // Due to an issue described by https://github.com/improbable-eng/ts-protoc-gen/pull/264
-  // we use a streaming response but only expect one response.
-  rpc UploadFile(stream UploadFileRequest) returns (stream UploadFileResponse) {}
+  rpc UploadFile(stream UploadFileRequest) returns (UploadFileResponse) {}
```
edaniels, line 19: Yeah so this was actually wrong and it was never bidi

buf.lock
```diff
@@ -4,4 +4,4 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: 62f35d8aed1149c291d606d958a7ce32
```
edaniels, line 7: you'll see some buf/goog changes to get us to latest build artifacts/tools

rpc/examples/echo/frontend/src/index.ts
```diff
@@ -57,101 +60,90 @@ async function getClients() {
...
+		// grpc-web cannot do bidi correctly. Previous versions
+		// of this code just stalled after the first message was received.
+		return;
+	}
 
-	done = new Promise<any>((resolve, reject) => {
-		pResolve = resolve;
-		pReject = reject;
-	});
+	const clientStream = createWritableIterable<PartialMessage<EchoBiDiRequest>>();
```
edaniels, line 113: Look around here for how to use connect-es. IMO, a lot more ergonomic. Even the bidi case isn't that nasty

rpc/js/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "module": "ESNext",
     "moduleResolution": "node",
     "verbatimModuleSyntax": false,
-    "lib": ["ESNext", "DOM", "DOM.iterable"]
+    "lib": ["ESNext", "DOM", "DOM.iterable"],
+    "exactOptionalPropertyTypes": false,
```
edaniels, line 11: needed for connect rpc passing around optionals

rpc/examples/echo/node-client/src/index.ts
maximpertsov: no opinion on whether this should be it's own SDK package or not, but it's nice that node is supported in _some_ way with this change

rpc/js/src/ClientChannel.ts
```diff
@@ -38,41 +50,32 @@ export class ClientChannel extends BaseChannel {
...
+    const { stream } = resp;
     if (stream === undefined) {
       console.error('no stream id; discarding');
       return;
     }
 
-    const id = stream.getId();
-    const activeStream = this.streams[id];
+    const { id } = stream;
+    const activeStream = this.streams[Number(id)];
```
maximpertsov, line 78:
```suggestion
    const activeStream = this.streams[id];
```

Not sure how likely we are to hit a stream count where this would matter, but maybe we should just use `BigInt` throughout? My understanding is that coercing between `BigInt` and `Number` could lead to precision loss per [MDN docs]( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#description):

> Be careful coercing values back and forth, however, as the precision of a BigInt value may be lost when it is coerced to a Number value.

maximpertsov (commented):
This is awesome!

I have [one minor comment](https://github.com/viamrobotics/goutils/pull/354/files#r1764946233) related to BigInt/Number coercion. Otherwise LGTM.

rpc/js/src/ClientChannel.ts
edaniels: The proto changed this to a bigint so we could also just change the property on the map?

maximpertsov: > change the property on the map

yep let's do that

edaniels: so bigint cannot be used as an index which makes sense, so how about the record key be a string and we go bigint->string on demand? it has a cost but I don't think we're in that realm of performance needs just yet

maximpertsov: that works too

maximpertsov (approved):

mcous (commented):
Very excited about this work, it's a huge piece of nagging tech debt off our shoulders!

I've done an initial review pass to highlight some areas of confusion for me that I'd like to focus on in our in-person review tomorrow. Also tried to highlight areas where I think extra cleanup, testing, and/or comments would be helpful, either in this PR or in the future, to ensure we're going to be able to maintain this moving forward.

If it can be practically done for this PR, we've been doing a lot of work in Fleet and Viz to use PR/commit stacking to break up large bits of work into smaller chunks. If you're down, I think splitting the updates to `examples` into their own PRs might make this easier to review, understand, and land. I also wonder if there's a way we could really minimize the examples changes to get a better sense of just how much of a "breaking" change this is - as opposed to opportunistic cleanup of the examples. (I think example cleanup is valuable but could happen in follow-up PRs by other contributors to gain familiarity.)

rpc/js/src/dial.ts
```diff
@@ -339,15 +353,15 @@ export async function dialWebRTC(
   // TODO(RSDK-2836): In general, this logic should be in parity with the golang implementation.
   // https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_client.go#L160-L175
   const config = await getOptionalWebRTCConfig(signalingAddress, host, opts);
```
ethanlook, line 355: Looks like we're doing this to get additional ice servers from the machine. Why is this necessary?

rpc/js/src/errors.ts
ethanlook, line 1: This error is used throughout the TS SDK - what's the rationale for removing it?

rpc/js/tsconfig.json
edaniels: I'm gonna try but not hopeful :)

rpc/js/src/errors.ts
edaniels: It wasn't used in this package but I'll move it back

rpc/js/tsconfig.json
edaniels: figured it out but it was nasty and i love it

rpc/js/src/dial.ts
```diff
@@ -90,180 +96,185 @@ export interface Credentials {
...
-    } catch {
-      TransFact = grpc.CrossBrowserHttpTransport;
-    }
-    return TransFact({ withCredentials: false })(opts);
+  let transFact: TransportFactory;
+  try {
+    transFact = window.VIAM.GRPC_TRANSPORT_FACTORY;
+  } catch {
+    transFact = createGrpcWebTransport;
+  }
```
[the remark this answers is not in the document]
edaniels: yep this is better

rpc/js/src/dial.ts
```diff
@@ -412,64 +400,58 @@ export async function dialWebRTC(
...
+      };
+      try {
+        await signalingClient.callUpdate(callRequestUpdate, {
+          headers: {
+            'rpc-host': host,
+          },
+        });
+      } catch (err) {
+        console.error(err);
+      }
```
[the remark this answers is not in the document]
edaniels: left a comment on it explaining why it's okay

rpc/examples/echo/frontend/src/index.ts
```diff
@@ -17,15 +20,15 @@ declare global {
...
-  const selector = `[data-testid="${type}-${method}"]`;
-  const elem = document.querySelector(selector);
-  if (!elem) {
-    throw new Error(`expecting to find selector '${selector}'`);
-  }
-  const inner = document.createElement("div");
-  inner.setAttribute("data-testid", "message");
-  inner.innerText = text;
-  elem.appendChild(inner);
+	const selector = `[data-testid="${type}-${method}"]`;
```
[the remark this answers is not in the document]
edaniels: this file is actually inconsistent on spacing/tabs so I think I'm going to leave this unless you really care

rpc/examples/echo/frontend/src/index.ts
```diff
@@ -57,101 +60,90 @@ async function getClients() {
...
-	const webRTCConn = await dialWebRTC(thisHost, webrtcHost, opts);
-	const webrtcClient = new EchoServiceClient(webrtcHost, { transport: webRTCConn.transportFactory });
-	await renderResponses(webrtcClient, "wrtc");
+	try {
+		const webRTCConn = await dialWebRTC(thisHost, webrtcHost, opts);
+		const webrtcClient = createPromiseClient(EchoService, webRTCConn.transport);
+		await renderResponses(webrtcClient, "wrtc");
+	} catch (err) {
+		console.error("error trying WebRTC", err);
+	}
```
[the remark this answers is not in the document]
edaniels: yep

rpc/examples/echo/node-client/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+import wrtc from "node-datachannel/polyfill";
+global.window = global.window || {};
+window.VIAM = {
+	GRPC_TRANSPORT_FACTORY: (opts: any) => createGrpcTransport({ httpVersion: "2", ...opts }),
+};
```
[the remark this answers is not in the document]
edaniels: it's more for the isomorphism. there's likely a better pattern but I don't want to spend the cycles on that right now since it's for lesser used use cases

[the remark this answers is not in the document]
edaniels: will do globalThis

rpc/js/src/ClientChannel.ts
```diff
@@ -38,41 +50,32 @@ export class ClientChannel extends BaseChannel {
...
     let resp: Response;
     try {
-      resp = Response.deserializeBinary(
-        new Uint8Array(event.data as ArrayBuffer)
-      );
+      resp = Response.fromBinary(new Uint8Array(event.data as ArrayBuffer));
     } catch (e) {
       console.error('error deserializing message', e);
       return;
     }
```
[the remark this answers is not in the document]
edaniels: sketchy indeed. will remove

rpc/examples/echo/frontend/src/index.ts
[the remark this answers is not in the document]
mcous: Lol of course. No, happy to let prettier deal with this later

rpc/js/src/ClientStream.ts
```diff
@@ -18,77 +38,285 @@ import {
...
+        stream: false,
+        url: '',
+        init: {},
+        service,
+        method,
+        header,
+        contextValues: contextValues ?? createContextValues(),
+        message,
+      },
+      next: async (req: UnaryRequest<I, O>): Promise<UnaryResponse<I, O>> => {
```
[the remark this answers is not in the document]
edaniels: refactor incoming

rpc/js/src/ClientChannel.ts
```diff
@@ -82,81 +85,145 @@ export class ClientChannel extends BaseChannel {
 
   private nextStreamID(): Stream {
     const stream = new Stream();
```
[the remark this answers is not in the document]
edaniels: did this everywhere as much as I could

rpc/js/src/ClientStream.ts
```diff
@@ -98,158 +326,134 @@ export class ClientStream extends BaseStream implements grpc.Transport {
...
-    md.set(key, strings);
+    strings.values = [value];
+    md[key] = strings;
   });
-  if (result.getMdMap().getLength() === 0) {
+  if (Object.keys(result.md).length === 0) {
     return undefined;
   }
   return result;
 };
```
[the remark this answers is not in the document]
edaniels: not adding tests here for both but put in your nice changes

rpc/js/src/dial.ts
```diff
@@ -90,180 +96,211 @@ export interface Credentials {
...
+  >(
+    service: ServiceType,
+    method: MethodInfo<I, O>,
+    signal: AbortSignal | undefined,
+    timeoutMs: number | undefined,
+    header: HeadersInit | undefined,
+    message: PartialMessage<I>,
+    contextValues?: ContextValues
+  ): Promise<UnaryResponse<I, O>> {
+    let newHeaders = cloneHeaders(header);
```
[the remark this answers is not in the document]
edaniels: I'll keep a lookout but also once we start linting

rpc/js/src/dial.ts
```diff
@@ -664,13 +591,78 @@ export async function dialWebRTC(
...
-      .catch((err) => clientEndReject(err));
-    await clientEnd;
-    await cc.ready;
+    const processCallResponses = async () => {
+      try {
+        for await (const response of callResponses) {
+          if (response.stage.case == 'init') {
+            if (haveInit) {
+              await sendError('got init stage more than once');
+              continue;
```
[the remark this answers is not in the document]
edaniels: fixed

rpc/js/src/dial.ts
```diff
@@ -664,13 +591,78 @@ export async function dialWebRTC(
       }, opts.dialTimeout);
     }
 
-    cc.ready
-      .then(() => clientEndResolve())
-      .catch((err) => clientEndReject(err));
-    await clientEnd;
-    await cc.ready;
+    const processCallResponses = async () => {
+      try {
```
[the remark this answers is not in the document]
edaniels: will address in refactor

rpc/js/src/dial.ts
edaniels: added a comment!

edaniels:
did a bunch of cleaning. next step is a signaling refactor and a comments pass if needed. then ready for re-review. @mcous @ethanlook feel free to take a look at the ClientStream/UnaryClientStream/StreamClientStream refactor. I think it's a lot easier to read now and it minimizes the complexity of the new Promise logic as minimally as I could.

edaniels:
@ethanlook @mcous ready for another review. since the last comment, I refactored signaling as requested. It's still complicated, but a little easier to read. Also split the commits up.

rpc/dialer.go
```diff
@@ -107,6 +107,7 @@ func (cd *cachedDialer) DialDirect(
 	opts ...grpc.DialOption,
 ) (ClientConn, bool, error) {
 	return cd.DialFunc("grpc", target, keyExtra, func() (ClientConn, func() error, error) {
+		//nolint:staticcheck
```
edaniels, line 110: @dgottlieb heads up for rdk and other go users of grpc that grpc is removing DialContext in favor of NewClient + Connect. I'm just ignoring these warnings here for now, but these may pop up in rdk.

edaniels:
Also made some changes to the buf yaml to reduce how many files this PR changes. There was an unneeded version upgrade.

rpc/js/src/BaseChannel.ts
```diff
@@ -1,4 +1,4 @@
-import type { ProtobufMessage } from '@improbable-eng/grpc-web/dist/typings/message';
+import { Message } from '@bufbuild/protobuf';
```
mcous, line 1: Import only used as type

```suggestion
import type { Message } from '@bufbuild/protobuf';
```

rpc/js/src/ClientStream.ts
```diff
@@ -1,7 +1,13 @@
-import { grpc } from '@improbable-eng/grpc-web';
+import {
```
mcous, line 1:
```suggestion
import type {
```

rpc/js/src/ClientChannel.ts
```diff
@@ -1,6 +1,18 @@
-import type { grpc } from '@improbable-eng/grpc-web';
+import {
```
mcous, line 1:
```suggestion
import type {
```

rpc/js/src/ClientChannel.ts
```diff
@@ -1,6 +1,18 @@
...
+  AnyMessage,
+  Message,
+  MethodInfo,
+  PartialMessage,
+  ServiceType,
+} from '@bufbuild/protobuf';
+import {
```
mcous, line 8:
```suggestion
import type {
```

rpc/js/src/ClientChannel.ts
```diff
@@ -1,6 +1,18 @@
...
+  ContextValues,
+  StreamResponse,
+  Transport,
+  UnaryResponse,
+} from '@connectrpc/connect';
 import { BaseChannel } from './BaseChannel';
-import { ClientStream } from './ClientStream';
+import { ClientStream, ClientStreamConstructor } from './ClientStream';
```
mcous, line 15:
```suggestion
import type { ClientStream, ClientStreamConstructor } from './ClientStream';
```

rpc/js/src/ClientChannel.ts
```diff
@@ -38,41 +52,26 @@ export class ClientChannel extends BaseChannel {
...
-    let resp: Response;
-    try {
-      resp = Response.deserializeBinary(
-        new Uint8Array(event.data as ArrayBuffer)
-      );
-    } catch (e) {
-      console.error('error deserializing message', e);
-      return;
-    }
+    let resp = Response.fromBinary(new Uint8Array(event.data as ArrayBuffer));
```
mcous, line 65: Do we need a runtime check or assert for this type cast to `as ArrayBuffer`? Or should `MessageEvent<any>` be defined more strictly?

rpc/js/src/ClientChannel.ts
```diff
@@ -38,41 +52,26 @@ export class ClientChannel extends BaseChannel {
...
     this.closeWithReason(new ConnectionClosedError('data channel closed'));
-    const err = new ConnectionClosedError('connection terminated');
     for (const streamId in this.streams) {
       const stream = this.streams[streamId]!;
-      stream.cs.closeWithRecvError(err);
+      stream.cs.closeWithRecvError();
     }
   }
 
   private onChannelMessage(event: MessageEvent<any>) {
```
mcous, line 64: Should this be `MessageEvent<I>` (or `MessageEvent<O>` if I'm misunderstanding the terminology here) rather than `any`?

rpc/js/src/main.ts
```diff
@@ -1,7 +1,10 @@
+import { Transport } from '@connectrpc/connect';
```
mcous, line 1:
```suggestion
import type { Transport } from '@connectrpc/connect';
```

rpc/js/src/main.ts
```diff
@@ -1,7 +1,10 @@
...
+
 declare global {
-  interface Window {
-    VIAM: any;
-  }
+  // eslint-disable-next-line vars-on-top,no-var
+  var VIAM: {
+    GRPC_TRANSPORT_FACTORY: (opts: any) => Transport;
+  };
```
mcous, line 7: Ensure these things are typed as potentially not existing

```suggestion
  var VIAM:
    | {
        GRPC_TRANSPORT_FACTORY?: (opts: any) => Transport;
      }
    | undefined;
```

rpc/examples/echo/node-client/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+import { EchoService } from "./gen/proto/rpc/examples/echo/v1/echo_connect.js";
+import { EchoBiDiRequest, EchoMultipleRequest, EchoRequest } from "./gen/proto/rpc/examples/echo/v1/echo_pb.js";
+
+import { createGrpcTransport } from '@connectrpc/connect-node';
+import wrtc from "node-datachannel/polyfill";
+declare global {
+	var VIAM: {
+		GRPC_TRANSPORT_FACTORY: (opts: any) => Transport;
+	};
+}
```
mcous, line 14: Not needed, since it's declared in `main.ts` of `@viamrobotics/rpc`
```suggestion

```

rpc/js/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "@viamrobotics/typescript-config/base.json",
   "compilerOptions": {
-    "target": "ES5",
+    "target": "esnext",
     "useDefineForClassFields": true,
     "importHelpers": true,
```
mcous, line 6: Some suggested TSConfig changes:

- Remove `useDefineForClassFields` and `importHelpers` as unnecessary
- Add `"noEmit": true`
    - Can then remove `--no-emit` from `scripts.typecheck` in `package.json`
- Remove `"verbatimModuleSyntax": false` to use our base default of `true`
    - This will catch all the `import type { ... };` problems, a few of which I pointed out
- Add `"noEmit": false` to `tsconfig.build.json` to keep the `.d.ts` file generation working

mcous (commented):
Still working through this, but here's a quick batch of housekeeping comments just to keep things moving

rpc/js/src/BaseChannel.ts
edaniels: shouldn't all these kinds of things be picked up by a linter?

mcous: Yup, see https://github.com/viamrobotics/goutils/pull/354#discussion_r1771674359

rpc/js/tsconfig.json
edaniels: didn't seem to catch `import { Message } from '@bufbuild/protobuf';` btw

rpc/js/src/ClientChannel.ts
edaniels: it's neither. this is a message coming from a data channel which will be bytes

rpc/js/tsconfig.json
mcous: Yeah, I've seen it miss some stuff in the app, too. As long as we get the bulk of them that's good.

I think `vite` throws some build warnings about unused imports, too (haven't looking into how to promote those to errors, yet) and I suspect newer versions of `@typescript-eslint` might do a better job of detecting this, too, so hopefully we can get full coverage once this is over in the TS SDK repo

rpc/js/src/ClientChannel.ts
edaniels: typed it more

rpc/js/src/signaling-exchange.ts
```diff
@@ -0,0 +1,391 @@
...
+      const callUpdateDuration =
+        callUpdateEnd.getTime() - callUpdateStart.getTime();
+      if (callUpdateDuration > this.maxCallUpdateDuration) {
+        this.maxCallUpdateDuration = callUpdateDuration;
+      }
+      this.totalCallUpdateDuration += callUpdateDuration;
+    } catch (error) {
+      if (
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        this.exchangeDone ||
```
mcous, line 289: [Not a blocker, not the best time to resolve] Same comment here: it looks a lot like `this.exchangeDone` is already checked above, and I don't immediately see any method calls that would change `exchangeDone` between the two checks. Is something coming in externally and causing it to change?

Either way, I think if it's enough to confuse ESLint, it's enough to confuse a future reader like me.

mcous (approved):
I think this is a great state for a testing release! Appreciate all your work on this

rpc/js/src/signaling-exchange.ts
edaniels: An await is used prior to exchange done bring checked which means based on how the event loop orders things, it may be set to true from another function. This is a case of unseen concurrency in JavaScript

rpc/js/src/dial.ts
```diff
@@ -90,641 +97,425 @@ export interface Credentials {
...
+        const transport = defaultFactory({ baseUrl: resolvedAddress });
+        const authClient = createPromiseClient(AuthService, transport);
+        const resp = await authClient.authenticate(request);
+        thisAccessToken = resp.accessToken;
       } else {
         thisAccessToken = opts.accessToken;
       }
 
-      accessToken = thisAccessToken;
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- await race
```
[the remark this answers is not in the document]
edaniels: Same response as the other due to await

rpc/js/src/signaling-exchange.ts
edaniels: Without this extra check, you'll see a needs atomic check lint and the two resolutions to the lints are in conflict. I've seen this several times for concurrent code in JS and it's frustrating to work around.

rpc/js/src/SignalingExchange.ts
ethanlook, line 1: Nice, this encapsulation is really helpful for understanding the signaling exchange and how it fits in the bigger picture of dialing. Much appreciated!

ethanlook (approved):
Looks great, thanks for the lint cleanup as well! And thank you again for walking us through this code.

rpc/js/src/signaling-exchange.ts
edaniels: managed to simplify the code since there's no more factories here really

rpc/js/src/dial.ts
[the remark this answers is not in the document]
edaniels: managed to simplify the code since there's no more factories here really

PR #356, opened 2024-09-23
RSDK-8566 Send gRPC heartbeats from signaling server to answerer

benjirewis:
https://viam.atlassian.net/browse/RSDK-8566

Sends heartbeats from signaling server to signaling answerer.

(cc @npmenard)


rpc/wrtc_signaling_answerer.go
```diff
@@ -212,9 +213,21 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
...
 			}
 
-			// `client.Recv` waits, typically for a long time, for a caller to show up. Which is
-			// when the signaling server will send a response saying someone wants to connect.
-			incomingCallerReq, err := client.Recv()
+			var incomingCallerReq *webrtcpb.AnswerRequest
+			for {
+				// `client.Recv` waits, typically for a long time, for a caller to show up. Which is
+				// when the signaling server will send a response saying someone wants to connect. It
+				// can also receive heartbeats every 15s. Discard heartbeats.
```
dgottlieb, line 220: What's the useful side-effect of having the signaling server send heartbeats if we're just going to discard them?

I'm guessing the act of the signaling server calling `SendMsg(heartbeat)` is what can now trigger an error that the server handles by cleaning up server-side resources? Or maybe we expect the server's internal gRPC logic handles that error in a way that makes our problem go away?

Whatever it is, can we document that? I think someone seeing this will see these heartbeats are intentional, but not know how to measure if it's having the intended effect. I think it's fine if we don't exactly know the magic here, but leaving as much detail as we do know can be useful.

I can see a developer coming across this and wanting to send a "heartbeat" in response, as that's the usual pattern. Documentation can help that developer understand if there would be an observable impact of such a change.

rpc/wrtc_signaling_server.go
```diff
@@ -331,6 +349,28 @@ func (srv *WebRTCSignalingServer) Answer(server webrtcpb.SignalingService_Answer
...
+	// heartbeats.
+	if HeartbeatsAllowedFromCtx(ctx) {
+		utils.PanicCapturingGo(func() {
+			for {
+				select {
+				case <-time.After(heartbeatInterval):
+					if err := server.Send(&webrtcpb.AnswerRequest{
+						Stage: &webrtcpb.AnswerRequest_Heartbeat{},
+					}); err != nil {
+						srv.logger.Debugw(
```
dgottlieb, line 362: Similar remark to the other comment re: side-effects. I buy these changes fix the long delays before robots can be connected to again. But because we just ignore the errors, it's going to be hard for a reader to pick up our expectation of what's changing under the hood. I know our information here is limited, but sometimes it's important to communicate what we don't know just as much as it is communicating what we do know.

dgottlieb:
Code changes look perfectly fine to me. Just want to leave as much context in the form of code documentation on how we think these messages help for future readers.

rpc/wrtc_signaling_answerer.go
benjirewis: > I'm guessing the act of the signaling server calling SendMsg(heartbeat) is what can now trigger an error that the server handles by cleaning up server-side resources?

This is my understanding: it's the error from heartbeating server-side that's really helping out here. I'll document a bit better.

rpc/wrtc_signaling_server.go
```diff
@@ -331,6 +349,34 @@ func (srv *WebRTCSignalingServer) Answer(server webrtcpb.SignalingService_Answer
 	}
 	defer srv.clearAdditionalICEServers(hosts)
 
+	// If heartbeats allowed (indicated by answerer), start goroutine to send
+	// heartbeats.
+	//
+	// The answerer does not respond to heartbeats. The signaling server is only
+	// using heartbeats to ensure the answerer is reachable. If the answerer is
+	// down, the heartbeat will error in the heartbeating goroutine below, the
+	// stream's context will be canceled, and we will stop handling interactions
```
dgottlieb, line 358: Does "the stream's context will be canceled, and we will stop handling interactions" refer to the `ctx` used in line 380 `RecvOffer(ctx, hosts)`?

benjirewis: It is indeed that context. More specifically, that `ctx` value is really `server.Context()` which is the context stored on the bidi stream passed into `Answer`.

dgottlieb: Ok, this makes a lot more sense, thanks for that explanation! Sorry for one (last?) ask: can we incorporate that information into the documentation? The context handling here deviates from common patterns (I think out of necessity), so it's not obvious that the line 365 `server.Send` getting an error implies the context used on line 380 for a blocking operation gets canceled. Knowing this, I can see your current documentation touching all the important pieces. I just think some more hand-holding/babying of the reader in this specific case will be helpful.

benjirewis: For sure! Thanks for asking for more clarity here; some of this stuff is opaque, so I agree that it's important to document what's going on here. I added
```
// .... We stop handling interactions because the stream's                          
// context (`ctx` here and below) is used in the `RecvOffer` call below this                                               
// goroutine that waits for a caller to attempt to establish a connection.  
```

I only added that piece to the signaling server side, as I think that concept might be more difficult to explain from the answerer side, but hopefully future NetCoders will be able to see the relationship between the server and answerer heartbeat logic.

dgottlieb: > I only added that piece to the signaling server side

For sure. In hindsight, now knowing where the action is happening, I would have better placed my questions. I think duplicating some of the content as documentation is perfectly fine though.

dgottlieb (approved):
I know you mentioned tests, but approving for now. Happy to look again after tests are added if there's something interesting.

maximpertsov (commented):
approach makes sense per the context you gave IRL - @dgottlieb covered most of my questions and I think the rationale behind this change is well-captured in the code comments

rpc/wrtc_signaling_test.go
```diff
@@ -199,3 +199,57 @@ func TestWebRTCAnswererImmediateStop(t *testing.T) {
...
+	// Assert that the answerer eventually logs received heartbeats.
+	testutils.WaitForAssertion(t, func(tb testing.TB) {
+		t.Helper()
+		test.That(tb, observer.FilterMessageSnippet(heartbeatReceivedLog).Len(),
+			test.ShouldBeGreaterThan, 0)
+	})
+
+	webrtcServer.Stop()
+	answerer.Stop()
+	grpcServer.Stop()
```
maximpertsov, line 253: [q]: does the order in which these stop matter?

maximpertsov (approved):
looks good! [one q](https://github.com/viamrobotics/goutils/pull/356/files#r1786665460) mostly out of curiosity if you happen to know.

rpc/wrtc_signaling_test.go
benjirewis: A great q; I copied from another test above. Does the ordering matter? From what I can tell: sort of (I think there _could_ be unexpected errors from sig server/answerer or machine if the ordering is particularly bad). I've changed the order here and in the test I copied to be the same order as the one in `simpleServer.Stop` to mimic what happens in prod. Left a comment, too.

dgottlieb (approved):

rpc/wrtc_signaling_server.go
```diff
@@ -361,6 +361,8 @@ func (srv *WebRTCSignalingServer) Answer(server webrtcpb.SignalingService_Answer
 	// goroutine that waits for a caller to attempt to establish a connection.
 	if HeartbeatsAllowedFromCtx(ctx) {
 		utils.PanicCapturingGo(func() {
+			// Capture as tests can mutate this value.
+			heartbeatInterval := heartbeatInterval
```
dgottlieb, line 365: Err, this doesn't fix the data race

benjirewis: Totally correct; we discussed offline. Apologies for the errant fix there 😮‍💨 . Passed `heartbeatInterval` to the signaling server constructor and used a `const defaultHeartbeatInterval` value instead.

dgottlieb (approved):

PR #357, opened 2024-09-24
SEC-39: Check if JWT is associated  with session

jr22:


web/sessions.go
```diff
@@ -216,6 +236,34 @@ func (mss *mongoDBSessionStore) Get(ctx context.Context, id string) (*Session, e
 	return s, nil
 }
 
+func (mss *mongoDBSessionStore) GetByToken(ctx context.Context, token string) (*Session, error) {
```
mcous, line 239: Discussed in person, consider switching to an exists query since we don't use the `Session` from this method in the added code

web/sessions.go
```diff
@@ -216,6 +238,21 @@ func (mss *mongoDBSessionStore) Get(ctx context.Context, id string) (*Session, e
...
 
+func (mss *mongoDBSessionStore) HasSessionWithToken(ctx context.Context, token string) (bool, error) {
+	ctx, span := trace.StartSpan(ctx, "MongoDBSessionStore::Get")
+	defer span.End()
+
+	count, err := mss.collection.CountDocuments(ctx, bson.M{accessTokenFieldPath: token}, options.Count().SetLimit(1))
+	if err != nil {
+		if errors.Is(err, mongo.ErrNoDocuments) {
+			return false, errNoSession
+		}
```
mcous, line 249: Unless I'm misunderstanding what `mongo.ErrNoDocuments` means, should we consider this an expected case and avoid returning the error? The `error` level log up top seems a little aggressive for the normal `count == 0` case

mcous (approved):

web/sessions.go
jr22: oh yeah this doenst make much sense with the new function name. i actually dont think i need this check now that i moved it from FindOne. count return this error. will remove!

mcous (approved):
LGTM pending lint complaint

PR #360, opened 2024-09-24
RSDK-8731 - hardcode loopback interface to listen on for mDNS

cheukt:
Have not run into a case where adding loopback to the list of interfaces causes errors.

The only time I found errors was if the mDNS client failed to listen on a socket for all interfaces passed in, but if any interface can listen to that socket, it doesn't error

manually tested with wlan up and down


rpc/dial.go
```diff
@@ -295,11 +298,38 @@ func dial(
...
 }
 
+func listMulticastInterfaces() []net.Interface {
+	var interfaces []net.Interface
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return nil
+	}
+	for _, ifi := range ifaces {
+		if (ifi.Flags & net.FlagUp) == 0 {
```
benjirewis, line 308: What is this checking for?

cheukt: if the interface is not up - the way flags are parsed and implemented is definitely a bit confusing

benjirewis: Makes sense; mind leaving a comment above this line stating that?

benjirewis (approved):
LGTM % that comment about the flag logical and-ing.

PR #361, opened 2024-09-26
Bugfix: unlock mutex before returning

penguinland:
See discussion at https://github.com/viamrobotics/goutils/pull/353/files#r1777286836. I also put in a comment on a part I wished had a comment in it. :wink: Without it, my first instinct is to go back to using `defer sw.mu.Unlock()` immediately after locking it.

### DIFF
--- a/stoppable_workers.go
+++ b/stoppable_workers.go
@@ -113,6 +113,9 @@ func (sw *StoppableWorkers) Stop() {
 		return
 	}
 	sw.cancelFunc()
+	// Make sure to unlock the mutex before waiting for background goroutines to shut down! That
+	// way, any goroutine that was waiting on this lock (e.g., it was trying to spawn another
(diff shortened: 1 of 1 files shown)


benjirewis (approved):
LGTM but what's up with the "Unchanged files with check annotations" section? I'm assuming that's not part of your changes?

dgottlieb:
Thanks for spotting that and I agree on the utility of the comment!

penguinland:
Benji, that's a super good question! You're right; it's not part of my change. I can find lots of people online complaining about how distracting it is, going back to 2019 (so, this isn't exactly new to github, despite the "beta" label), but haven't yet seen a clear description of where it's configured.

benjirewis:
Gotcha; that's what I saw, too, from a quick google. I think it's fine to merge.

PR #364, opened 2024-10-03
Update token endpoint for better error handling

DTCurrie:
Update the token endpoint to read the existing bearer token (if it exists) and better handle situations a user may run into. 

- User attempts to get a token with no existing bearer token and no token cookies present:
  - Return a 400
- User attempts to get a token with an existing bearer token and no token cookies present:
  - If the bearer token is valid, do nothing and return a 204 (the user is in an active session)
  - If the bearer token is invalid, return a 401 (the user is in an inactive session)

Otherwise, we will continue with the existing behavior.

Left comments in to explain the flow, can delete/simplify them before merging.


mcous (approved):
LGTM

web/auth.go
```diff
@@ -413,25 +413,68 @@ type tokenResponse struct {
...
-		return
+	if err != nil || token.Value == "" {
+		failedToGetCookies = true
 	}
 
 	expiry, err := r.Cookie("viam.auth.expiry")
-	if HandleError(w, err, h.logger, "getting expiry cookie") {
+	if err != nil || token.Value == "" {
+		failedToGetCookies = true
+	}
```
jr22, line 453: are these supposed to be checking `token.Value` from `r.Cookie("viam.auth.token")` or is it supposed to be checking `current`?

jr22: if its the former, can we get rid of it in the second two checks since it will always be true if its true the first time?

DTCurrie: Nope, just some copy paste errors, will resolve them.

jr22 (approved):

web/auth.go
```diff
@@ -413,25 +413,69 @@ type tokenResponse struct {
...
+
+	// user calls with a token in the header, no cookies exist
+	// - check if the token in the header is valid
+	//   - if so:
+	//     - do nothing, return 204
+	//   - if not:
+	//     - return an unauthenticated error 401
+	//     - force user logout
+	if failedToGetCookies && current != "" {
+		isValid := h.state.sessions.HasSessionWithAccessToken(ctx, current)
```
DTCurrie, line 471: @jr22 Ok so when testing, I noticed I could logout, and use my old token from the logged out session to get myself a new token and start an active session. Just want to check, I would assume if the value of `current` is a token that was logged out, then this would return `false`, but it appears that may not be happening given my testing on @mcous 's [PR](https://github.com/viamrobotics/app/pull/6392). Is my assumption correct?

jr22: hm yeah i wouldnt think that should work

DTCurrie:
@jr22 I have been fighting this linter for way too long. For some reason, I can't run `make lint` on my local. Seeing these warnings before the process terminates itself:

```bash
WARN [linters_context] running gomodguard failed: unable to parse module file go.mod: go.mod:3: invalid go version '1.21.13': must match format 1.23: if you are not using go modules it is suggested to disable this linter 
WARN [linters_context] running gomoddirectives failed: failed to get module file: go.mod:3: invalid go version '1.21.13': must match format 1.23: if you are not using go modules it is suggested to disable this linter 
```

It seems to run fine in CI. I just wiped my local repo, re-cloned, and tried again, and I am seeing the same issues. There are linting issues, but I can't parse what the problem is.

web/auth.go
jr22: did you fix this/ figure out what was going on?

jr22:
> @jr22 I have been fighting this linter for way too long. For some reason, I can't run `make lint` on my local. Seeing these warnings before the process terminates itself:
> 
> ```shell
> WARN [linters_context] running gomodguard failed: unable to parse module file go.mod: go.mod:3: invalid go version '1.21.13': must match format 1.23: if you are not using go modules it is suggested to disable this linter 
> WARN [linters_context] running gomoddirectives failed: failed to get module file: go.mod:3: invalid go version '1.21.13': must match format 1.23: if you are not using go modules it is suggested to disable this linter 
> ```
> 
> It seems to run fine in CI. I just wiped my local repo, re-cloned, and tried again, and I am seeing the same issues. There are linting issues, but I can't parse what the problem is.

might be because the `go.mod` for this repo is `1.21.13` but your local version is 1.23. you might have to downgrade the version of go youre running to get this to work (app is on go 1.23 so would make sense that locally youre running tht)

DTCurrie:
> > @jr22 I have been fighting this linter for way too long. For some reason, I can't run `make lint` on my local. Seeing these warnings before the process terminates itself:
> > ```shell
> > WARN [linters_context] running gomodguard failed: unable to parse module file go.mod: go.mod:3: invalid go version '1.21.13': must match format 1.23: if you are not using go modules it is suggested to disable this linter 
> > WARN [linters_context] running gomoddirectives failed: failed to get module file: go.mod:3: invalid go version '1.21.13': must match format 1.23: if you are not using go modules it is suggested to disable this linter 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It seems to run fine in CI. I just wiped my local repo, re-cloned, and tried again, and I am seeing the same issues. There are linting issues, but I can't parse what the problem is.
> 
> might be because the `go.mod` for this repo is `1.21.13` but your local version is 1.23. you might have to downgrade the version of go youre running to get this to work (app is on go 1.23 so would make sense that locally youre running tht)

Ah that makes sense, thanks.

DTCurrie:
@mcous updated the code per our last conversation, should we link it on your PR so we can test it out?

DTCurrie:
@jr22 @mcous Got tests passing on the app PR: https://github.com/viamrobotics/app/pull/6392

I think we are good to final review and merge this, cut a release, and then merge that app-side PR.

web/auth.go
```diff
@@ -413,79 +423,124 @@ type tokenResponse struct {
...
+func (h *tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+	defer cancel()
+
+	_, span := trace.StartSpan(ctx, r.URL.Path)
+	defer span.End()
+
+	current := getBearerToken(r)
+	data := getAuthCookieValues(r)
+	clearAuthCookies(w)
```
mcous, line 505: [Minor] We might want to combine these into a single `getAndClearAuthCookies` (or whatever name to communicate that the cookies are one-time-use) to make this difficult to accidentally break in the future

mcous (approved):
See smoke testing results at: https://github.com/viamrobotics/app/pull/6392#pullrequestreview-2429958898

web/auth.go
DTCurrie: Makes sense

web/auth.go
```diff
@@ -413,79 +423,120 @@ type tokenResponse struct {
...
+
+func (h *tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+	defer cancel()
+
+	_, span := trace.StartSpan(ctx, r.URL.Path)
+	defer span.End()
+
+	current := getBearerToken(r)
+	data := getAndClearAuthCookieValues(w, r)
```
jr22, line 501: note that theres a small change in behavior here where we will still clear cookies even if the json marshling fails. i think this is fine but calling it out

jr22 (approved):
2 small comments (not blocking) otherwise looks good

web/auth.go
DTCurrie: Good callout. I agree it's fine, if the marshalling fails it probably means something was broken/corrupted in the auth flow and it is probably best to start again with a clean state.

web/auth.go
```diff
@@ -413,79 +423,124 @@ type tokenResponse struct {
...
-	_, err = w.Write(data)
-	utils.UncheckedError(err)
+func (h *tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+	defer cancel()
+
+	_, span := trace.StartSpan(ctx, r.URL.Path)
+	defer span.End()
+
+	current := getBearerToken(r)
```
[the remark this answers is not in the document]
DTCurrie: Ah yeah, missed that when reordering some logic. I will update it.

PR #365, opened 2024-10-07
Remove rpc/js as it is now in TypeScript SDK

edaniels:
@maximpertsov I'll defer to you and team on deleting this e2e test here. I don't know how much value it servers compared to someone writing a viam machine specific version in TS-SDK.


edaniels:
Tests wont pass since workflow is running off of main. But we're just removing things here. There's no more js and just all go(utils)

maximpertsov:
> I'll defer to you and team on deleting this e2e test here. I don't know how much value it servers compared to someone writing a viam machine specific version in TS-SDK.

I added that E2E test because the JS rpc code didn't have _any_ test coverage before. You added some tests when you migrated to connect-es, right? If so then I think we can just rely on those and delete the existing E2E test.

maximpertsov (approved):

edaniels:
I'd say webrtc is probably not covered now in the TS SDK without an E2E test. But there may be one in app to pull over

PR #366, opened 2024-10-07
Do not connect to unix-schemed URIs through SOCKS proxy

benjirewis:
cc @viamrobotics/netcode ; forgot to exclude addresses starting with `unix` when behind SOCKS proxies.

### DIFF
--- a/rpc/dialer.go
+++ b/rpc/dialer.go
@@ -254,7 +254,8 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts dialOptions, logg
 	// Use SOCKS proxy from environment as gRPC proxy dialer. Do not use
 	// if trying to connect to a local address.
 	if proxyAddr := os.Getenv(SocksProxyEnvVar); proxyAddr != "" &&
-		!(strings.HasPrefix(address, "[::]") || strings.HasPrefix(address, "localhost")) {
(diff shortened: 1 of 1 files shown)


rpc/dialer.go
```diff
@@ -254,7 +254,8 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts dialOptions, logg
 	// Use SOCKS proxy from environment as gRPC proxy dialer. Do not use
 	// if trying to connect to a local address.
 	if proxyAddr := os.Getenv(SocksProxyEnvVar); proxyAddr != "" &&
-		!(strings.HasPrefix(address, "[::]") || strings.HasPrefix(address, "localhost")) {
+		!(strings.HasPrefix(address, "[::]") || strings.HasPrefix(address, "localhost") ||
+			strings.HasPrefix(address, "unix")) {
```
maximpertsov, line 258: basically a `unix` address means we are dialing locally right?

kind feels like this could either pull out a function - `validSOCKSAddress` or something - which could be a good place add docs about various inclusions/exclusions as needed. anyway, won't block, just a thought

maximpertsov (approved):

rpc/dialer.go
maximpertsov: works for me!

PR #367, opened 2024-10-08
RSDK-8952 Skip MDB tests until MDB instance is recreated

benjirewis:


testutils/mongodb.go
```diff
@@ -148,12 +148,18 @@ func backingMongoDBClientWithOptions(baseOptions *options.ClientOptions) (*mongo
...
-	client, err := backingMongoDBClient()
-	if err != nil {
-		skipWithError(tb, err)
-		return nil
-	}
-	return client
+	// TODO(RSDK-8952): Re-enable tests that interact with MDB once
+	// we resurrect the downed MBD instance or create a new one.
+	tb.Skip("skipping MongoDB related test for now; see RSDK-8952")
+	return nil
```
dgottlieb, line 154: This all looks fine -- but given the `tb.Skip` did you need to early return and avoid the underlying code? My understanding of skip is that it won't execute any more code.

dgottlieb: https://github.com/golang/go/blob/5428570af7f668c840569a596cb1c23644f408cf/src/testing/testing.go#L1128

dgottlieb (approved):

PR #370, opened 2024-10-09
Use pion/webrtc/v4

edaniels:
I'd suggest moving to v4 since v3 is no longer supported.


edaniels:
Test failure unrelated

dgottlieb:
All of the changes you made in the v3 branch for us are in the v4 branch? Particularly the ones that weren't included in the old v3 branch because they were deemed backwards breaking? 

I think the main one being the rtp passthrough related change where OnTrackReceived will be called once the SDP is renegotiated (as opposed to also requiring the first packet to be sent).

edaniels:
> > All of the changes you made in the v3 branch for us are in the v4 branch? Particularly the ones that weren't included in the old v3 branch because they were deemed backwards breaking?
> > I think the main one being the rtp passthrough related change where OnTrackReceived will be called once the SDP is renegotiated (as opposed to also requiring the first packet to be sent).
> 
> Yeah that RTP one would still need to be addressed here. Everything else in the fork is present in v4.

I'm offering up https://github.com/pion/webrtc/pull/2933 to see if we can get it in without simulcast changes (that viam doesn't use).

edaniels:
@dgottlieb I got the OnTrack bit in as a setting

rpc/wrtc_base_channel.go
```diff
@@ -206,12 +206,13 @@ func (ch *webrtcBaseChannel) onChannelClose() {
...
 func isUserInitiatedAbortChunkErr(err error) bool {
 	return err != nil && errors.Is(err, sctp.ErrChunk) &&
-		strings.Contains(err.Error(), "User Initiated Abort: Close called")
+		strings.Contains(err.Error(), "User Initiated Abort:")
 }
 
 func (ch *webrtcBaseChannel) onChannelError(err error) {
 	if errors.Is(err, sctp.ErrResetPacketInStateNotExist) ||
 		isUserInitiatedAbortChunkErr(err) {
+		ch.onChannelClose()
```
edaniels, line 215: @dgottlieb this plus a few lines above ends up fixing a case where some tests don't see the channel close with an error "in time"

edaniels:
@dgottlieb do you think we should move forward with this or abandon?

dgottlieb:
> I'm offering up https://github.com/pion/webrtc/pull/2933 to see if we can get it in without simulcast changes (that viam doesn't use).

> @dgottlieb I got the OnTrack bit in as a setting

Oops, I missed this update. That changes the calculus. Just to prepare for how painful this is to propagate/rollback: we really only have to change this repo + rdk? I think `app` just pulls this in as an indirect dependency?

dgottlieb (approved):

edaniels:
Yeah it'll be easy to revert. And yes for app too, shouldn't matter.

PR #374, opened 2024-10-21
SEC-53 don't backto to external urls

DTCurrie:
Critical security fix, but double-check my logic. See: https://viam.atlassian.net/browse/SEC-53

Would love help testing this.

### DIFF
--- a/web/auth.go
+++ b/web/auth.go
@@ -527,8 +527,14 @@ func (h *loginHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	})
 
 	if r.FormValue("backto") != "" {
-		session.Data["backto"] = r.FormValue("backto")
+		backto := r.FormValue("backto")
+
+		_, err := url.ParseRequestURI(backto)
+		if err == nil {
+			session.Data["backto"] = backto
+		}
 	}
+
 	if session.Data["backto"] == "" {
 		session.Data["backto"] = r.Header.Get("Referer")
 	}


jr22:
put up this app PR with temp env so we can test: https://github.com/viamrobotics/app/pull/6610

web/auth.go
```diff
@@ -527,8 +527,14 @@ func (h *loginHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	})
 
 	if r.FormValue("backto") != "" {
-		session.Data["backto"] = r.FormValue("backto")
+		backto := r.FormValue("backto")
+
+		_, err := url.ParseRequestURI(backto)
+		if err == nil {
+			session.Data["backto"] = backto
+		}
```
jr22, line 535: from dms

> My theory is any external url would be valid so we shouldn’t allow that, only set backto when we detect an invalid url.

so should this be `err != nil`?

jr22: i tested in the temp env from the linked app pr and still run into the same issue described into the ticket: 

https://pr-6610-appmain-bplesliplq-uc.a.run.app/login?backto=https://www.sumitsahoo.com/

DTCurrie: Sorry yes, `!=` I switched the logic around and forgot to swap the operator, thanks

jr22: ok confirmed in test env that it doesnt work anymore. have been trying to find places where we do leverage `backto` to make sure it doesnt break but none of them seem to be working. now confirming that this is existing behavior

DTCurrie: @dgottlieb posted in app-chat earlier that he noticed he wasn't being redirected back to the right place post-login: https://viaminc.slack.com/archives/C02KFSY4H89/p1729521583147719

DTCurrie:
@ohEmily  @jr22 I tested on the temp env on the [app PR](https://github.com/viamrobotics/app/pull/6610) and I think it is working. Take a look and feel free to approve/merge this PR if we are good.

web/auth.go
```diff
@@ -527,8 +527,14 @@ func (h *loginHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	})
 
 	if r.FormValue("backto") != "" {
-		session.Data["backto"] = r.FormValue("backto")
+		backto := r.FormValue("backto")
+
+		_, err := url.ParseRequestURI(backto)
+		if err != nil {
```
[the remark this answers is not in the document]
DTCurrie: Yep, good call

DTCurrie:
> tested and this does fix the exact problem described. not sure if it fully closes the security hole (not saying that because i can think of something but because i'm honestly not sure) but feels good enough for now. alternatively, i couldnt get any of the `backto` logic to work correctly (on this or on prod) so possible we could remove this all together

It's possible the frontend is no longer sending the `backto` param to `/login`. I can check that, but it will be a PR in `app` so not blocking for this PR. Going to merge.

PR #375, opened 2024-10-25
RSDK-9150 - Client log interceptor should not use zap.Logger

cheukt:
no reason to convert the logger to a zap logger and then stick it in the context

before changes:
```
2024-10-25T13:29:18.659-0400    ERROR   rdk.resource_manager.rdk:remote:/robot2.networking.networking.webrtc     zap/options.go:212 finished client unary call
{"system": "grpc", "span.kind": "client", "grpc.service": "viam.robot.v1.RobotService",
"grpc.method": "ResourceNames", "error": "rpc error: code = Unknown desc = ggg", "grpc.code": "Unknown", "grpc.time_ms": 0.6}

2024-10-25T15:14:21.806-0400    DEBUG   rdk.resource_manager.rdk:remote:/robot2.networking.networking.webrtc   zap/options.go:212 finished client unary call     
 {"system": "grpc", "span.kind": "client", "grpc.service": "viam.robot.v1.RobotService", "grpc.method": "ResourceNames", "grpc.code": "OK", "grpc.time_ms": 0.538}
```

after changes:
```
2024-10-25T19:20:50.385Z        ERROR   rdk.resource_manager.rdk:remote:/robot2.networking.networking.webrtc   rpc/wrtc_client_channel.go:358 finished client unary call
{"error":"rpc error: code = Unknown desc = hello","grpc.code":"Unknown","grpc.time_ms":0.76,"span.kind":"client","system":"grpc","grpc.service":"viam.robot.v1.RobotService","grpc.method":"ResourceNames"}

2024-10-25T19:19:17.405Z        DEBUG   rdk.resource_manager.rdk:remote:/robot2.networking.networking.webrtc   rpc/wrtc_client_channel.go:354 finished client unary call
{"grpc.code":"OK","grpc.time_ms":0.345,"span.kind":"client","system":"grpc","grpc.service":"viam.robot.v1.RobotService","grpc.method":"ResourceRPCSubtypes"}
```

field order is different, since zap loggers put the With() fields in front whereas the RDK logger puts them in the back


logger.go
```diff
@@ -24,6 +25,7 @@ type ILogger interface {
 // ZapCompatibleLogger is a basic logging interface for golog.Logger (alias for *zap.SugaredLogger) and RDK loggers.
 type ZapCompatibleLogger interface {
 	Desugar() *zap.Logger
+	Level() zapcore.Level
```
cheukt, line 28: adding Level seems to be the easiest thing to do (since it's already present in the RDK version of this interface), but curious if there are other options.

rpc/wrtc_client_channel.go
```diff
@@ -336,29 +333,44 @@ func (ch *webrtcClientChannel) writeReset(stream *webrtcpb.Stream) error {
...
+			logger.Debugw(msg, fields...)
+		case zap.InfoLevel:
+			logger.Infow(msg, fields...)
+		case zap.ErrorLevel:
+			logger.Errorw(msg, fields...)
+		case zap.WarnLevel:
+			fallthrough
+		default:
+			logger.Warnw(msg, fields...)
+		}
```
cheukt, line 363: this is pretty ugly, but the `Write()` methods differ between zap loggers and rdk loggers so I could write another conversion method like `AddFieldsToLogger` or do this

rpc/wrtc_client_channel.go
```diff
@@ -336,29 +333,44 @@ func (ch *webrtcClientChannel) writeReset(stream *webrtcpb.Stream) error {
...
+func newClientLoggerFields(fullMethodString string) []any {
 	service := path.Dir(fullMethodString)[1:]
 	method := path.Base(fullMethodString)
-	return []zapcore.Field{
-		systemField,
-		clientField,
-		zap.String("grpc.service", service),
-		zap.String("grpc.method", method),
+	return []any{
+		"span.kind", "client",
```
cheukt, line 371: note: rdk loggers' WithFields do not play nicely with zapcore.Field at all. which probably means that zap loggers will not play nicely with `AddFieldsToLogger`

rpc/wrtc_client_channel.go
```diff
@@ -336,29 +333,42 @@ func (ch *webrtcClientChannel) writeReset(stream *webrtcpb.Stream) error {
...
+	duration := float32(time.Since(startTime).Nanoseconds()/1000) / 1000
+	fields := []any{}
 	if err == nil {
 		level = zap.DebugLevel
+	} else {
+		fields = append(fields, "error", err)
+	}
+	fields = append(fields, "grpc.code", code.String(), "grpc.time_ms", duration)
+
+	if logger.Level().Enabled(level) {
```
dgottlieb, line 350: Why is this necessary? I would expect if the level were not enabled, then the calls to `logger.Debuw`/`logger.Infow`/etc... would early return. Are we not observing that?

cheukt: ya this is probably unnecessary, let me check. I was thinking too much about replicating the functionality of grpc_zap.DefaultMessageProducer, which also does a check before logging

dgottlieb: Cool, definitely speak up if removing this is problematic.

If removing is not problematic, then we can cascade and delete the [interface change](https://github.com/viamrobotics/goutils/pull/375/files#diff-ff87b7c4777a35588053a509583d66c9f404ccbea9e1c71d2a5f224d7ad1323eR28) that added `Level` and the dummy implementations for the [testing](https://github.com/viamrobotics/goutils/pull/375/files#diff-55a2db70576eb079880167200221ea3787a01fa4bbc958503769eba24084d842R68) [cases](https://github.com/viamrobotics/goutils/pull/375/files#diff-55a2db70576eb079880167200221ea3787a01fa4bbc958503769eba24084d842R87).

rpc/wrtc_client_channel.go
```diff
@@ -336,29 +333,42 @@ func (ch *webrtcClientChannel) writeReset(stream *webrtcpb.Stream) error {
...
 // taken from
 // https://github.com/grpc-ecosystem/go-grpc-middleware/blob/560829fc74fcf9a69b7ab01d484f8b8961dc734b/logging/zap/client_interceptors.go
-
-func logFinalClientLine(ctx context.Context, startTime time.Time, err error, msg string) {
+func logFinalClientLine(logger utils.ZapCompatibleLogger, startTime time.Time, err error, msg string) {
 	code := grpc_logging.DefaultErrorToCode(err)
 	level := grpc_zap.DefaultCodeToLevel(code)
+
+	// this calculation is done because duration.Milliseconds() will return an integer, which is not precise enough.
+	duration := float32(time.Since(startTime).Nanoseconds()/1000) / 1000
```
benjirewis, line 341: Seems fair, but why did you feel we needed the extra precision?

logger.go
benjirewis: What you did seems good to me.

rpc/wrtc_client_channel.go
benjirewis: Thanks for filing that ticket; sounds like zap loggers will almost certainly not play nicely with `AddFieldsToLogger`. But, from the sample "after changes" output you put in the PR description, the added fields do seem to be getting added. Is that right?

logger.go
cheukt: removed since it was unnecessary

rpc/wrtc_client_channel.go
cheukt: yep, but the way they're getting passed in now is different - no longer a list of zap.Fields

rpc/wrtc_client_channel.go
cheukt: updated to remove

rpc/wrtc_client_channel.go
cheukt: a lot of calls take less than 1ms, so would just show up as 0ms. the math here ends up giving 3 decimal places, which is pretty good

benjirewis (approved):
Thanks!

dgottlieb (approved):

PR #380, opened 2024-10-30
Fix backto check

DTCurrie:
Quick fix: It looks like paths were not throwing an error, so I added a check to ensure a valid URL does not have a hostname. This should restrict it to just internal paths. Tested and this resolves the issue we were seeing with e2e tests in app trying to bump the goutils version.

### DIFF
--- a/web/auth.go
+++ b/web/auth.go
@@ -529,9 +529,11 @@ func (h *loginHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if r.FormValue("backto") != "" {
 		backto := r.FormValue("backto")
 
-		// to prevent redirecting to an external URL we only set the session data when we fail to parse backto
-		_, err := url.ParseRequestURI(backto)
-		if err != nil {
+		// to prevent redirecting to an external URL we only set the session data when:
+		// 1. we fail to parse backto
+		// 2. backto does not include a hostname
+		parsed, err := url.ParseRequestURI(backto)
+		if err != nil || parsed.Hostname() == "" {
 			session.Data["backto"] = backto
 		}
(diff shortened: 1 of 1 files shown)


web/auth.go
```diff
@@ -529,9 +529,11 @@ func (h *loginHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		backto := r.FormValue("backto")
 
-		// to prevent redirecting to an external URL we only set the session data when we fail to parse backto
-		_, err := url.ParseRequestURI(backto)
-		if err != nil {
+		// to prevent redirecting to an external URL we only set the session data when:
+		// 1. we fail to parse backto
+		// 2. backto does not include a hostname
+		parsed, err := url.ParseRequestURI(backto)
+		if err != nil || parsed.Hostname() == "" {
```
DTCurrie, line 536: Wondering if we need 1 if paths are always passing that parse. May just be happy to reduce to:

```suggestion
		// to prevent redirecting to an external URL we only set the session data when backto does not include a hostname
		parsed, err := url.ParseRequestURI(backto)
		if err != nil && parsed.Hostname() == "" {
```

zaporter-work: [unfamiliar with this area of the codebase]

I don't think `&& parsed.Hostname()` will work -- I think it will panic because `parsed` is nil when err != nil.

This works as `||` because the second check is short-circuted and not evaluated when err != nil

DTCurrie: Yeah I wrote this up quickly during a meeting but thinking I am ok with the code as is unless anyone has major complaints.

jr22 (approved):

PR #385, opened 2024-11-07
RSDK-8819: Instrument some grpc/webrtc FTDC stats.

dgottlieb:
There are some (minimal) RDK changes to get this hooked up. Here's an example of how these stats show up.

image

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/edaniels/zeroconf"
@@ -102,6 +103,10 @@ type Server interface {
 	http.Handler
 
 	EnsureAuthed(ctx context.Context) (context.Context, error)
+
+	// Stats returns a structure containing numbers that can be interesting for graphing over time
+	// as a diagnostics tool.
+	Stats() any
 }
 
 type simpleServer struct {
@@ -146,6 +151,15 @@ type simpleServer struct {
 
 	// authIssuer is the JWT issuer (iss) that will be used for our service.
 	authIssuer string
+
+	counters struct {
+		TcpGrpcRequestsStarted      atomic.Int64
+		TcpGrpcWebRequestsStarted   atomic.Int64
+		TcpOtherRequestsStarted     atomic.Int64
+		TcpGrpcRequestsCompleted    atomic.Int64
+		TcpGrpcWebRequestsCompleted atomic.Int64
+		TcpOtherRequestsCompleted   atomic.Int64
+	}
 }
 
 var errMixedUnauthAndAuth = errors.New("cannot use unauthenticated and auth handlers at same time")
@@ -721,13 +735,19 @@ func (ss *simpleServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r = requestWithHost(r)
 	switch ss.getRequestType(r) {
 	case requestTypeGRPC:
+		ss.counters.TcpGrpcRequestsStarted.Add(1)
 		ss.grpcServer.ServeHTTP(w, r)
+		ss.counters.TcpGrpcRequestsCompleted.Add(1)
 	case requestTypeGRPCWeb:
+		ss.counters.TcpGrpcWebRequestsStarted.Add(1)
 		ss.grpcWebServer.ServeHTTP(w, r)
+		ss.counters.TcpGrpcWebRequestsCompleted.Add(1)
 	case requestTypeNone:
 		fallthrough
 	default:
+		ss.counters.TcpOtherRequestsStarted.Add(1)
 		ss.grpcGatewayHandler.ServeHTTP(w, r)
+		ss.counters.TcpOtherRequestsCompleted.Add(1)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


rpc/wrtc_server.go
```diff
@@ -43,6 +44,48 @@ type webrtcServer struct {
...
+	ret := WebRTCGrpcStats{
+		PeersConnected:            srv.counters.PeersConnected.Load(),
+		PeersDisconnected:         srv.counters.PeersDisconnected.Load(),
+		PeerConnectionErrors:      srv.counters.PeerConnectionErrors.Load(),
+		HeadersProcessed:          srv.counters.HeadersProcessed.Load(),
+		CallTicketsAvailable:      srv.counters.CallTicketsAvailable.Load(),
+		TotalTimeConnectingMillis: srv.counters.TotalTimeConnectingMillis.Load(),
+	}
+	ret.PeersActive = ret.PeersConnected - ret.PeersDisconnected
+	ret.AverageTimeConnectingMillis = float64(ret.TotalTimeConnectingMillis) / float64(ret.PeersConnected)
```
dgottlieb, line 87: The graph for this was all zeroes. The values this is derived from looked good so I need to figure out what happened here.

dgottlieb: Figured this out. Because `PeersConnected` starts at 0, golang assigns `NaN` to `AverageTimeConnectingMillis`.

And the ftdc format did not see that going from nan -> y(es)an as a "diff".

rpc/server.go
```diff
@@ -146,6 +151,15 @@ type simpleServer struct {
 
 	// authIssuer is the JWT issuer (iss) that will be used for our service.
 	authIssuer string
+
+	counters struct {
+		TcpGrpcRequestsStarted      atomic.Int64
```
benjirewis, line 156: Obviously lint on all of these, but probably also some comments describing what they're counting (I realize it's pretty obvious.)

rpc/server.go
```diff
@@ -861,6 +881,34 @@ func (ss *simpleServer) Stop() error {
 	return err
 }
 
+type SimpleServerStats struct {
```
benjirewis, line 884: Same comment about lint and comments here.

Do types returned by `Stats` methods need to be exported for FTDC's reflection to work? If not, should we just keep all of this API private since it is all within the `rpc` package?

rpc/wrtc_server.go
```diff
@@ -43,6 +44,51 @@ type webrtcServer struct {
...
 	onPeerAdded   func(pc *webrtc.PeerConnection)
 	onPeerRemoved func(pc *webrtc.PeerConnection)
+
+	counters struct {
+		PeersConnected       atomic.Int64
+		PeersDisconnected    atomic.Int64
+		PeerConnectionErrors atomic.Int64
+		HeadersProcessed     atomic.Int64
+		// CallTicketsAvailable is technically a guage. It increments and decrements. Rather than
+		// just going up.
```
benjirewis, line 54: Are you calling this out because you would recommend monotonically-increasing metrics instead of ones that increment and decrement?

rpc/wrtc_server.go
```diff
@@ -43,6 +44,51 @@ type webrtcServer struct {
 
 	onPeerAdded   func(pc *webrtc.PeerConnection)
 	onPeerRemoved func(pc *webrtc.PeerConnection)
+
+	counters struct {
```
benjirewis, line 48: I think this an awesome start for WebRTC stats + will not recommend any others for the time being.

rpc/wrtc_server.go
```diff
@@ -43,6 +44,51 @@ type webrtcServer struct {
...
+type WebRTCGrpcStats struct {
+	PeersConnected            int64
+	PeersDisconnected         int64
+	PeerConnectionErrors      int64
+	PeersActive               int64
+	HeadersProcessed          int64
+	CallTicketsAvailable      int32
+	TotalTimeConnectingMillis int64
+
+	// When the FTDC frontend is more feature reach, we can remove this and let the frontend compute
```
benjirewis, line 71:
```suggestion
	// When the FTDC frontend is more feature rich, we can remove this and let the frontend compute
```

rpc/wrtc_server.go
```diff
@@ -85,6 +131,7 @@ func newWebRTCServerWithInterceptorsAndUnknownStreamHandler(
 		streamInt:         streamInt,
 		unknownStreamDesc: unknownStreamDesc,
 	}
+	srv.counters.CallTicketsAvailable.Store(int32(DefaultWebRTCMaxGRPCCalls))
```
benjirewis, line 134: Are you not using `len(callTickets)` in `Stats` (and relying on this atomic) because of worries about a data race? I didn't think `len(callTickets)` concurrently with `<-callTickets` or `callTickets<-` would be cause for a race, but maybe I am wrong.

benjirewis (commented):
Nice! Some questions (let me know if I'm going too deep on this stuff and you just want to get something in).

rpc/wrtc_server.go
dgottlieb: Forgot len/cap worked on channels -- changed.

rpc/server.go
benjirewis: Sure; sounds good to me. Thanks for the reasoning.

benjirewis (approved):
LGTM! Thanks.

PR #386, opened 2024-11-11
RSDK-9239: Fix off by one buffer threshold waiting semantics.

dgottlieb:
Not a problem that I believe is responsible for anything we've seen. But worth fixing none the less to avoid any uncertainty.

https://viam.atlassian.net/browse/RSDK-9239 has more code links that justify my claim of this being an off-by-one error that can cause a hang


Kschappacher (commented):
Can you give me a bit more on why this off by one check stops RDK from hanging?

dgottlieb:
In case we raced, I missed a file save so the comment didn't come in the first GH push.

Additionally I put more context in the JIRA ticket (linked from this PR description, which I think is a better long-term spot). But if you've read that (code links were provided) and still aren't clear, I'm happy to talk it through.

dgottlieb:
> stops RDK from hanging?

In case there's confusion, I would characterize the symptom of this bug as, an individual PeerConnection goes dormant and can't make progress sending (already generated) responses.

Because we do hold "call tickets" while doing this "waiting", that could* snowball into the RDK hanging due to a lack of available call tickets.

Kschappacher (approved):
Amazing, just read the jira ticket and the logic is sound. LGTM

PR #388, opened 2024-11-11
RSDK-8779: Remove additional cases where signaling server errors override successful PeerConnection attempts.

dgottlieb:
Sorry I didn't really take the time to explain the step-by-step. I'm a little impatient with these errors at this point.


rpc/wrtc_client.go
```diff
@@ -230,37 +218,51 @@ func dialWebRTC(
...
 			}
+
 			// must spin off to unblock the ICE gatherer
 			utils.PanicCapturingGo(func() {
 				if icecandidate != nil {
 					defer pendingCandidates.Done()
 				}
 				select {
 				case <-remoteDescSet:
+					// We've sent the `init` offer. We can now proceed with sending individual
```
benjirewis, line 252: I thought `remoteDescSet` gets closed when we've _received_ an init offer from the answerer in a `CallResponse_Init`.
```suggestion
					// We've received the `init` offer. We can now proceed with sending individual
```

rpc/wrtc_client.go
```diff
@@ -365,41 +372,29 @@ func dialWebRTC(
...
+		// Happy path
+		sendDone()
+		successful = true
+	case <-exchangeCtx.Done():
+		sendDoneOnce.Do(func() {
 			_, err = signalingClient.CallUpdate(exchangeCtx, &webrtcpb.CallUpdateRequest{
 				Uuid: uuid,
 				Update: &webrtcpb.CallUpdateRequest_Error{
-					Error: ErrorToStatus(callErr).Proto(),
+					Error: ErrorToStatus(exchangeCtx.Err()).Proto(),
```
benjirewis, line 391: With your changes here, the only error that a caller can send back the signaling server (and potentially to the answerer through the signaling server) is a context canceled or deadline exceeded from the `exchangeCtx`? That makes sense to me but wanted to clarify.

rpc/wrtc_client.go
```diff
@@ -187,27 +190,12 @@ func dialWebRTC(
...
-	}
 	var uuid string
 	// only send once since exchange may end or ICE may end
-	var sendDoneErrorOnce sync.Once
+	var sendDoneOnce sync.Once
 	sendDone := func() error {
 		var err error
-		sendDoneErrorOnce.Do(func() {
+		sendDoneOnce.Do(func() {
 			_, err = signalingClient.CallUpdate(exchangeCtx, &webrtcpb.CallUpdateRequest{
```
benjirewis, line 199: Do we want to `Warn`-log any errors from this `CallUpdate` and have `sendDone` return void? I'm assuming you'll get rid of the "`sendDone` return value ignored" lint errors somehow but curious if you'll want to log the errors or literally ignore them. I would vote for the former.

rpc/wrtc_client.go
```diff
@@ -230,37 +218,51 @@ func dialWebRTC(
...
 
 		var pendingCandidates sync.WaitGroup
+
+		// waitOneHost is closed when the first ICE candidate of type `Host` (e.g: 127.0.0.1) is
+		// found.
 		waitOneHost := make(chan struct{})
 		var waitOneHostOnce sync.Once
 		peerConn.OnICECandidate(func(icecandidate *webrtc.ICECandidate) {
 			if exchangeCtx.Err() != nil {
+				// We've decided to bail.
```
benjirewis, line 228:
```suggestion
				// Client has canceled the dial or a timeout has occurred.
```

benjirewis (commented):
Definitely agree with the changes, some small nits.

rpc/wrtc_client.go
dgottlieb: I think there's two issues here:
* I chose words poorly -- it's not important that we sent the offer, but received the "init response".
* While that response does contain the "answer" (not the "offer") -- the thing this function actually cares about is the "exchange UUID". It's legal to [read this variable](https://github.com/viamrobotics/goutils/blob/e8de5415d7fc5848e5d7a80e62c04f4fa2f38154/rpc/wrtc_client.go?plain=1#L267) because we've observed the channel being closed (i.e: in happens-before relationship notation, the uuid being written to happens before the channel is closed).

rpc/wrtc_client.go
dgottlieb: Done, but used "Caller" instead of "Client"

rpc/wrtc_client.go
benjirewis: > "Thanks for picking up the phone! I just wanted to say I'm calling off the wedding because you didn't pick up the phone."

😆

benjirewis (approved):
This LGTM % the lint issues.

rpc/wrtc_client.go
benjirewis: Yeah good clarification about the UUID; what you have now makes sense to me.

dgottlieb:
There's a test unhappy with these changes. It wants to assert on error messages. The test is otherwise valuable, so I'm evaluating options.

logger.go
```diff
@@ -25,6 +25,10 @@ type ILogger interface {
 type ZapCompatibleLogger interface {
 	Desugar() *zap.Logger
 
+	// Not defined: Named(name string) *zap.SugaredLogger
```
dgottlieb, line 28: I had a whole episode of adding `Named` here before realizing why we explicitly need to avoid it.

rpc/dial.go
```diff
@@ -178,19 +179,17 @@ func dial(
...
 				// This will skip all Authenticate/AuthenticateTo calls for the signaler.
 				if !dOpts.usingMDNS && dOpts.authMaterial == "" && dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial != "" {
-					logger.Debug("using signaling's external auth as auth material")
+					webrtcLogger.Debug("using signaling's external auth as auth material")
 					dOpts.authMaterial = dOpts.webrtcOpts.SignalingExternalAuthAuthMaterial
 					dOpts.creds = Credentials{}
 				}
 			}
 
-			if dOpts.debug {
```
dgottlieb, line 187: Felt unnecessary now

rpc/dial_test.go
```diff
@@ -1150,10 +1150,11 @@ func TestDialMulticastDNS(t *testing.T) {
 
 	t.Run("fix mdns instance name", func(t *testing.T) {
 		rpcServer, err := NewServer(
-			logger,
+			logger.Named("server"),
```
dgottlieb, line 1153: Not 100% intentional. I think* these are fine given the logger is golog and this is in tests and not called from RDK.

rpc/wrtc_signaling_server.go
```diff
@@ -152,8 +152,37 @@ func HeartbeatsAllowedFromCtx(ctx context.Context) bool {
 	return md[HeartbeatsAllowedMetadataField][0] == "true"
 }
 
+func (srv *WebRTCSignalingServer) asyncSetOfferError(host string, uuid string, offerErr error) {
```
dgottlieb, line 155: This was the big lambda i factored out.

rpc/wrtc_signaling_server.go
```diff
@@ -152,8 +152,37 @@ func HeartbeatsAllowedFromCtx(ctx context.Context) bool {
...
+	// Atomically check if the cancelCtx was canceled, and if not, add to the `callWorkers`.
+	if err := srv.cancelCtx.Err(); err != nil {
+		srv.callMu.RUnlock()
+		return
+	}
+	srv.callWorkers.Add(1)
+	srv.callMu.RUnlock()
+
+	utils.PanicCapturingGo(func() {
+		defer srv.callWorkers.Done()
```
dgottlieb, line 166: This changed to a `defer`

rpc/wrtc_signaling_server.go
```diff
@@ -171,70 +200,27 @@ func (srv *WebRTCSignalingServer) Call(req *webrtcpb.CallRequest, server webrtcp
 	}
 	defer sendCancel()
 
-	defer func() {
-		var errToSend error
-
-		if callErr == nil {
```
dgottlieb, line 177: Notable behavior change is that this `callErr == nil` block will never result in calling `SendOfferError`.

rpc/wrtc_signaling_server.go
```diff
@@ -152,8 +152,37 @@ func HeartbeatsAllowedFromCtx(ctx context.Context) bool {
...
+	srv.callMu.RUnlock()
+
+	utils.PanicCapturingGo(func() {
+		defer srv.callWorkers.Done()
+
+		// Use a timeout to not block shutdown.
+		sendCtx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+		defer cancel()
+
+		sendErr := srv.callQueue.SendOfferError(sendCtx, host, uuid, offerErr)
```
dgottlieb, line 172: Also referenced in the deleted code: a "context timeout" that's concurrent with a "return nil // the connection has succeeded" should never result in this `SendOfferError` being invoked.

rpc/wrtc_signaling_server.go
```diff
@@ -263,6 +250,8 @@ func (srv *WebRTCSignalingServer) Call(req *webrtcpb.CallRequest, server webrtcp
 				},
 			},
 		}); err != nil {
+			// Don't set an error for the connection attempt. Some candidates may have been
+			// exchanged that can result in a successful connection.
```
benjirewis, line 254: Thanks as always for the in-line comments; totally agree with not sending error here.

benjirewis (commented):
Pretty much LGTM; one comment about naming.

rpc/dial_test.go
dgottlieb: > Any reason you needed to create better logger names here, though? Easier to read test output or something?

"easier to read" is a polite way of putting it. At least in the `authenticated` subtest, we make 3 separate dials. It was nigh impossible to parse which "base channel/stream" was client and which was server.

rpc/wrtc_signaling_server.go
```diff
@@ -152,8 +152,37 @@ func HeartbeatsAllowedFromCtx(ctx context.Context) bool {
 	return md[HeartbeatsAllowedMetadataField][0] == "true"
 }
 
+func (srv *WebRTCSignalingServer) asyncSetOfferError(host, uuid string, offerErr error) {
```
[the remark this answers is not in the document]
dgottlieb: Definitely a selfish change that I can just revert -- not a lot of utility and is a bit of a lightning rod. Almost all of the value of "publishing" the error is for tests. And specifically the tests that are setting up a failing connection attempt and want to assert we failed for the right reason.

And in those cases, we are using the "in memory" signaling service.

So your intuition on set vs. send was the intention I was aiming to achieve.

PR #389, opened 2024-11-11
RSDK-8941: Remove mutex dependence on base channel `closed` state variable.

dgottlieb:
In the absence of a lock ordering between wrtc channels and their streams, we can avoid deadlocks by using atomics instead of locks. To the degree that substitution is _correct_ is for another time.

### DIFF
--- a/rpc/wrtc_base_channel.go
+++ b/rpc/wrtc_base_channel.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/pion/dtls/v2"
 	"github.com/pion/sctp"
@@ -22,7 +23,7 @@ type webrtcBaseChannel struct {
 	ctx                     context.Context
 	cancel                  func()
 	ready                   chan struct{}
-	closed                  bool
+	closed                  atomic.Bool
 	closedReason            error
 	activeBackgroundWorkers sync.WaitGroup
 	logger                  utils.ZapCompatibleLogger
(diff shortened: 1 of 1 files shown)


rpc/wrtc_server_stream.go
```diff
@@ -221,31 +221,34 @@ func (s *webrtcServerStream) SendMsg(m interface{}) (err error) {
 }
 
 func (s *webrtcServerStream) onRequest(request *webrtcpb.Request) {
+	// Error cases here are logged at the warn level. It's not a server error to find client
```
dgottlieb, line 224: This was found with blinky and already having 3-4 PRs open I decided to just lump these in here rather than make a new ticket.

Specifically, blinky got an error on [this assertion](https://github.com/viamrobotics/rdk/blob/a3262bfbf30c391a7f45804144e9d9184fc30d43/robot/impl/resource_manager_test.go#L1756).

The log that caused the failure seemed typical:
```
    wrtc_server_stream.go:243: 2024-11-11T21:02:13.100Z	ERROR	remote1.networking.grpc_requests	rpc/wrtc_server_stream.go:243	error closing	{"error":"io: read/write on closed pipe","method":"/grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo"}
```

Alternatively, we can remove that assertion. It's unclear to me what it was protecting against. But as I stated above, I didn't feel the desire to create a new ticket/branch/PR just to remove that assertion.

dgottlieb: Half strike this -- my first fix was incomplete in that the `closedReason` was now being accessed without a mutex.

Rather than moving that to another "atomic" value, I decided to just remove it. It was only being used to fill in log messages.

rpc/wrtc_server_stream.go
```diff
@@ -354,44 +340,34 @@ func (s *webrtcServerStream) processMessage(msg *webrtcpb.RequestMessage) {
...
-		if writeErr == nil || errors.Is(writeErr, sctp.ErrStreamClosed) {
-			writeErr = nil
-		}
-	}()
 	defer func() {
 		s.webrtcBaseStream.mu.Lock()
 		defer s.webrtcBaseStream.mu.Unlock()
 		s.close()
 	}()
-	if err != nil && (errors.Is(err, io.ErrClosedPipe)) {
```
dgottlieb, line 368: I'm reintroducing this

dgottlieb:
Based on the lint error, I'm just going to remove `closeWithError`.

dgottlieb:
I'm hitting a data race/leaked goroutines with this patch. I need to revisit/double check this is a regression.

PR #391, opened 2024-11-18
RSDK-8941: Remove mutex dependence on base channel closed state variable.

dgottlieb:
In the absence of a lock ordering between wrtc channels and their streams, we
can avoid deadlocks by using atomics instead of locks. To the degree that
substitution is _correct_ is for another time.


cheukt (approved):
LGTM

PR #392, opened 2024-11-18
RSDK-9290 - Bump go to 1.23

cheukt:
it was more work to get the linter to work properly


etc/.golangci.yaml
```diff
@@ -1,57 +1,53 @@
...
-    - genfiles$
-    - gen$
-    - vendor$
-    - test$
+  skip-dirs-use-default: false
   tests: true
 linters:
   enable-all: true
   disable:
     - asasalint
```
cheukt, line 12: all the changes to disabled linters are mirrored from rdk

perf/exporters.go
```diff
@@ -57,7 +57,7 @@ func NewCloudExporter(opts CloudOptions) (Exporter, error) {
 		zone := os.Getenv("GCP_COMPUTE_ZONE")
 		if zone == "" {
 			// Get from GCP Metadata
-			if zone, err = metadata.Zone(); err != nil {
+			if zone, err = metadata.ZoneWithContext(sdOpts.Context); err != nil {
```
cheukt, line 60: could see just using a background context as well, not clear if the context used here is the right one

abe-winter (approved):

dgottlieb (approved):

PR #394, opened 2024-11-20
RSDK-9316 Disallow client stream creation when channel has been closed

benjirewis:
RSDK-9316

Checks the underlying channel's context for error before creating a new client stream. Returns a new `ErrDisconnected` error if the context has errored. A PR like [this one](https://github.com/viamrobotics/rdk/pull/4571) would make `rdk` sensitive to that new error instead of just waiting for `io.ErrClosedPipe` after an eventual read/write (?) from/to the closed channel.

Confirmed that this PR plus the linked `rdk` one gets rid of the race seen in https://github.com/viamrobotics/rdk/pull/4566 CI.

### DIFF
--- a/rpc/wrtc_client_stream.go
+++ b/rpc/wrtc_client_stream.go
@@ -58,7 +58,9 @@ func newWebRTCClientStream(
 	// We could rely on eventual reads/writes from/to the stream failing with a
 	// `io.ErrClosedPipe`, but not checking the channel's context here will mean
 	// we can create a stream _while_ the channel is closing/closed, which can
-	// result in data races and undefined behavior.
+	// result in data races and undefined behavior. The caller to this function
+	// is holding the channel mutex that's also acquired in the "close" path that
+	// will cancel `channel.ctx`.
 	if channel.ctx.Err() != nil {
 		return nil, ErrDisconnected
 	}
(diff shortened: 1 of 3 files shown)


rpc/wrtc_client_stream.go
```diff
@@ -45,7 +50,14 @@ func newWebRTCClientStream(
 	stream *webrtcpb.Stream,
 	onDone func(id uint64),
 	logger utils.ZapCompatibleLogger,
-) *webrtcClientStream {
+) (*webrtcClientStream, error) {
+	// Assume that cancelation of the client channel's context means the peer
```
dgottlieb, line 54: This is a good start to a comment. But I feel like there's a bit of "what's the consequence of removing this"? This just looks like an optimization to inform the caller sooner that their invocation/stream isn't going to work.

benjirewis: Added some elaboration there; let me know what you think. Thanks for asking for more, will be helpful for posterity.

benjirewis: For sure; included (almost) exactly what you said there.

dgottlieb (approved):

cheukt (approved):

PR #395, opened 2024-11-20
Fix nil check on nil pointers to structures.

dgottlieb:


cheukt:
https://github.com/dgottlieb/smarty-assertions/commit/0eedb634a98691dd4259a712f279be27d8954199 is the relevant commit I should be looking at?

dgottlieb:
If you'd like to, yep

cheukt (approved):

PR #399, opened 2024-12-27
RSDK-9591 - Add KillGroup to ManagedProcess

cheukt:
This is part one of two PRs that will hopefully help with shutting down all module processes before viam-server exits. Part 2 is [here](https://github.com/viamrobotics/rdk/pull/4657)

This is still a draft as I'm looking for thoughts and ideas around making this better. 

The idea behind this is for a Kill() call to propagate from the viam-server, and we should not block on anything if possible. The `kill()`s here do not wait for the execed process to complete, and this is intentional - I couldn't find if syscall.Kill waits for completion, so did not use. I wasn't able to find documentation about whether `kill -9` blocks on the signal getting received or if it immediately returns, so opted for `.Start()` instead of `.Run()`, which means we do not `Wait()` for the process to return. This does mean that the assumption is that the viam-server ends immediately after and that the system will properly await the processes that we spawned to kill the managed process, otherwise those processes will be zombie'd. I can see changing it so that we do a `.Run()` instead, but might mean we would have to do the module process killing in parallel in viam-server.

There is a todo to add Kill to the processManager, but a quick glance at how Close works for processManager makes me think that it should be a followup task

### DIFF
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -34,7 +34,8 @@ type ManagedProcess interface {
 
 	// KillGroup will attempt to kill the process group and not wait for completion. Only use this if
 	// comfortable with leaking resources (in cases where exiting the program as quickly as possible is desired).
-	KillGroup()
+	// Returns kill process that can be waited on if desired.
+	KillGroup() (*exec.Cmd, error)
 
 	// Status return nil when the process is both alive and owned.
 	// If err is non-nil, process may be a) alive but not owned or b) dead.
@@ -438,7 +439,7 @@ func (p *managedProcess) Stop() error {
 }
 
 // KillGroup kills the process group.
-func (p *managedProcess) KillGroup() {
+func (p *managedProcess) KillGroup() (*exec.Cmd, error) {
 	// Minimally hold a lock here so that we can signal the
 	// management goroutine to stop. We will attempt to kill the
 	// process even if p.stopped is true.
@@ -450,7 +451,7 @@ func (p *managedProcess) KillGroup() {
 
 	if p.cmd == nil {
 		p.mu.Unlock()
-		return
+		return nil, errAlreadyStopped
 	}
 	p.mu.Unlock()
 
@@ -460,6 +461,5 @@ func (p *managedProcess) KillGroup() {
 	// without a lock held.
 	// We are intentionally not checking the error here, we are already
 	// in a bad state.
-	//nolint:errcheck,gosec
-	p.forceKillGroup()
+	return p.forceKillGroup()
 }
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -34,8 +34,7 @@ type ManagedProcess interface {
 
 	// KillGroup will attempt to kill the process group and not wait for completion. Only use this if
 	// comfortable with leaking resources (in cases where exiting the program as quickly as possible is desired).
-	// Returns kill process that can be waited on if desired.
-	KillGroup() (*exec.Cmd, error)
+	KillGroup()
 
 	// Status return nil when the process is both alive and owned.
 	// If err is non-nil, process may be a) alive but not owned or b) dead.
@@ -439,7 +438,7 @@ func (p *managedProcess) Stop() error {
 }
 
 // KillGroup kills the process group.
-func (p *managedProcess) KillGroup() (*exec.Cmd, error) {
+func (p *managedProcess) KillGroup() {
 	// Minimally hold a lock here so that we can signal the
 	// management goroutine to stop. We will attempt to kill the
 	// process even if p.stopped is true.
@@ -451,7 +450,7 @@ func (p *managedProcess) KillGroup() (*exec.Cmd, error) {
 
 	if p.cmd == nil {
 		p.mu.Unlock()
-		return nil, errAlreadyStopped
+		return
 	}
 	p.mu.Unlock()
 
@@ -461,5 +460,6 @@ func (p *managedProcess) KillGroup() (*exec.Cmd, error) {
 	// without a lock held.
 	// We are intentionally not checking the error here, we are already
 	// in a bad state.
-	return p.forceKillGroup()
+	//nolint:errcheck,gosec
+	p.forceKillGroup()
 }
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -32,6 +32,10 @@ type ManagedProcess interface {
 	// there's any system level issue stopping the process.
 	Stop() error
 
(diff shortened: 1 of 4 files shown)


pexec/managed_process.go
```diff
@@ -32,6 +32,10 @@ type ManagedProcess interface {
 	// there's any system level issue stopping the process.
 	Stop() error
 
+	// Kill will attempt to kill the process group and not wait for completion. Only use this if
```
dgottlieb, line 35: What's the story behind killing the process group? Is the `viam-server` part of the same process group? If we're intentionally trying to kill the viam-server + all modules running in one swift process group `kill -9`, what's the utility in adding this functionality to the goutils/managed process code?

cheukt: the viam-server is not part of the same process group, each ManagedProcess is its own process group, I'm guessing the intention is to be able to `Kill` or `Stop` all spawned processes from one ManagedProcess at once

dgottlieb: > the viam-server is not part of the same process group, each ManagedProcess is its own process group

I thought* that was a possibility. But I didn't see the code that did that. Is that in our code or the SDK side?

cheukt: https://github.com/viamrobotics/goutils/blob/8961a203d9bc1ae41b4d750389a5f566e69a95bd/pexec/managed_process_unix.go#L69 sets the process group id, and since we didn't give the process group id a value, it will default to the child process' id https://pkg.go.dev/syscall#SysProcAttr.Setpgid

pexec/managed_process_unix.go
```diff
@@ -126,6 +127,14 @@ func (p *managedProcess) kill() (bool, error) {
 	return forceKilled, nil
 }
 
+// forceKill kills everything in the process group. This will not wait for completion and may result in a zombie process.
```
benjirewis, line 130: > a zombie process

I have heard this term used in a couple different ways. What do you mean by "zombie process"?

pexec/managed_process_unix.go
```diff
@@ -126,6 +127,14 @@ func (p *managedProcess) kill() (bool, error) {
...
+func (p *managedProcess) forceKill() error {
+	pidStr := strconv.Itoa(-p.cmd.Process.Pid)
+	p.logger.Infof("killing entire process group %d", p.cmd.Process.Pid)
```
benjirewis, line 133:
```suggestion
	p.logger.Infof("killing entire process group of process %d", p.cmd.Process.Pid)
```
ID is process ID not process group ID.

pexec/managed_process_windows.go
```diff
@@ -107,6 +106,13 @@ func (p *managedProcess) kill() (bool, error) {
 	return forceKilled, nil
 }
 
+// forceKill kills everything in the process tree. This will not wait for completion and may result in a zombie process.
+func (p *managedProcess) forceKill() error {
+	pidStr := strconv.Itoa(p.cmd.Process.Pid)
+	p.logger.Infof("force killing entire process tree %d", p.cmd.Process.Pid)
+	return exec.Command("taskkill", "/t", "/f", "/pid", pidStr).Start()
```
benjirewis, line 113: Did you test this on windows?

benjirewis (commented):
Nice! Just some initial comments.

pexec/managed_process_unix.go
cheukt: the process group id is the process id, because of https://github.com/viamrobotics/goutils/pull/399#discussion_r1901046413

pexec/managed_process_windows.go
cheukt: not yet, I just copied this from kill() a bit above

pexec/managed_process.go
```diff
@@ -32,6 +32,10 @@ type ManagedProcess interface {
...
+	// comfortable with leaking resources (in cases where exiting the program as quickly as possible is desired).
+	Kill()
```
benjirewis, line 37: Should we call this `KillGroup`?

pexec/managed_process_unix.go
benjirewis:
```suggestion
// forceKill kills everything in the process group. This will not wait for completion and may result in the kill becoming a zombie process.
```

pexec/managed_process_unix.go
```diff
@@ -126,6 +127,14 @@ func (p *managedProcess) kill() (bool, error) {
 	return forceKilled, nil
 }
 
+// forceKill kills everything in the process group. This will not wait for completion and may result in a zombie process.
+func (p *managedProcess) forceKill() error {
```
benjirewis, line 131: Should we call this `forceKillGroup`?

benjirewis (changes requested):
Could we add a test that a managed process that starts another process in the same process can both be killed by `Kill` (or `KillGroup` if you take my renaming suggestion)?

dgottlieb (approved):

pexec/managed_process.go
benjirewis, line 35:
```suggestion
	// KillGroup will attempt to kill the process group and not wait for completion. Only use this if
```
Surprised lint passes.

pexec/managed_process.go
```diff
@@ -432,3 +436,29 @@ func (p *managedProcess) Stop() error {
 	}
 	return errors.Errorf("non-successful exit code: %d", p.cmd.ProcessState.ExitCode())
 }
+
+func (p *managedProcess) KillGroup() {
```
benjirewis, line 440:
```suggestion
// KillGroup kills the process group.
func (p *managedProcess) KillGroup() {
```
For lint.

pexec/managed_process_unix.go
```diff
@@ -126,6 +127,15 @@ func (p *managedProcess) kill() (bool, error) {
 	return forceKilled, nil
 }
 
+// forceKill kills everything in the process group. This will not wait for completion and may result the
```
benjirewis, line 130:
```suggestion
// forceKillGroup kills everything in the process group. This will not wait for completion and may result the
```

pexec/managed_process_windows.go
```diff
@@ -107,6 +106,13 @@ func (p *managedProcess) kill() (bool, error) {
 	return forceKilled, nil
 }
 
+// forceKill kills everything in the process tree. This will not wait for completion and may result in a zombie process.
```
benjirewis, line 109:
```suggestion
// forceKillGroup kills everything in the process tree. This will not wait for completion and may result in a zombie process.
```

benjirewis (approved):

cheukt:
you know, I'm surprised that it passes too

PR #401, opened 2025-01-09
RSDK-9714: Call WaitGroup.Done in too many in-flight requests error path.

dgottlieb:


dgottlieb:
I don't have the application logs/FTDC data that would definitively say whether "too many in-flight requests" happened, but there was a case of wrtc server shutdown waiting for that waitgroup to drain, and there were no goroutines inside of `processHeaders` beyond the `.Add(1)` step.

PR #402, opened 2025-01-14
[RSDK-8666] Use StoppableWorkers in WebRTC Signaling Process

bashar-515:
_**WIP**_

_Not done yet. Do not approve/merge._

### DIFF
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -100,7 +100,7 @@ func (ans *webrtcSignalingAnswerer) Start() {
 	ans.sw.Add(func(ctx context.Context) {
 		for ans.conn == nil {
 			select {
-			case <-ans.closeCtx.Done():
+			case <-ctx.Done():
 				return
 			default:
 			}
@@ -111,12 +111,12 @@ func (ans *webrtcSignalingAnswerer) Start() {
 			if proxyAddr := os.Getenv(SocksProxyEnvVar); proxyAddr != "" {
 				timeout = answererConnectTimeoutBehindProxy
 			}
-			setupCtx, timeoutCancel := context.WithTimeout(ans.closeCtx, timeout)
+			setupCtx, timeoutCancel := context.WithTimeout(ctx, timeout)
 			conn, err := Dial(setupCtx, ans.address, ans.logger, ans.dialOpts...)
 			timeoutCancel()
 			if err != nil {
 				ans.logger.Errorw("error connecting answer client", "error", err)
-				if !utils.SelectContextOrWait(ans.closeCtx, answererReconnectWait) {
+				if !utils.SelectContextOrWait(ctx, answererReconnectWait) {
 					return
 				}
 				continue
@@ -178,7 +178,7 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 		}()
 		for {
 			select {
-			case <-ans.closeCtx.Done():
+			case <-ctx.Done():
 				return
 			default:
 			}
@@ -189,7 +189,7 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 			if err != nil {
 				if checkExceptionalError(err) != nil {
 					ans.logger.Warnw("error communicating with signaling server", "error", err)
-					if !utils.SelectContextOrWait(ans.closeCtx, answererReconnectWait) {
+					if !utils.SelectContextOrWait(ctx, answererReconnectWait) {
 						return
 					}
 				}
@@ -221,7 +221,7 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 			if err != nil {
 				if checkExceptionalError(err) != nil {
 					ans.logger.Warnw("error communicating with signaling server", "error", err)
-					if !utils.SelectContextOrWait(ans.closeCtx, answererReconnectWait) {
+					if !utils.SelectContextOrWait(ctx, answererReconnectWait) {
 						return
 					}
 				}
@@ -253,16 +253,16 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 			var answerCtx context.Context
 			var answerCtxCancel func()
 			if deadline := initStage.Init.GetDeadline(); deadline != nil {
-				answerCtx, answerCtxCancel = context.WithDeadline(aa.closeCtx, deadline.AsTime())
+				answerCtx, answerCtxCancel = context.WithDeadline(ctx, deadline.AsTime())
 			} else {
-				answerCtx, answerCtxCancel = context.WithTimeout(aa.closeCtx, getDefaultOfferDeadline())
+				answerCtx, answerCtxCancel = context.WithTimeout(ctx, getDefaultOfferDeadline())
 			}
 
 			if err = aa.connect(answerCtx); err != nil {
 				answerCtxCancel()
 				// We received an error while trying to connect to a caller/peer.
 				ans.logger.Errorw("error connecting to peer", "error", err)
-				if !utils.SelectContextOrWait(ans.closeCtx, answererReconnectWait) {
+				if !utils.SelectContextOrWait(ctx, answererReconnectWait) {
 					return
 				}
 			}
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -148,6 +148,7 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 		md := metadata.New(nil)
 		md.Append(RPCHostMetadataField, ans.hosts...)
 		md.Append(HeartbeatsAllowedMetadataField, "true")
+		// use StoppableWorkers.Context() so that instantiation of answer client responds to StoppableWorkers.Stop()
 		answerCtx := metadata.NewOutgoingContext(ans.bgWorkers.Context(), md)
 		answerClient, err := client.Answer(answerCtx)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


rpc/wrtc_signaling_answerer.go
```diff
@@ -67,16 +62,17 @@ func newWebRTCSignalingAnswerer(
...
-		cancelBgWorkers: cancel,
-		closeCtx:        closeCtx,
-		logger:          logger,
+		address:      address,
+		hosts:        hosts,
+		server:       server,
+		dialOpts:     dialOptsCopy,
+		webrtcConfig: webrtcConfig,
+		sw:           sw,
+		closeCtx:     closeCtx,
```
bashar-515, line 74: Had to maintain a context here because it gets used [here](https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_signaling_answerer.go#L163) (outside of a worker that takes in a context as a parameter :/)

rpc/wrtc_signaling_answerer.go
```diff
@@ -93,11 +96,8 @@ func (ans *webrtcSignalingAnswerer) Start() {
 	ans.startStopMu.Lock()
 	defer ans.startStopMu.Unlock()
 
-	// No lock is necessary here. It is illegal to call `ans.Stop` before `ans.Start` returns.
-	ans.bgWorkers.Add(1)
-
 	// attempt to make connection in a loop
-	utils.ManagedGo(func() {
+	ans.sw.Add(func(ctx context.Context) {
```
bashar-515, line 100: [qu] this worker now _does_ utilize a lock. Is this an unnecessary overhead, or is it fine being the case?

rpc/wrtc_signaling_answerer.go
```diff
@@ -93,11 +96,8 @@ func (ans *webrtcSignalingAnswerer) Start() {
 	ans.startStopMu.Lock()
 	defer ans.startStopMu.Unlock()
 
-	// No lock is necessary here. It is illegal to call `ans.Stop` before `ans.Start` returns.
```
bashar-515, line 96: [qu] this worker now _does_ utilize a lock since its managed by a `StoppableWorkers` instance. Is this an unnecessary overhead, or is it fine being the case?

rpc/wrtc_signaling_answerer.go
```diff
@@ -68,12 +69,14 @@ func newWebRTCSignalingAnswerer(
...
+	sw := utils.NewStoppableWorkers(closeCtx)
 	return &webrtcSignalingAnswerer{
 		address:         address,
 		hosts:           hosts,
 		server:          server,
 		dialOpts:        dialOptsCopy,
 		webrtcConfig:    webrtcConfig,
+		sw:              sw,
 		cancelBgWorkers: cancel,
 		closeCtx:        closeCtx,
```
bashar-515, line 81: [qu] we have to continue storing this context because it's used [here](https://github.com/viamrobotics/goutils/blob/main/rpc/wrtc_signaling_answerer.go#L163), _not_ inside of a worker managed by `sw`. We could opt to use the `NewBackgroundStoppableWorkers` instantiater and _not_ store the context, but then we'd have to call `.Context()` on the stoppable workers instance in the previously linked line of code instead of passing the stored context. Do you have any thought/preferences here @benjirewis?

rpc/wrtc_signaling_answerer.go
```diff
@@ -300,12 +279,18 @@ func (ans *webrtcSignalingAnswerer) Stop() {
...
+	// TODO(Bashar): investigate deadlock
+
 	ans.bgWorkersMu.Lock()
 	ans.cancelBgWorkers()
 	// Background workers require the `bgWorkersMu`. Release the mutex before calling `Wait`.
 	ans.bgWorkersMu.Unlock()
 	ans.bgWorkers.Wait()
 
+	// TODO(Bashar)
+	// ans.sw.Stop()
```
bashar-515, line 292: This is causing a deadlock.

rpc/wrtc_signaling_answerer.go
```diff
@@ -160,32 +158,15 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 		md := metadata.New(nil)
 		md.Append(RPCHostMetadataField, ans.hosts...)
 		md.Append(HeartbeatsAllowedMetadataField, "true")
-		answerCtx := metadata.NewOutgoingContext(ans.closeCtx, md)
+		answerCtx := metadata.NewOutgoingContext(ans.closeCtx, md) // TODO(Bashar): can we use a different context here
```
bashar-515, line 161: [qu]: see TODO comment.

rpc/wrtc_signaling_answerer.go
```diff
@@ -300,12 +279,18 @@ func (ans *webrtcSignalingAnswerer) Stop() {
 	// Code adding workers must atomically check the `closeCtx` before adding to the `bgWorkers`
 	// wait group. Canceling the context must not split those two operations. We ensure this
 	// atomicity by acquiring the `bgWorkersMu` write lock.
+
+	// TODO(Bashar): investigate deadlock
```
bashar-515, line 283: The stoppable workers `Stop()` function causes a a deadlock since it calls its own `cancel` function. I created a new Stop() function identical to it but with a cancel function passed in as a parameter to use instead of the sw instance's internal one, and that seemed to do the trick when using it and passing ans.cancelBgWorkers. Does that make sense? Do you know why this might be happening? @benjirewis

rpc/wrtc_signaling_answerer.go
```diff
@@ -38,17 +38,18 @@ type webrtcSignalingAnswerer struct {
 	// bgWorkersMu should be `Lock`ed in `Stop` to `Wait` on ongoing background workers in startAnswerer/answer.
 	// bgWorkersMu should be `RLock`ed when starting a new background worker (allow concurrent background workers) to
 	// `Add` to bgWorkers.
-	bgWorkersMu     sync.RWMutex
-	bgWorkers       sync.WaitGroup
-	cancelBgWorkers func()
+	bgWorkersMu     sync.RWMutex   // TODO(Bashar): remove after removing deadlock
+	bgWorkers       sync.WaitGroup // TODO(Bashar): remove after removing deadlock
+	cancelBgWorkers func()         // TODO(Bashar): remove after removing deadlock
+	sw              *utils.StoppableWorkers
```
benjirewis, line 44: Let's still call this `bgWorkers`. FWIW, in golang, we generally use short variables names like `sw` only when that variable is local/has a very limited scope over which it is not difficult to interpret what it does. When it's a field on a struct, we can afford to encode a bit more information in the name.

rpc/wrtc_signaling_answerer.go
benjirewis: Great question; the place you're referencing is a good spot to use the `.Context()` method. We don't want to use `context.Background()` because we want the instantiation of the answer client to also be governed by the context that is sensitive to answerer `Stop`ping (in other words, instantiation should halt as soon as a `Stop` comes in.)

rpc/wrtc_signaling_answerer.go
benjirewis: Another good question; it should be fine. Let's discuss offline about why that might be.

rpc/wrtc_signaling_answerer.go
benjirewis: See my other comment; use `sw.Context()`.

rpc/wrtc_signaling_answerer.go
benjirewis: I'm having trouble interpreting the entirety of your comment here, but let's discuss offline.

rpc/wrtc_signaling_answerer.go
bashar-515: I actually think this is what was causing the deadlock! I had only used a `context.Background()` and the stored `closeCtx` (which doesn't respond to `sw.Stop()`) here. Just tried it using `sw.Context()` and all good now :)

rpc/wrtc_signaling_answerer.go
```diff
@@ -160,32 +148,15 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 		md := metadata.New(nil)
 		md.Append(RPCHostMetadataField, ans.hosts...)
 		md.Append(HeartbeatsAllowedMetadataField, "true")
-		answerCtx := metadata.NewOutgoingContext(ans.closeCtx, md)
+		answerCtx := metadata.NewOutgoingContext(ans.bgWorkers.Context(), md)
```
benjirewis, line 151: [nit] Can we add a small comment above this clarifying why we're using `bgWorkers.Context()`?

benjirewis (commented):
Nice! Much simpler design. Would you mind also looking at `rpc/wrtc_server.go`, `rpc/wrtc_call_queue_memory.go`, and `rpc/wrtc_call_queue_mongodb.go` and adding similar changes? They all use waitgroups and contexts similarly.

dgottlieb (approved):
lgtm pending the StoppableWorkers documentation refinement. None of my other comments are necessary, but I do think they are nice.

dgottlieb:
Oh, let's definitely run RDK tests on these changes. Probably worth doing a quick check locally before merging.

Then post-merge, immediately create a tag and start an RDK PR to run the tests in our CI (which has a knack for finding bugs that won't repro easily locally!)

benjirewis (approved):
LGTM! Nice job :)

bashar-515:
> Oh, let's definitely run RDK tests on these changes. Probably worth doing a quick check locally before merging.
> 
> Then post-merge, immediately create a tag and start an RDK PR to run the tests in our CI (which has a knack for finding bugs that won't repro easily locally!)

local tests look good

PR #405, opened 2025-01-16
[RSDK 9149] Use RDK Logger Across Interceptors

bashar-515:
Stop converting the logger into a zap logger and instead use the native zap compatible logger. See pics below for visible change (the error log in between the highlighted lines is the same as what's referenced in the [ticket](https://viam.atlassian.net/browse/RSDK-9149)).

**Note**
re: all the 'from here' comments..
To no longer use an ordinary zap logger, I replaced calls to grpc interceptor methods with calls to copies of those methods that I modify to use an RDK logger instead. An effect of this change was the additional copying over of all private types employed by these functions in the grpc library we use,, is there a better way about this? It seems like [this](https://github.com/viamrobotics/goutils/pull/375) PR used a similar tactic to get client interceptors to use the preferred logger.

Before:
Screenshot 2025-01-08 at 11 01 25 AM

~After~: notice here that the logger (from goutils/logger.go)is outputting logs that are _not_ ERROR level or higher. This is not desired behavior. This has since been fixed; the original behavior has been restored.
Screenshot 2025-01-08 at 11 03 04 AM

After:
Screenshot 2025-01-09 at 1 22 26 PM

**Testing**
Modify the the [module.go](https://github.com/viamrobotics/rdk/blob/main/examples/customresources/demos/simplemodule/module.go) RDK file in the simple module example custom resource so that the DoCommand() implementation always returns an error. Then, follow the instructions provided in the custom resources [README.md](https://github.com/viamrobotics/rdk/blob/main/examples/customresources/README.md) file to see the logs output by the robot server.

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -20,14 +20,11 @@ import (
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/google/uuid"
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
-	grpc_zap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
 	grpc_recovery "github.com/grpc-ecosystem/go-grpc-middleware/recovery"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/improbable-eng/grpc-web/go/grpcweb"
 	"github.com/pkg/errors"
 	"go.uber.org/multierr"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 	"golang.org/x/net/http2/h2c"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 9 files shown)


rpc/server_options.go
bashar-515, line 1: @dgottlieb @cheukt This, in addition to [this](https://github.com/viamrobotics/rdk/pull/4723), are the changes to stop the leaked goroutine in RDK testing from happening. The RDK PR cannot merge until this goutils PR does and the goutils version is bumped in RDK. Once this merges and goutils in RDK is bumped, the RDK tests will still fail (leak takes place) until the other PR merges.

rpc/server.go
```diff
@@ -404,6 +397,8 @@ func NewServer(logger utils.ZapCompatibleLogger, opts ...ServerOption) (Server,
 		serverOpts = append(serverOpts, grpc.StatsHandler(sOpts.statsHandler))
 	}
 
+	serverOpts = append(serverOpts, grpc.WaitForHandlers(sOpts.waitForHandlers))
```
cheukt, line 400: @dgottlieb another option is to just have `grpcServer` call `GracefulStop()` instead of `Stop()` in [L864](https://github.com/viamrobotics/goutils/blob/d2f65e2ad6b070fb5395d35b7af7e1b727971632/rpc/server.go#L864). It is a riskier change but also not sure why that Stop() wasn't graceful, when all of the other stops in the function are.

dgottlieb: Yeah, `GracefulStop` works too. I have no strong opinions on this. Just confirming: I expect all servers to be constructed from this `NewServer` call. Given that, I don't see any behavior difference between using an option vs GracefulStop.

cheukt (approved):

PR #408, opened 2025-01-22
[RSDK-8666] Use Stoppable Workers

bashar-515:
_**WIP**_

This PR is a continuation of #402 re: [this](https://github.com/viamrobotics/goutils/pull/402#pullrequestreview-2553454718) comment from @benjirewis (i.e., it updates `rpc/wrtc_server.go`, `rpc/wrtc_call_queue_memory.go`, and `rpc/wrtc_call_queue_mongodb.go` to use `StoppableWorkers` - and some other files as a side effect).

- [x] rpc/wrtc_server.go
- [x] rpc/wrtc_call_queue_memory.go
- [x] rpc/wrtc_call_queue_mongodb.go

### DIFF
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -1,5 +1,6 @@
-// TODO(RSDK-8666): use stoppable workers
 package rpc
+
+// TODO(RSDK-8666): use stoppable workers
 
 import (
 	"context"
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -114,7 +114,7 @@ func (queue *memoryWebRTCCallQueue) SendOfferInit(
 	hostQueueForSend.activeOffers[offer.uuid] = exchange
 	hostQueueForSend.mu.Unlock()
 
-	queue.activeBackgroundWorkers.Add(func(ctx context.Context) {
+	queue.activeBackgroundWorkers.Add(func(_ context.Context) {
 		select {
 		case <-sendCtx.Done():
 		case <-ctx.Done():
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -49,6 +49,7 @@ func newMemoryWebRTCCallQueue(uuidDeterministic bool, logger utils.ZapCompatible
 				return
 			}
 			now := time.Now()
+			queue.mu.Lock()
 			for _, hostQueue := range queue.hostQueues {
 				hostQueue.mu.Lock()
 				for offerID, offer := range hostQueue.activeOffers {
@@ -58,6 +59,7 @@ func newMemoryWebRTCCallQueue(uuidDeterministic bool, logger utils.ZapCompatible
 				}
 				hostQueue.mu.Unlock()
 			}
+			queue.mu.Unlock()
 		}
 	})
 	return queue
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -44,9 +44,6 @@ func newMemoryWebRTCCallQueue(uuidDeterministic bool, logger utils.ZapCompatible
 		logger:            logger,
 	}
 	queue.activeBackgroundWorkers = utils.NewStoppableWorkerWithTicker(5*time.Second, func(ctx context.Context) {
-		if ctx.Err() != nil {
-			return
-		}
 		now := time.Now()
 		queue.mu.Lock()
 		for _, hostQueue := range queue.hostQueues {
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -128,9 +128,7 @@ func (queue *memoryWebRTCCallQueue) SendOfferInit(
 	hostQueueForSend.activeOffers[offer.uuid] = exchange
 	hostQueueForSend.mu.Unlock()
 
-	queue.activeBackgroundWorkers.Add(1)
-	utils.PanicCapturingGo(func() {
-		queue.activeBackgroundWorkers.Done()
+	queue.activeStoppableWorkers.Add(func(ctx context.Context) {
 		select {
 		case <-sendCtx.Done():
 		case <-ctx.Done():
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -19,9 +19,8 @@ import (
 // A memoryWebRTCCallQueue is an in-memory implementation of a call queue designed to be used for
 // testing and single node/host deployments.
 type memoryWebRTCCallQueue struct {
-	activeStoppableWorkers  *utils.StoppableWorkers
 	mu                      sync.Mutex
-	activeBackgroundWorkers sync.WaitGroup
+	activeBackgroundWorkers *utils.StoppableWorkers
 	hostQueues              map[string]*singleWebRTCHostQueue
 
 	cancelCtx  context.Context
@@ -52,7 +51,7 @@ func newMemoryWebRTCCallQueue(uuidDeterministic bool, logger utils.ZapCompatible
 		uuidDeterministic: uuidDeterministic,
 		logger:            logger,
 	}
-	queue.activeStoppableWorkers = utils.NewStoppableWorkerWithTicker(5*time.Second, func(ctx context.Context) {
+	queue.activeBackgroundWorkers = utils.NewStoppableWorkerWithTicker(5*time.Second, func(ctx context.Context) {
 		for {
 			if cancelCtx.Err() != nil {
 				return
@@ -128,7 +127,7 @@ func (queue *memoryWebRTCCallQueue) SendOfferInit(
 	hostQueueForSend.activeOffers[offer.uuid] = exchange
 	hostQueueForSend.mu.Unlock()
 
-	queue.activeStoppableWorkers.Add(func(ctx context.Context) {
+	queue.activeBackgroundWorkers.Add(func(ctx context.Context) {
 		select {
 		case <-sendCtx.Done():
 		case <-ctx.Done():
@@ -220,8 +219,7 @@ func (queue *memoryWebRTCCallQueue) RecvOffer(ctx context.Context, hosts []strin
 // Close cancels all active offers and waits to cleanly close all background workers.
 func (queue *memoryWebRTCCallQueue) Close() error {
 	queue.cancelFunc()
-	queue.activeBackgroundWorkers.Wait()
-	queue.activeStoppableWorkers.Stop()
+	queue.activeBackgroundWorkers.Stop()
 	return nil
 }
 
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -49,7 +49,6 @@ func newMemoryWebRTCCallQueue(uuidDeterministic bool, logger utils.ZapCompatible
 				return
 			}
 			now := time.Now()
-			queue.mu.Lock()
 			for _, hostQueue := range queue.hostQueues {
 				hostQueue.mu.Lock()
 				for offerID, offer := range hostQueue.activeOffers {
@@ -59,7 +58,6 @@ func newMemoryWebRTCCallQueue(uuidDeterministic bool, logger utils.ZapCompatible
 				}
 				hostQueue.mu.Unlock()
 			}
-			queue.mu.Unlock()
 		}
 	})
 	return queue
--- a/rpc/wrtc_call_queue_memory.go
+++ b/rpc/wrtc_call_queue_memory.go
@@ -44,23 +44,21 @@ func newMemoryWebRTCCallQueue(uuidDeterministic bool, logger utils.ZapCompatible
 		logger:            logger,
 	}
 	queue.activeBackgroundWorkers = utils.NewStoppableWorkerWithTicker(5*time.Second, func(ctx context.Context) {
-		for {
-			if ctx.Err() != nil {
-				return
-			}
-			now := time.Now()
-			queue.mu.Lock()
-			for _, hostQueue := range queue.hostQueues {
-				hostQueue.mu.Lock()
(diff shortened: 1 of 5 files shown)


rpc/wrtc_call_queue_memory.go
```diff
@@ -41,28 +38,17 @@ func newMemoryWebRTCCallQueueTest(logger utils.ZapCompatibleLogger) *memoryWebRT
...
-	ticker := time.NewTicker(5 * time.Second)
-	utils.ManagedGo(func() {
+	queue.activeBackgroundWorkers = utils.NewStoppableWorkerWithTicker(5*time.Second, func(ctx context.Context) {
 		for {
-			if cancelCtx.Err() != nil {
-				return
-			}
-			select {
-			case <-cancelCtx.Done():
+			if ctx.Err() != nil {
```
bashar-515, line 48: Removing this check results in a deadlock/hang when running tests in the `rpc` package.

rpc/wrtc_server_channel.go
```diff
@@ -37,7 +37,7 @@ func newWebRTCServerChannel(
 	logger utils.ZapCompatibleLogger,
 ) *webrtcServerChannel {
 	base := newBaseChannel(
-		server.ctx,
+		server.processHeadersWorkers.Context(),
```
bashar-515, line 40: An artifact of no longer storing a `context.Context` inside of the `webrtcServer` struct.

rpc/wrtc_server_stream.go
```diff
@@ -269,30 +269,24 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
...
-	s.ch.server.processHeadersMu.RLock()
-	s.ch.server.processHeadersWorkers.Add(1)
-	s.ch.server.processHeadersMu.RUnlock()
 
 	// Check if context has errored: underlying server may have been `Stop`ped,
-	// in which case we mark this processHeaders worker as `Done` and return.
-	if err := s.ch.server.ctx.Err(); err != nil {
-		s.ch.server.processHeadersWorkers.Done()
+	// in which case we return.
+	if err := s.ch.server.processHeadersWorkers.Context().Err(); err != nil {
```
bashar-515, line 275: Also an artifact of removing the `webrtcServer` struct's `context.Context`.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -200,30 +202,32 @@ func NewMongoDBWebRTCCallQueue(
...
+		logger:                 utils.AddFieldsToLogger(logger, "operator_id", operatorID),
 
 		csStateUpdates:        make(chan changeStreamStateUpdate),
 		callExchangeSubs:      map[string]map[*mongodbCallExchange]struct{}{},
 		waitingForNewCallSubs: map[string]map[*mongodbNewCallEventHandler]struct{}{},
 		activeAnswerersfunc:   &activeAnswerersfunc,
 	}
 
-	queue.activeBackgroundWorkers.Add(2)
+	// TODO(RSDK-8666): using StoppableWorkers is causing a data race
```
bashar-515, line 218: :/

rpc/wrtc_call_queue_memory.go
```diff
@@ -41,25 +38,15 @@ func newMemoryWebRTCCallQueueTest(logger utils.ZapCompatibleLogger) *memoryWebRT
...
-		cancelCtx:         cancelCtx,
-		cancelFunc:        cancelFunc,
 		uuidDeterministic: uuidDeterministic,
 		logger:            logger,
 	}
-	queue.activeBackgroundWorkers.Add(1)
-	ticker := time.NewTicker(5 * time.Second)
-	utils.ManagedGo(func() {
+	queue.activeBackgroundWorkers = utils.NewStoppableWorkerWithTicker(5*time.Second, func(ctx context.Context) {
 		for {
```
benjirewis, line 47: I would _think_ that the `for` loop here is no longer necessary given the logic within it is now being called every 5s. I.e., the looping is handled by the stoppable workers objects itself?

rpc/wrtc_call_queue_memory.go
benjirewis: Yeah we'll still want to check context. This looks correct. Ticker case should be removed as you've done, too. Just see my comment about not needing the for loop surrounding this at all.

rpc/wrtc_call_queue_mongodb.go
benjirewis: The liveness loop is racing? Could you follow up with me offline about what the race looks like?

rpc/wrtc_server_channel.go
benjirewis: Gotcha; this makes sense to me. There's a hierarchy here of WebRTC server owns WebRTC server channels owns WebRTC server streams. We want the context cancelations to propagate in that order. So, I think what you have here is correct. Once `Stop()` is called on the WebRTC server, we _should_ cancel all the way down.

rpc/wrtc_server_stream.go
```diff
@@ -269,30 +269,24 @@ func (s *webrtcServerStream) processHeaders(headers *webrtcpb.RequestHeaders) {
...
 	select {
 	case s.ch.server.callTickets <- struct{}{}:
 	default:
-		s.ch.server.processHeadersWorkers.Done()
 		s.closeWithSendError(status.Error(codes.ResourceExhausted, "too many in-flight requests"))
 		return
 	}
 
 	s.headersReceived = true
 	utils.PanicCapturingGo(func() {
```
benjirewis, line 288: I think this should probably be a `s.ch.server.processHeadersWorkers.Add(` call instead of a `PanicCapturingGo` call.

benjirewis (commented):
Nice! These are definitely harder than the others. We can chat offline about the data race/general design here.

rpc/wrtc_call_queue_mongodb.go
benjirewis, line 1: This file is pretty critical to the functioning of distributed signaling for Viam. I'm ok not even making changes to it, since we haven't yet seen any issues in its usage of wait groups/contexts. If you and @dgottlieb want to get stoppable workers into this file, though, then I think we should test a local version of app + MongoDB to ensure signaling still works. Can also probably do it with an app PR that creates a temporary staging environment.

rpc/wrtc_server_stream.go
benjirewis: @dgottlieb do you remember why we need this context check _before_ even taking a ticket?

benjirewis: Cool; that's my thought, too. Wanna remove this whole if block @bashar-515 ?

rpc/wrtc_call_queue_memory.go
```diff
@@ -41,42 +38,27 @@ func newMemoryWebRTCCallQueueTest(logger utils.ZapCompatibleLogger) *memoryWebRT
...
-			now := time.Now()
-			queue.mu.Lock()
-			for _, hostQueue := range queue.hostQueues {
-				hostQueue.mu.Lock()
-				for offerID, offer := range hostQueue.activeOffers {
-					if d, ok := offer.offer.answererDoneCtx.Deadline(); ok && d.Before(now) {
-						delete(hostQueue.activeOffers, offerID)
-					}
+	queue.activeBackgroundWorkers = utils.NewStoppableWorkerWithTicker(5*time.Second, func(ctx context.Context) {
+		if ctx.Err() != nil {
```
dgottlieb, line 47: Before, this was necessary to exit the loop. But this is no longer serving a purpose. The stoppable workers caller checks the state/context before calling this method.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -344,7 +339,7 @@ func (queue *mongoDBWebRTCCallQueue) operatorLivenessLoop() {
 	ticker := time.NewTicker(operatorStateUpdateInterval)
 	defer ticker.Stop()
 	for {
-		if !utils.SelectContextOrWaitChan(queue.cancelCtx, ticker.C) {
+		if !utils.SelectContextOrWaitChan(queue.activeStoppableWorkers.Context(), ticker.C) {
```
dgottlieb, line 342: As per the comment on having `operatorLivenessLoop` accepting a `ctx`, this is the line that caught my attention. We should just be checking the input `context` and not need to do this lookup on the stoppable workers itself.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -390,15 +385,16 @@ func (queue *mongoDBWebRTCCallQueue) operatorLivenessLoop() {
 			})
 		}
 
-		if _, err := queue.operatorsColl.UpdateOne(queue.cancelCtx, bson.D{{webrtcOperatorIDField, queue.operatorID}}, bson.D{
-			{
-				"$set",
-				bson.D{
-					{webrtcOperatorExpireAtField, time.Now().Add(operatorHeartbeatWindow)},
-					{webrtcOperatorHostsField, hostSizes},
+		if _, err := queue.operatorsColl.UpdateOne(queue.activeStoppableWorkers.Context(), bson.D{{webrtcOperatorIDField, queue.operatorID}},
```
dgottlieb, line 388: Same remark about using an input `ctx`.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -427,7 +423,7 @@ func (queue *mongoDBWebRTCCallQueue) changeStreamManager() {
 	for {
 		// Note(erd): this could use condition variables instead in order to be efficient about
 		// change stream restarts, but it does not feel worth the complexity right now :o)
-		if !utils.SelectContextOrWaitChan(queue.cancelCtx, ticker.C) {
+		if !utils.SelectContextOrWaitChan(queue.activeStoppableWorkers.Context(), ticker.C) {
```
dgottlieb, line 426: Same remark about `changeStreamManager` being a function that accepts a context.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -460,7 +456,7 @@ func (queue *mongoDBWebRTCCallQueue) changeStreamManager() {
 
 		// note(roxy): this is updating the changestream based on whether there is a new
 		// answerer that is coming online or if there is a new caller that is coming online
-		cs, err := queue.callsColl.Watch(queue.cancelCtx, []bson.D{
+		cs, err := queue.callsColl.Watch(queue.activeStoppableWorkers.Context(), []bson.D{
```
dgottlieb, line 459: ctx

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -500,11 +496,11 @@ func (queue *mongoDBWebRTCCallQueue) changeStreamManager() {
 		queue.csStateMu.Unlock()
 		activeHosts.Set(queue.operatorID, int64(len(hosts)))
 
-		nextCSCtx, nextCSCtxCancel := context.WithCancel(queue.cancelCtx)
+		nextCSCtx, nextCSCtxCancel := context.WithCancel(queue.activeStoppableWorkers.Context())
```
dgottlieb, line 499: ctx

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -235,11 +230,11 @@ func NewMongoDBWebRTCCallQueue(
 			})
 			queue.subscriptionManager(newState.ChangeStream)
```
dgottlieb, line 231: Also pass the `ctx` into `subscriptionManager`

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -677,14 +673,14 @@ func (queue *mongoDBWebRTCCallQueue) processNextSubscriptionEvent(next mongoutil
 func (queue *mongoDBWebRTCCallQueue) subscriptionManager(currentCS <-chan mongoutils.ChangeEventResult) {
 	var waitForNextCS bool
 	for {
-		if queue.cancelCtx.Err() != nil {
+		if queue.activeStoppableWorkers.Context().Err() != nil {
```
dgottlieb, line 676: As per above, if we add in the `ctx` to this call, we can swap out accessing `activeStoppableWorkers` here.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -889,7 +885,7 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
 	sendCtx, sendCtxCancel := context.WithDeadline(ctx, offerDeadline)
 
 	// need to watch before insertion to avoid a race
-	sendAndQueueCtx, sendAndQueueCtxCancel := utils.MergeContext(sendCtx, queue.cancelCtx)
+	sendAndQueueCtx, sendAndQueueCtxCancel := utils.MergeContext(sendCtx, queue.activeStoppableWorkers.Context())
```
dgottlieb, line 888: I suppose we can keep this to minimize the changes as part of this refactor. But this feels clunky. We're just merging here to reduce the number of cases inside the below selects by one.

And I think there are only two select statements.

rpc/wrtc_server_channel.go
dgottlieb: Agreed this will do the right thing. It would be more customary to just have `newWebRTCServerChannel` accept a context to bound its lifetime, but no need to change how things are structured right now.

rpc/wrtc_call_queue_memory.go
```diff
@@ -135,9 +117,7 @@ func (queue *memoryWebRTCCallQueue) SendOfferInit(
 	hostQueueForSend.activeOffers[offer.uuid] = exchange
 	hostQueueForSend.mu.Unlock()
 
-	queue.activeBackgroundWorkers.Add(1)
-	utils.PanicCapturingGo(func() {
-		queue.activeBackgroundWorkers.Done()
+	queue.activeBackgroundWorkers.Add(func(ctx context.Context) {
 		select {
 		case <-sendCtx.Done():
 		case <-ctx.Done():
```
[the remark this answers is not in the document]
bashar-515: nice catch totally missed this

dgottlieb (approved):

PR #410, opened 2025-01-30
[RSDK-8293] Pass in Connection to Answerer in WebRTC Server

bashar-515:
This PR allows for a preexisting connection to be passed to the constructor for our WebRTC signaling answerer to use via a dial option.

**Testing**

4 cases:

@cheukt and I ran the first tests by running viam-server on my laptop while connected to the Viam-Guest WiFi. He ran a client Go script on his laptop connected to the Viam WiFi. The client was able to connect to the robot via WebRTC both times.

1. client on different network can connect if rdk starts up offline then online (force webrtc option)
2. client on different network can connect if rdk starts up online then goes offline then online (force webrtc option)
3. client on same network can connect if rdk starts up offline then online (force webrtc option, but using the internal signaler)
offline-online-internal
4. client on same network can connect if rdk starts up online then goes offline then online (force webrtc option, but using the internal signaler)
online-offline-internal

### DIFF
--- a/rpc/dial_options.go
+++ b/rpc/dial_options.go
@@ -66,7 +66,7 @@ type dialOptions struct {
 	streamInterceptor grpc.StreamClientInterceptor
 
 	// conn can be used to force the webrtcSignalingAnswerer to use a preexisting connection instead of dialing and managing its own.
-	externalSignalerConn ClientConn
+	externalSignalingConn ClientConn
 }
 
 // DialMulticastDNSOptions dictate any special settings to apply while dialing via mDNS.
@@ -284,9 +284,9 @@ func WithForceDirectGRPC() DialOption {
 	})
 }
 
-// WithConn provides a preexisting connection to use. This option forces the webrtcSignalingAnswerer to not dial or manage a connection.
-func WithConn(externalSignalerConn ClientConn) DialOption {
+// WithExternalSignalingConn provides a preexisting connection to use. This option forces the webrtcSignalingAnswerer to not dial or manage a connection.
+func WithExternalSignalingConn(externalSignalingConn ClientConn) DialOption {
 	return newFuncDialOption(func(o *dialOptions) {
-		o.externalSignalerConn = externalSignalerConn
+		o.externalSignalingConn = externalSignalingConn
 	})
 }
--- a/rpc/dial_options.go
+++ b/rpc/dial_options.go
@@ -66,7 +66,7 @@ type dialOptions struct {
 	streamInterceptor grpc.StreamClientInterceptor
 
 	// conn can be used to force the webrtcSignalingAnswerer to use a preexisting connection instead of dialing and managing its own.
-	conn ClientConn
+	externalSignalerConn ClientConn
 }
 
 // DialMulticastDNSOptions dictate any special settings to apply while dialing via mDNS.
@@ -285,8 +285,8 @@ func WithForceDirectGRPC() DialOption {
 }
 
 // WithConn provides a preexisting connection to use. This option forces the webrtcSignalingAnswerer to not dial or manage a connection.
-func WithConn(conn ClientConn) DialOption {
+func WithConn(externalSignalerConn ClientConn) DialOption {
 	return newFuncDialOption(func(o *dialOptions) {
-		o.conn = conn
+		o.externalSignalerConn = externalSignalerConn
 	})
 }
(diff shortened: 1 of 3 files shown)


rpc/dial_options.go
```diff
@@ -280,3 +283,10 @@ func WithForceDirectGRPC() DialOption {
 		o.disableDirect = false
 	})
 }
+
+// WithConn provides a preexisting connection to use. This option forces the webrtcSignalingAnswerer to not dial or manage a connection.
+func WithConn(conn ClientConn) DialOption {
```
cheukt, line 288: maybe ExternalSignalerConn? we should be clear that this would be used for the external signaler answerer

rpc/wrtc_signaling_answerer.go
```diff
@@ -71,6 +76,11 @@ func newWebRTCSignalingAnswerer(
 		bgWorkers:    bgWorkers,
 		logger:       logger,
 	}
+	if options.conn != nil {
```
cheukt, line 79: the internal signaler answer shouldn't be using the same connection - it is instantiated here https://github.com/viamrobotics/goutils/blob/cf6c56e1b17668c662781e7e4187b6b866ccc9f9/rpc/server.go#L659-L665

cheukt (changes requested):
some comments but looks pretty good!

rpc/wrtc_signaling_answerer.go
bashar-515: This doesn't require any change does it, since its instantiation has its own fresh set of options passed?

rpc/dial_options.go
```diff
@@ -280,3 +283,10 @@ func WithForceDirectGRPC() DialOption {
 		o.disableDirect = false
 	})
 }
+
+// WithConn provides a preexisting connection to use. This option forces the webrtcSignalingAnswerer to not dial or manage a connection.
+func WithConn(externalSignalerConn ClientConn) DialOption {
```
cheukt, line 288: rename `WithConn` as well

cheukt (commented):
this looks generally good, will hold off on approving until the corresponding RDK pr is in

rpc/wrtc_signaling_answerer.go
```diff
@@ -40,8 +40,9 @@ type webrtcSignalingAnswerer struct {
 	// conn is used to share the direct gRPC connection used by the answerer workers. As direct gRPC connections
 	// reconnect on their own, custom reconnect logic is not needed. However, keepalives are necessary for the connection
 	// to realize it's been disconnected quickly and start reconnecting.
```
cheukt, line 42: update comment here

rpc/dial_options.go
```diff
@@ -64,6 +64,9 @@ type dialOptions struct {
 	// interceptors
 	unaryInterceptor  grpc.UnaryClientInterceptor
 	streamInterceptor grpc.StreamClientInterceptor
+
+	// conn can be used to force the webrtcSignalingAnswerer to use a preexisting connection instead of dialing and managing its own.
```
cheukt, line 68:
```suggestion
	// conn can be used to force the webrtcSignalingAnswerer to use a preexisting connection instead of dialing and managing its own.
```

rpc/wrtc_signaling_answerer.go
```diff
@@ -71,6 +76,11 @@ func newWebRTCSignalingAnswerer(
 		bgWorkers:    bgWorkers,
 		logger:       logger,
 	}
+	if options.externalSignalingConn != nil {
+		ans.conn = options.externalSignalingConn
+		ans.sharedConn = true
+	}
+	return ans
```
cheukt, line 83: I know we use different opts for the internal signaling answerer, but there should be a check here to check whether this new signaling answerer is internal or external.
or we could rename the option to `signalingConn` and then explicitly call this out when we create the signaling answerers

rpc/dial_options.go
bashar-515: was this an accident?

cheukt: ya

cheukt, line 68:
```suggestion
	// signalingConn can be used to force the webrtcSignalingAnswerer to use a preexisting connection instead of dialing and managing its own.
```

cheukt (approved):
LGTM with a suggestion

PR #412, opened 2025-03-05
Add error-retrying wrapper, RetryNTimes

zaporter-work:
Talked with Benji offline -- he did not know of any wrapper that does this. I suspect we have duplicates of this littered around our codebase.

Needs tests. Will add tests after I confirm that:
a) We want to do this
b) This doesn't already exist somewhere in our codebase or in golang

I didn't put too much time into this. If we don't like this, I'll just implement retires as a one-off around my function in app.

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -24,7 +24,7 @@ func (e *RetryError) Unwrap() error {
 
 // RetryNTimesWithSleep will run `fallibleFunc` `retryAttempts` times before failing with the last error it got from the function.
 // If `retryableErrors` is supplied, only those errors will be retried.
-// It will wait for `retryDelay` between attempts
+// It will wait for `retryDelay` between attempts.
 func RetryNTimesWithSleep[T any](
 	ctx context.Context,
 	fallibleFunc func() (T, error),
--- a/retry.go
+++ b/retry.go
@@ -6,6 +6,7 @@ import (
 	"slices"
 )
 
+// RetryError is emitted by RetryNTimes if all the attempts fail. It unwraps to the last error from retrying.
 type RetryError struct {
 	inner    error
 	attempts int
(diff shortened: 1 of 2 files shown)


retry.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// RetryNTimes will run `fallibleFunc` `retryAttempts` times before failing with the last error it got from the function.
+// If `retryableErrors` is supplied, only those errors will be retried.
+func RetryNTimes[T any](fallibleFunc func() (T, error), retryAttempts int, retryableErrors ...error) (T, error) {
+	var lastError error
+
+	for numRetries := 0; numRetries < retryAttempts; numRetries++ {
+		val, err := fallibleFunc()
+		if err == nil || (len(retryableErrors) != 0 &&
```
dgottlieb, line 30: Looks like we raced on the precedence question!

retry.go
```diff
@@ -0,0 +1,39 @@
...
+	return fmt.Sprintf("failed after %d retry attempts: %v", e.attempts, e.inner)
+}
+
+func (e *RetryError) Unwrap() error {
+	return e.inner
+}
+
+// RetryNTimes will run `fallibleFunc` `retryAttempts` times before failing with the last error it got from the function.
+// If `retryableErrors` is supplied, only those errors will be retried.
+func RetryNTimes[T any](fallibleFunc func() (T, error), retryAttempts int, retryableErrors ...error) (T, error) {
```
dgottlieb, line 25: More feedback about the function itself, I'm curious about the use-case you have in mind, but I think most callers are expecting to do some kind of I/O and intend to retry on an availability problem. And in those hand-rolled cases, there's typically some kind of sleep before retrying. Like [here](https://github.com/viamrobotics/rdk/blob/923e3b900f653f9cca24995fddd68b4d201105cb/robot/web/stream/backoff.go?plain=1#L71-L74).

benjirewis (commented):
Love it; same comment about sleeping.

retry.go
zaporter-work: In my case, it is a lazy retry over unique but incrementing ID generation. I don't need any sleep, but I agree that most uses of this will want some sleep behavior. 

I've implemented your suggestion @benjirewis in this commit b7ffe8a

dgottlieb (approved):

benjirewis (approved):
Looks awesome thank you!

PR #413, opened 2025-03-10
RSDK-10165 Suppress internal RST_STREAM gRPC errors in signaling answerer code

benjirewis:


dgottlieb (commented):
I'm not sure we just want to straight up lower the log level

rpc/wrtc_signaling_answerer.go
```diff
@@ -217,7 +217,7 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 			}
 			if err != nil {
 				if isNetworkError(err) {
-					ans.logger.Warnw("error communicating with signaling server", "error", err)
+					ans.logger.Debugw("error communicating with signaling server", "error", err)
```
benjirewis, line 220: I just lowered the level of this log to Debug rather than suppress the RST_STREAM. We're already "suppressing" all "network" errors: we just warn log them and `continue`. I _could_ explicitly suppress `RST_STREAM`s from the signaling server instead.

benjirewis: Just saw your other comments; won't go this route.

dgottlieb (approved):

PR #414, opened 2025-03-10
RSDK-9990- Expose SOCKS proxy as fallback dialer

cheukt:
this change will make SOCKS proxy a fallback option (if SocksProxyEnvVar is specified), but will not actively replace a connection if it is already established. If OnlySocksProxyEnvVar is specified, only the SOCKS connection is used.

It adds a small delay to the SOCKs dialer as a way to prioritize a connection made without the SOCKS dialer.

Conceptually, pretty similar to our Dial code and it borrows heavily from the net.DialContext code.

tested a few scenarios:
* only SOCKS initially, then a network blip and turn on both phone proxy and wifi (wifi connected in the end)
* only wifi initially, then a network blip and turn on both phone proxy  and wifi (wifi connected in the end)
* wifi and SOCKS initially, then a network blip and only turn on phone proxy (SOCKS connected in the end)
* wifi and SOCKS initially, then a network blip and only turn on wifi (wifi connected in the end)

cc: @dgottlieb if interested

### DIFF
--- a/rpc/const.go
+++ b/rpc/const.go
@@ -13,4 +13,8 @@ var (
 	// proxies to indicate the address through which to route all network traffic
 	// via SOCKS5.
 	SocksProxyEnvVar = "SOCKS_PROXY"
+
+	// OnlySocksProxyEnvVar is the name of an environment variable used if all network
+	// traffic should be done through SOCKS5.
+	OnlySocksProxyEnvVar = "ONLY_SOCKS_PROXY"
 )
--- a/rpc/dialer.go
+++ b/rpc/dialer.go
@@ -282,6 +282,7 @@ func SocksProxyFallbackDialContext(
 			return conn, err
 		}
 
+		// the block below heavily references https://go.dev/src/net/dial.go#L585
 		type dialResult struct {
 			net.Conn
 			error
@@ -316,6 +317,7 @@ func SocksProxyFallbackDialContext(
 		defer primaryCancel()
 		wg.Add(1)
 		primaryDial := func(ctx context.Context) (net.Conn, error) {
+			// create a zero-valued net.Dialer to use net.Dialer's default DialContext method
 			var zeroDialer net.Dialer
 			return zeroDialer.DialContext(ctx, network, addr)
 		}
@@ -356,6 +358,8 @@ func SocksProxyFallbackDialContext(
 				} else {
 					fallback = res
 				}
+				// if both primary and fallback are done with errors, this means neither connection attempt succeeded.
+				// return the error from the primary dial attempt in that case.
 				if primary.done && fallback.done {
 					return nil, primary.error
 				}
--- a/rpc/dialer.go
+++ b/rpc/dialer.go
@@ -254,9 +254,7 @@ func socksProxyDialContext(ctx context.Context, network, proxyAddr, addr string)
 // which will allow dialers to use the default DialContext.
 // If SocksProxyEnvVar is set, it will prioritize a connection made without a proxy but will fall back to a SOCKS proxy connection.
 func SocksProxyFallbackDialContext(
-	ctx context.Context,
-	network, addr string,
-	logger utils.ZapCompatibleLogger,
+	addr string, logger utils.ZapCompatibleLogger,
 ) func(ctx context.Context, network, addr string) (net.Conn, error) {
 	// Use SOCKS proxy from environment as gRPC proxy dialer. Do not use SOCKS proxy if trying to connect to a local address.
 	localAddr := strings.HasPrefix(addr, "[::]") || strings.HasPrefix(addr, "localhost") || strings.HasPrefix(addr, "unix")
@@ -386,10 +384,10 @@ func dialDirectGRPC(ctx context.Context, address string, dOpts dialOptions, logg
 	// if the returned function is not nil.
 	//
 	// use "tcp" since gRPC uses HTTP/2, which is built on top of TCP.
-	socksProxydialContext := SocksProxyFallbackDialContext(ctx, "tcp", address, logger)
-	if socksProxydialContext != nil {
+	socksProxyDialContext := SocksProxyFallbackDialContext(address, logger)
+	if socksProxyDialContext != nil {
 		dialOpts = append(dialOpts, grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
-			return socksProxydialContext(ctx, "tcp", addr)
+			return socksProxyDialContext(ctx, "tcp", addr)
 		}))
 	}
 
--- a/rpc/dialer.go
+++ b/rpc/dialer.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/edaniels/golog"
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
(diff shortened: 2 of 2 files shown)


rpc/const.go
```diff
@@ -13,4 +13,8 @@ var (
 	// proxies to indicate the address through which to route all network traffic
 	// via SOCKS5.
 	SocksProxyEnvVar = "SOCKS_PROXY"
+
+	// OnlySocksProxyEnvVar is the name of an environment variable used if all network
+	// traffic should be done through SOCKS5.
+	OnlySocksProxyEnvVar = "ONLY_SOCKS_PROXY"
```
benjirewis, line 19: [q] Will there be an equivalent `ONLY_INTERNET`? Or, is the idea that having `SOCKS_PROXY` unset means "internet only"?

rpc/dialer.go
```diff
@@ -240,6 +241,145 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
 	return dialInner(ctx, address, logger, dOpts)
 }
 
+func socksProxyDialContext(ctx context.Context, network, proxyAddr, addr string) (net.Conn, error) {
+	dialer, err := proxy.SOCKS5(network, proxyAddr, nil, proxy.Direct)
+	if err != nil {
+		return nil, fmt.Errorf("error creating SOCKS proxy dialer to address %q from environment: %w",
+			proxyAddr, err)
+	}
+	conn, err := dialer.(proxy.ContextDialer).DialContext(ctx, network, addr)
```
benjirewis, line 250: Why do you have to cast to `proxy.ContextDialer` here? Is `DialContext` not available on the SOCKS5 dialer?

rpc/dialer.go
```diff
@@ -240,6 +241,145 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
...
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// SocksProxyFallbackDialContext will return nil if SocksProxyEnvVar is not set or if trying to connect to a local address,
+// which will allow dialers to use the default DialContext.
+// If SocksProxyEnvVar is set, it will prioritize a connection made without a proxy but will fall back to a SOCKS proxy connection.
+func SocksProxyFallbackDialContext(
```
benjirewis, line 260: Let's walk through this method offline if that SGTY. I'm a little confused about the functional return, concurrency, and "delay" you've added to the SOCKS method. Not that I disagree with them; it's just tough for me to follow what's happening.

rpc/dialer.go
```diff
@@ -240,6 +241,145 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
...
+		if os.Getenv(OnlySocksProxyEnvVar) != "" {
+			logger.Infow("Both SOCKS_PROXY and ONLY_SOCKS_PROXY specified, only SOCKS proxy will be used for outgoing connection")
+			conn, err := socksProxyDialContext(ctx, network, proxyAddr, addr)
+			if err == nil {
+				logger.Infow("connected with SOCKS proxy")
+			}
+			return conn, err
+		}
+
+		type dialResult struct {
```
cheukt, line 285: reference https://go.dev/src/net/dial.go#L585

rpc/const.go
cheukt: I think that should be ok, + internet will almost always be prioritized/chosen even without this knob

benjirewis: Discussed offline; this is also helpful for development, as we can force all traffic through the SOCKS proxy even with a WiFi connection (ssh still possible.)

rpc/dialer.go
benjirewis: Discussed offline; I have a better sense of how this function now works.

rpc/dialer.go
benjirewis: Could you leave this reference in-line in the code, too? Just in case we come back to it later.

rpc/dialer.go
```diff
@@ -240,6 +241,152 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
...
+			}
+		}
+
+		logger.Infow("SOCKS_PROXY specified, SOCKS proxy will be used as a fallback for outgoing connection")
+		// start the main dial attempt.
+		primaryCtx, primaryCancel := context.WithCancel(ctx)
+		defer primaryCancel()
+		wg.Add(1)
+		primaryDial := func(ctx context.Context) (net.Conn, error) {
+			var zeroDialer net.Dialer
```
benjirewis, line 319: > `zeroDialer`

What's a "zero" dialer? Like the zero-valued `net.Dialer`? Just default settings?

rpc/dialer.go
```diff
@@ -240,6 +241,152 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
...
+					return nil, primary.error
+				}
+				if res.primary && fallbackTimer.Stop() {
+					// If we were able to stop the timer, that means it
+					// was running (hadn't yet started the fallback), but
+					// we just got an error on the primary path, so start
+					// the fallback immediately (in 0 nanoseconds).
+					fallbackTimer.Reset(0)
+				}
+			}
```
benjirewis, line 369: Should there not be a
```
case <-ctx.Done():
```
here? Will this not spawn infinite SOCKS-proxy-dialing goroutines if neither WiFi not SOCKS proxy are available?

rpc/dialer.go
```diff
@@ -240,6 +241,152 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
...
+			case <-fallbackTimer.C:
+				wg.Add(1)
+				fallbackDial := func(ctx context.Context) (net.Conn, error) {
+					return socksProxyDialContext(ctx, network, proxyAddr, addr)
+				}
+				go dialer(fallbackCtx, fallbackDial, false)
+			case res := <-results:
+				if res.error == nil {
+					if res.primary {
+						logger.Infow("connected without SOCKS proxy")
```
benjirewis, line 348:
```suggestion
						logger.Infow("connected with ethernet/wifi")
```
[nit] Just to be more clear about the two methods of connection here.

rpc/dialer.go
```diff
@@ -240,6 +241,152 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
...
+					}
+					return res.Conn, nil
+				}
+				if res.primary {
+					primary = res
+				} else {
+					fallback = res
+				}
+				if primary.done && fallback.done {
+					return nil, primary.error
```
benjirewis, line 360: [q] Can you explain what's happening in this case? No connection can be returned even though they both completed?

benjirewis (commented):
Thanks for discussion offline 🫡

rpc/dialer.go
cheukt: yep

rpc/dialer.go
cheukt: if both primary and fallback are done with errors, this means neither connection attempt succeeded. return the error from the primary dial attempt in that case.

rpc/dialer.go
cheukt: once primary and fallback dial functions are done, we will exit because of L359. If ctx gets cancelled, the two dials will end and also end up at L359

cheukt (commented):
added some comments but definitely point out other confusing bits if you find any!

rpc/dialer.go
benjirewis: Cool; assume you mean L364 in the new line numbers.

rpc/dialer.go
```diff
@@ -240,6 +241,156 @@ func DialDirectGRPC(ctx context.Context, address string, logger utils.ZapCompati
...
+		wg.Add(1)
+		primaryDial := func(ctx context.Context) (net.Conn, error) {
+			// create a zero-valued net.Dialer to use net.Dialer's default DialContext method
+			var zeroDialer net.Dialer
+			return zeroDialer.DialContext(ctx, network, addr)
+		}
+		go dialer(primaryCtx, primaryDial, true)
+
+		// wait a small amount before starting the fallback dial (to prioritize the primary connection method).
+		fallbackTimer := time.NewTimer(time.Second)
```
benjirewis, line 327: > time.Second

If I'm reading Golang source code correctly, their fallback delay is customizable and defaults to 300ms. I understand not exposing this as a knob, but any reason to bump to 1s? Did you see SOCKS being "over" prioritized with 300ms?

benjirewis (approved):
LGTM, just a question about fallback delay.

rpc/dialer.go
cheukt: nope, let me test and drop to 300ms if no issues

cheukt:
@benjirewis given the new direction of the ticket around dynamic switching of connection types, going to cut a release here first to fully support agent using the socks proxy. I also unexported one of the functions to make it easier to change the impls. tested manually with the same test cases and also with the agent

cheukt:
see https://github.com/viamrobotics/agent/pull/79 for corresponding agent changes

benjirewis (approved):
Still LGTM!

PR #417, opened 2025-03-12
RSDK-9969: Perform "smart close" when learning of disconnection via an ice state change.

dgottlieb:


rpc/wrtc_base_channel.go
```diff
@@ -67,6 +67,15 @@ func newBaseChannel(
 		// to receive messages when the peer is done.
 		ch.cancel()
```
benjirewis, line 68: I think we can remove this `ch.cancel()`.

rpc/wrtc_base_channel.go
```diff
@@ -67,6 +67,15 @@ func newBaseChannel(
...
+		// other things, poke condition variables with the appropriate locks to ensure that any
+		// goroutines blocked on sending data/servicing a request will wake up and exit. Returning
+		// their resources (e.g: call ticket).
+		//
+		// See RSDK-9969 for reproduction steps.
+		ch.Close()
+
 		if !peerDoneOnce && onPeerDone != nil {
 			peerDoneOnce = true
 			onPeerDone()
```
benjirewis, line 81: Can't comment on the line, but I think we'll need a _bit_ more logic about marking `ch.activeBackgroundWorkers.Done()` _before_ calling `doPeerDone()`. I found a change that's ugly but works; let's discuss offline.

rpc/wrtc_base_channel.go
```diff
@@ -83,8 +92,6 @@ func newBaseChannel(
 				return
 			}
 
-			ch.mu.Lock()
-			defer ch.mu.Unlock()
 			switch connectionState {
 			case webrtc.ICEConnectionStateDisconnected,
 				webrtc.ICEConnectionStateFailed,
```
benjirewis, line 97: Can't comment on the line, but I think the `peerConn.GracefulClose` on L120 is now redundant.

benjirewis (commented):
Thanks! Pointed out some now redundant "cleanup" code and a potential deadlock.

I also filed RSDK-10234 to separate `wrtc_base_*` into `wrtc_client_*` and `wrtc_server_*` so we don't have to consider both in the same code (no conditionals around `onPeerDone != nil` for example.)

rpc/wrtc_base_channel.go
```diff
@@ -61,101 +61,99 @@ func newBaseChannel(
...
-					logger.Debugw("Connection state changed", connectionStateChangedLogFields...)
-				}
+				}()
 			}
-		})
-	}
-	peerConn.OnICEConnectionStateChange(connStateChanged)
-
-	// fire once
-	connStateChanged(peerConn.ICEConnectionState())
```
dgottlieb, line 158: I claim that for the cases where this is a observable behavior change, those are the cases we're uninterested in.

rpc/wrtc_base_channel.go
```diff
@@ -61,101 +61,96 @@ func newBaseChannel(
...
+
+			// The Disconnected/Failed/Closed states are all interpreted as terminal. In theory,
+			// WebRTC allows for moving from the disconnected -> connected state if the network
+			// heals or new candidates are presented. We forgo that optimization of preserving an
+			// existing PeerConnection object and instead opt for clients to redial and perform a
+			// new round of signaling/answering.
+			connIDMu.Lock()
+			currConnID := connID
+			connIDMu.Unlock()
+			if currConnID == "" { // make sure we've gathered information before
```
benjirewis, line 83: > // make sure we've gathered information before

I know this was preexisting, but is this comment missing something? I don't understand what it means or what it's referring to.

rpc/wrtc_base_channel.go
```diff
@@ -61,101 +61,96 @@ func newBaseChannel(
...
-				if currConnID == "" { // make sure we've gathered information before
-					return
-				}
+			if connectionState == webrtc.ICEConnectionStateDisconnected {
+				// Disconnections happen when the client and server are no longer communicating. But
+				// neither side has closed the peer connection. These events can happen in more
+				// interesting states than typical shutdown (e.g: there may be a bunch of data that
+				// has been queued up to send due to the network problem) such that we feel it's
+				// warranted to log at a higher level.
+				logger.Warnw("connection state changed",
```
benjirewis, line 93: > Warnw

Why not `Infow`? I worry users will see this log pretty frequently and worry something is wrong if it's at `Warn` level.

rpc/wrtc_base_channel.go
benjirewis: > // fire once

Yeah... I don't even know what this was doing.

benjirewis (approved):
Pretty much LGTM thanks! Just some questions.

cheukt:
generally following, let me know @dgottlieb if you want an explicit approval - I'm definitely missing a bit of context around some of the later changes

cheukt (approved):
LGTM

rpc/wrtc_base_channel.go
dgottlieb: Warn felt appropriate to me -- nobody asked to be disconnected. And it's exceptional from our perspective where more things can be concurrently happening compared to failed/closed.

But I don't really search by log level -- so warning isn't particularly useful to me. Will move it down to info.

rpc/wrtc_base_channel.go
dgottlieb: I agree the comment is bad. I added in why* connID would be empty/non-empty that gives a better picture to the intent.

rpc/wrtc_base_channel.go
dgottlieb: Removed

rpc/wrtc_base_channel.go
dgottlieb: You're right this was broken because ch.Close waits on background workers. And this ice state change handler was considered a background worker.

I changed the ice state change handler to not kick off a background worker for this disconnecting case (in addition to many of the other cases).

rpc/wrtc_base_channel.go
dgottlieb: done

rpc/wrtc_server_channel_test.go
```diff
@@ -22,18 +23,21 @@ func TestWebRTCServerChannel(t *testing.T) {
...
 	}()
 
 	server := newWebRTCServer(logger)
+	defer server.Stop()
 	// use signaling server just as some random service to test against.
 	// It helps that it is in our package.
 	queue := newMemoryWebRTCCallQueueTest(logger)
 	defer queue.Close()
-	signalServer := NewWebRTCSignalingServer(queue, nil, logger, defaultHeartbeatInterval)
+	signalServer := NewWebRTCSignalingServer(queue, nil, utils.Sublogger(logger, "client"), defaultHeartbeatInterval)
```
cheukt, line 40:
```suggestion
	signalServer := NewWebRTCSignalingServer(queue, nil, utils.Sublogger(logger, "signalingServer"), defaultHeartbeatInterval)
```
?

rpc/wrtc_server_channel_test.go
```diff
@@ -252,18 +256,21 @@ func TestWebRTCServerChannelResetStream(t *testing.T) {
...
```
cheukt, line 273:
```suggestion
	signalServer := NewWebRTCSignalingServer(queue, nil, utils.Sublogger(logger, "signalingServer"), defaultHeartbeatInterval)
```
?

cheukt (approved):
LGTM, thanks for the extra comments!

PR #419, opened 2025-03-25
fix(rpc/wrtc_client_channel): Add no-op handler for StaticMethodCallOption

HipsterBrown:
The [speechio module](https://app.viam.com/module/viam-labs/speech) depends on a custom proto API called the [speech-service-api](https://github.com/viam-labs/speech-service-api), which builds Go language bindings using the protoc-gen-go-grpc. It appears that newer versions of that protoc plugin automatically include the StaticMethod CallOption in the generated code, which leads to the error output seen below from the `rpc/wrtc_client_channel.go` logic:

image

To resolve this I added a no-op case handler for the `StaticMethodCallOption`, which is only used ["as a signal to stats handler that a method is safe to use as an instrument key"](https://github.com/grpc/grpc-go/releases/tag/v1.62.0) for opentelemetry. 

I also updated the protoc-gen-go-grpc dependency to the latest version and regenerated the service code to keep things up to date. Since this is not required, I'm happy to revert those changes if asked.


dgottlieb:
@HipsterBrown , not opposed to this by any means. Just as someone that's uninformed about all this other grpc stuff, can you share a way for me to exercise this `case grpc.StaticMethodCallOption:` logic?

Just so I can get some context on how that gets called in the context of requests over the wire.

go.mod
```diff
@@ -54,7 +54,7 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20240827150818-7e3bb234dfed
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1
 	google.golang.org/grpc v1.66.0
-	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.5.1
```
benjirewis, line 57: Does this upgrade control the version of the Golang proto compiler invoked by `buf`?

rpc/wrtc_client_channel.go
```diff
@@ -155,6 +155,7 @@ func (ch *webrtcClientChannel) invoke(
 				if clientStream.trailers != nil {
 					*optV.TrailerAddr = clientStream.trailers.Copy()
 				}
+			case grpc.StaticMethodCallOption:
```
benjirewis, line 158: I think we'll want a comment above this `case` stating why we're accepting and ignoring this call option.

benjirewis (commented):
Same thoughts as @dgottlieb, a question, and a nit.

go.mod
benjirewis: Makes sense; thanks!

HipsterBrown:
@dgottlieb @benjirewis 

It appears that testing this code path requires creating a WebRTC client connection using the RDK and augmenting one of the gRPC service client methods (such as `GetMachineStatus` in the `RobotClient.MachineStatus` method) to include `grpc.StaticMethod()` in the list of `opts` it accepts. We don't expose this through any of the public RDK methods as far as I can tell. 

Example update to `robot/client/client.go` in the RDK on line 1108:
```go
func (rc *RobotClient) MachineStatus(ctx context.Context) (robot.MachineStatus, error) {
	mStatus := robot.MachineStatus{}

	req := &pb.GetMachineStatusRequest{}
	resp, err := rc.client.GetMachineStatus(ctx, req, googlegrpc.StaticMethod())
```

Then building and using the RDK in an example script:

```go
package main

import (
	"context"

	"go.viam.com/rdk/logging"
	"go.viam.com/rdk/robot/client"
	"go.viam.com/utils/rpc"
)

func main() {
	logger := logging.NewDebugLogger("client")
	machine, err := client.New(
		context.Background(),
		"<address of machine>",
		logger,
		client.WithDialOptions(rpc.WithEntityCredentials( 
            /* Replace "<API-KEY-ID>" (including brackets) with your machine's API key ID */
			"<API-KEY-ID>",
			rpc.Credentials{
				Type:    rpc.CredentialsTypeAPIKey, 
                /* Replace "<API-KEY>" (including brackets) with your machine's API key */
				Payload: "<API-KEY>",
			})),
	)
	if err != nil {
		logger.Fatal(err)
	}

	defer machine.Close(context.Background())
	logger.Info("Status:")
	logger.Info(machine.MachineStatus())
}
```

You'll need to update the `go.mod` file for the above script to use the local version of the RDK using the `replace` directive. 

I'm not sure of a simpler way to test this in a real scenario based on my understanding of the system.

HipsterBrown:
I just recreated this setup on my laptop and confirmed the error message no longer logs for the StaticMethod CallOption without issue.

benjirewis:
Thanks much for that example. I didn't even what [`grpc.CallOption`](https://pkg.go.dev/google.golang.org/grpc#CallOption)s were or that you could pass them as a variadic argument to all calls. Seems like a reasonable change to me.

rpc/wrtc_client_channel.go
benjirewis: LGTM as long as we get a comment in here.

benjirewis (approved):

dgottlieb (approved):

PR #421, opened 2025-04-01
APP-6331 upgrade imports to remove transitive dep on square/go-jose

abe-winter:
## What changed
- Update direct dependencies to get rid of a transitive dependency
## Why
dependabot (square/go-jose package should be replaced by go-jose/go-jose)
## Follow-up work
- consume this change in internal code, test that `jwks.NewCachingOIDCJWKKeyProvider` features still work


mcous (approved):
Changes LGTM, transitive dep in question _is_ removed from `go.mod`

PR #422, opened 2025-04-08
RSDK-10402 - Fix pings to machine-name.local 

cheukt:
might not be strictly necessary for what the ask is (disabling TLS), but this was already inflight, so figured I should finish this.

tested pings from another pi
tested Go SDK mDNS connections from same computer
tested pings from same computer

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -550,7 +550,8 @@ func NewServer(logger utils.ZapCompatibleLogger, opts ...ServerOption) (Server, 
 					server.mdnsServers = append(server.mdnsServers, mdnsServer)
 
 					// register a second address to match queries for machine-name.local
-					// RSDK-10409 - This may not be necessary
+					// RSDK-10409 - Depending on if we need the previous block to register mDNS addresses with
+					// the system hostname, we may be able to combine the two separate registrations into one.
 					if host != hostname {
 						mdnsServer, err = zeroconf.RegisterProxy(
 							host,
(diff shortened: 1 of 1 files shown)


rpc/server.go
```diff
@@ -545,6 +603,33 @@ func NewServer(logger utils.ZapCompatibleLogger, opts ...ServerOption) (Server,
 						break
 					}
 					server.mdnsServers = append(server.mdnsServers, mdnsServer)
+
+					// register a second address to match queries for machine-name.local
+					// RSDK-10409 - This may not be necessary
```
dgottlieb, line 608: Maybe this is what you mean by not necessary and (the ticket's description) of too many mdns servers?

But I'd expect this `RegisterProxy` call to happen in the line 574 else-block, but outside of the line 575 "for host in instance names" loop.

cheukt: we do need these changes for pings to machine-name.local to work, and I do believe we need to register an extra mdns server for each instance name (so machine.location.viam + machine.local.location.viam + their dash variants)

the "necessary" part relates to the improvement we talked about around whether we need to register the system hostname as the hostname for the mDNS address (the current state), so if we don't need to do that, we can consider only combining/removing the one of the registrations for each host

cheukt: updated the comments to clarify

dgottlieb (approved):
Cool, I think I just projected out from the reproducer in the wrong way. We want every "robot name.local" to be resolveable by ping.

And that hostname check is just taking care of a the case where robot name == system hostname.

PR #427, opened 2025-04-17
APP-7773: ignore pollWait goleak caused by (probably) GCS client

mcous:
## Overview

We're still getting flaky `goleak` failures in app, even after skipping some tests that we know were leaking goroutines.

Found this in other OSS projects. I have no sense for how much of a nuclear option this is - as in, I don't know how often legitimate problems in our code would have this at the top of the leak stack - but putting up this PR in case it's a reasonable to our allowlist

## Review requests

Is this at all acceptable?


mcous:
> Ignoring poll wait is probably not acceptable no

👌 will dig up a recent failure and send it your way

PR #428, opened 2025-04-21
adjust deadline forward when system clock is off

abe-winter:
## What changed
- factor out deadline logic to `getDeadline`, add test
- more detailed logging
## Why
Necessary to establish a connection on a machine with an unsynchronized clock
## Notes
We talked about using default offer deadline -- realized afterwards that because of network latency, we'd seldom be within this threshold. Sticking with 5 seconds for now. I think this logic needs redesign, but for now publishing as-is to patch.

### DIFF
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -151,6 +151,20 @@ func isNetworkError(err error) bool {
 	return true
 }
 
+// logic for computing actual deadline from server-provided deadline
+func getDeadline(ctx context.Context, logger utils.ZapCompatibleLogger, initStage *webrtcpb.AnswerRequest_Init) (context.Context, func()) {
+	if deadline := initStage.Init.GetDeadline(); deadline != nil {
+		asTime := deadline.AsTime()
+		earliestDeadline := time.Now().Add(5 * time.Second)
+		if asTime.Before(earliestDeadline) {
+			logger.Warnw("updating deadline to be ahead of system clock", "received", asTime.String(), "adjusted", earliestDeadline.String())
+			asTime = earliestDeadline
+		}
+		return context.WithDeadline(ctx, asTime)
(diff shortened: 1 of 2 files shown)


rpc/wrtc_signaling_answerer.go
```diff
@@ -151,6 +151,20 @@ func isNetworkError(err error) bool {
 	return true
 }
 
+// logic for computing actual deadline from server-provided deadline.
+func getDeadline(ctx context.Context, logger utils.ZapCompatibleLogger, initStage *webrtcpb.AnswerRequest_Init) (context.Context, func()) {
+	if deadline := initStage.Init.GetDeadline(); deadline != nil {
```
dgottlieb, line 156: I appreciate trying to respect the init stage deadline (or more candidly known as -- some random ass number that has no bearing on reality). I'm inclined to just do a `context.WithTimeout(ctx, 10*time.Second)` here and call it a day (SDKs[1] use 10 seconds).

Feel free to merge this as-is and I can loop back. Appreciate all you've done and don't want this ask to be a burden!

[1] Python until recently was only 5 seconds.

dgottlieb (approved):

rpc/wrtc_signaling_answerer.go
cheukt: I have a slight concern that 5s is not enough https://github.com/viamrobotics/goutils/blob/a21ac14344db965df01d577ad52f690a5213538d/rpc/wrtc_client.go#L157 for the go sdk will always wait at least 5s. I would agree that maybe the easier thing is to log a warning and just use 10s all the time

abe-winter: yes, going to merge as is so we can get this into stable RDK

I think you're right from a broader design perspective, but want to give us breathing room to think those thoughts

PR #432, opened 2025-05-07
RSDK-10618: Fix race contition in managedProcess

jmatth:
This PR fixes an issue where a race could occur between calls to `managedProcess.Stop` and parts of `managedProcess.manage` if the process crashed just before the call to `Stop`. This will have a matching change on the rdk side where the race caused duplicate module processes to be created during reconfiguration.

One thing worth noting: this change has the potential to create a deadlock if calling code tries to hold the same lock goroutines calling `Stop` and in their restart callback. This was an issue I fixed in my rdk modmanager changes but there may be other places this code is used that need to be audited for potential deadlocks.

### DIFF
--- a/pexec/managed_process.go
+++ b/pexec/managed_process.go
@@ -101,7 +101,6 @@ type managedProcess struct {
 	shouldLog bool
 	cmd       *exec.Cmd
 
-	stopped          bool
 	onUnexpectedExit func(int) bool
 	managingCh       chan struct{}
 	killCh           chan struct{}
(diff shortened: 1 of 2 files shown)


jmatth:
@dgottlieb I'm still testing things on the rdk side but I think this PR can be looked at now. The most important update from when we last went over these changes is `manage` now drops the lock before calling the unexpected exit handler like you recommended. That does mean that it's the responsibility of whatever code is using pexec to prevent races between the callback and any other code that calls `Stop` etc, which is what I'm working on over in the rdk modmanager.

dgottlieb (approved):

dgottlieb:
This all looks great to me. Very nice job

PR #436, opened 2025-06-03
RSDK-10183: Track PeerConnection disconnects in addition to active clients in FTDC.

dgottlieb:
image

### DIFF
--- a/rpc/wrtc_base_channel.go
+++ b/rpc/wrtc_base_channel.go
@@ -63,6 +63,9 @@ func newBaseChannel(
 
 	var connID string
 	var connIDMu sync.Mutex
+
+	// In the normal connection flow, the client closes the WebRTC PeerConnection. Which closes the
+	// client DataChannel which results in the server DataChannel having its `OnClose` being called.
 	peerConn.OnICEConnectionStateChange(func(connectionState webrtc.ICEConnectionState) {
 		switch connectionState {
 		case webrtc.ICEConnectionStateDisconnected,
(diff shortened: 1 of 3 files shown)


rpc/wrtc_base_channel.go
```diff
@@ -144,9 +147,20 @@ func newBaseChannel(
 func (ch *webrtcBaseChannel) Close() {
 	// RSDK-8941: Having this instead early return when `closed` is set will result in `TestServer`
 	// to leak goroutines created by `dialWebRTC`.
```
benjirewis, line 149: Is this comment relevant anymore? Seems like we early return now, but we do wait for `activeBackgroundWorkers`. Is that enough to stop the leaked goroutines?

rpc/wrtc_signaling_answerer.go
```diff
@@ -400,6 +400,9 @@ func (aa *answerAttempt) connect(ctx context.Context) (err error) {
 		}
 	}()
 
+	// `NewChannel` instantiates our webrtc wrapper around DataChannels. This includes callback
+	// handlers for transitioning to the open/closed/error states. As well as backpressure when the
+	// amount of data to send gets high.
```
benjirewis, line 405: > backpressure when the amount of data to send gets high

Is this referring to the buffer threshold? What is "backpressure"?

dgottlieb: Yes, the buffer threshold is backpressure.

Backpressure is when a function, after determining it's using too many resources, either:
* fails back to the caller saying "i'm too busy" or
* blocks until there are more available resources before processing the function

call tickets is another example of backpressure.

PR #446, opened 2025-06-24
Turn refactor

jmatth:
This changed from "reduce the number of variables" to "just refactor and add tests" as I worked on it. My initial plan was to reduce it to 2 variabes: one to filter the turn server and another to switch to turns, but then that second variable would be responsible for switching the scheme to turns, the protocol to tcp, and the port to 443. Instead this renames a bunch of the variables to be more intuitive and replaces "TURN_HOST" with "TURN_URI" to allow for more specific filtering with just that variable.

I can also switch this PR to remove all but the "TURN_URI" variable, and leave out the scheme, port, and transport manipulation. That' probably better in the long term as long as we don't think we'll need it soon and we're willing to modify app to provide a wider array of turn servers. Making that change should just involve reverting the second commit so it's no extra work on this PR either way.

### DIFF
--- a/rpc/const.go
+++ b/rpc/const.go
@@ -19,20 +19,24 @@ var (
 	OnlySocksProxyEnvVar = "ONLY_SOCKS_PROXY"
 
 	// TURNURIEnvVar is the name of an environment variable used to select at
-	// most one TURN server.
+	// most one TURN server. This parameter is experimental and may be changed or
+	// removed in future versions.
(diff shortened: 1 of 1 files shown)


dgottlieb:
I'll take myself off the review list. There's a test which acts as a description of how to use the feature, so my main concern is covered.

cheukt (approved):
LGTM, I think keeping the extra env vars is fine as long as we can keep it sane and most invocations only end up needing one or two of them

PR #450, opened 2025-07-02
RSDK-11186 Elaborate on stream and call ticket limit logs

benjirewis:


cheukt (approved):
I do believe that we'll be imminently getting rid of these limits, but if it'll help the investigation I'm all for it

benjirewis:
Closing since, as @cheukt pointed out, imminent work for [gRPC limit per resource](https://viam.atlassian.net/browse/RSDK-10794) will get rid of these code paths. The changes I made are not particularly useful for the investigation; they were to be informative to users.

PR #453, opened 2025-07-10
RSDK-11255, RSDK-11258: use syscall.Kill instead of shelling out to system's kill; fix TestManagedProcessKillGroup

aldenh-viam:
Details in linked tickets. This kills the process group successfully on my Linux machine, and more reliably tests that they've been killed.


cheukt (approved):
thanks!

PR #454, opened 2025-07-11
RSDK-11230: Change semantics of when `managedProcess.Stop` returns an error.

dgottlieb:


pexec/managed_process.go
```diff
@@ -472,36 +472,17 @@ func (p *managedProcess) Stop() error {
...
-			}
-		}
-
-		unknownStatus = unknownStatus || isWaitErrUnknown(p.lastWaitErr.Error(), forceKilled)
-		if unknownStatus {
-			p.logger.Debug("unable to check exit status")
-			return nil
-		}
-		return p.lastWaitErr
+		p.logger.Warnw("Stopped module did not exit cleanly", "err", p.lastWaitErr)
```
dgottlieb, line 482: This bit of removing all of the error handling to simply log a warn is the real change. The changes to unix/windows are just removing an unused function. And otherwise changing the expectations of tests.

PR #461, opened 2025-08-06
RSDK-11434 Managed Process Status on Windows

bkaravan:
Implemented a `Status` method on windows, which is a best effort approach to translate the sate of the process on windows to the `os.ErrProcessDone` error that we are looking for. 

This have been discussed offline, but implementing this `Status` exposed an interesting code epath that lead to deadlock on Windows. In particular, these steps led to a deadlock: 
- When a module manager attempts to [add](https://github.com/viamrobotics/rdk/blob/1b7f3c914c1508dd6571a70bfae1c5bc345c64d3/module/modmanager/manager.go#L228) a module, it will acquire a lock. 
- When a module fails to start, [cleanupAfterStartupFailure()](https://github.com/viamrobotics/rdk/blob/1b7f3c914c1508dd6571a70bfae1c5bc345c64d3/module/modmanager/module.go#L383) is called, which will call `stopProcess()`
- Eventually, `Stop()` is called, and our line of interest is this call to [kill](https://github.com/viamrobotics/goutils/blob/bde94ecac9bc73d6c096e3f359a5089deb6a8559/pexec/managed_process.go#L475)
- The return value of this kill was different based on the architecture. On Unix, an attempt to send SIGTERM to the whole process group would result in an ESRCH, which is a "no such process" error. This error would be returned from kill, and then propagated to mod manager. The mod manager would special case this error message, and finish its execution normally
- On windows, "no such process" error is caught slightly differently, and instead returns without an error. This would cause `Stop()` to wait on the managing channel, which is usually closed by the `manage` go routine. However, this routine has a [callback](https://github.com/viamrobotics/goutils/blob/bde94ecac9bc73d6c096e3f359a5089deb6a8559/pexec/managed_process.go#L313) back into the module manager, which will _also_ try to acquire the same lock that `Add` is still holding. Therefore, the program enters a deadlock, with the modmanager lock being attempted by two places, as well as `Stop` waiting on the managing channel

In a best effort approach to unify Unix and Windows code paths for now, @jmatth offered to created a custom error that will be returned from both `kill` variants. So, to finish this PR, another change in the RDK needs to be made, which will special case check our new custom `ErrProcessNotExist` rather than a string match. While this patch fixes Status on windows and makes a way around this deadlock, a bigger rewrite of managed process (with it migrating to rdk) might be required to better handle these synchronization issues.

### DIFF
--- a/pexec/managed_process_unix.go
+++ b/pexec/managed_process_unix.go
@@ -107,7 +107,13 @@ func (p *managedProcess) kill() (bool, error) {
 	select {
 	case <-timer.C:
 		p.logger.Infof("stopping entire process group %d with signal %s", p.cmd.Process.Pid, p.stopSig)
-		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig); err != nil && !errors.Is(err, os.ErrProcessDone) {
+		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig); err != nil {
+			errno, ok := err.(syscall.Errno)
+			if ok && errno == syscall.ESRCH {
+				return false, &ErrProcessNotExist{err}
+			} else if errors.Is(err, os.ErrProcessDone) {
+				return false, &ErrProcessNotExist{err}
+			}
 			return false, errors.Wrapf(err, "error signaling process group %d with signal %s", p.cmd.Process.Pid, p.stopSig)
 		}
 	case <-p.managingCh:
@@ -121,7 +127,13 @@ func (p *managedProcess) kill() (bool, error) {
 	select {
 	case <-timer2.C:
 		p.logger.Infof("killing entire process group %d", p.cmd.Process.Pid)
-		if err := syscall.Kill(-p.cmd.Process.Pid, syscall.SIGKILL); err != nil && !errors.Is(err, os.ErrProcessDone) {
+		if err := syscall.Kill(-p.cmd.Process.Pid, syscall.SIGKILL); err != nil {
+			errno, ok := err.(syscall.Errno)
+			if ok && errno == syscall.ESRCH {
+				return false, &ErrProcessNotExist{err}
+			} else if errors.Is(err, os.ErrProcessDone) {
+				return false, &ErrProcessNotExist{err}
+			}
 			return false, errors.Wrapf(err, "error killing process group %d", p.cmd.Process.Pid)
 		}
 		forceKilled = true
--- a/pexec/managed_process_unix.go
+++ b/pexec/managed_process_unix.go
@@ -87,7 +87,7 @@ func (p *managedProcess) sysProcAttr() (*syscall.SysProcAttr, error) {
 	return attrs, nil
 }
 
-func (p *managedProcess) Status() error {
+func (p *managedProcess) status() error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	return p.cmd.Process.Signal(syscall.Signal(0))
--- a/pexec/managed_process_unix.go
+++ b/pexec/managed_process_unix.go
@@ -87,6 +87,12 @@ func (p *managedProcess) sysProcAttr() (*syscall.SysProcAttr, error) {
 	return attrs, nil
 }
 
+func (p *managedProcess) Status() error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	return p.cmd.Process.Signal(syscall.Signal(0))
+}
+
 func (p *managedProcess) kill() (bool, error) {
 	p.logger.Infof("stopping process %d with signal %s", p.cmd.Process.Pid, p.stopSig)
 	// First let's try to directly signal the process.
--- a/pexec/managed_process_windows.go
+++ b/pexec/managed_process_windows.go
@@ -117,7 +117,7 @@ func (p *managedProcess) forceKillGroup() error {
 
 // Status is a best effort method to return an os.ErrProcessDone in case the process no
 // longer exists.
-func (p *managedProcess) Status() error {
+func (p *managedProcess) status() error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	pid, err := p.UnixPid()
--- a/pexec/managed_process_windows.go
+++ b/pexec/managed_process_windows.go
@@ -50,7 +50,7 @@ func (p *managedProcess) kill() (bool, error) {
 			p.logger.Debug("must force terminate process")
 			shouldJustForce = true
 		case strings.Contains(string(out), "not found"):
-			return false, nil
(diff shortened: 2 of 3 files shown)


cheukt:
can we add windows CI tests like in https://github.com/viamrobotics/rdk/blob/main/.github/workflows/test.yml#L93? even if only for the pexec package or a subset. if none of the current tests work, we should add at least one new test to run that tests the behavior we're fixing here

pexec/managed_process_unix.go
```diff
@@ -101,7 +107,13 @@ func (p *managedProcess) kill() (bool, error) {
 	select {
 	case <-timer.C:
 		p.logger.Infof("stopping entire process group %d with signal %s", p.cmd.Process.Pid, p.stopSig)
-		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig); err != nil && !errors.Is(err, os.ErrProcessDone) {
```
benjirewis, line 104: [q] To clarify my understanding, we're no longer "swallowing" the `os.ErrProcessDone` error to move on to killing the entire process group? And, we're now returning a custom `ErrProcessNotExist` error to be caught in [`modmanager/manager.go`](https://github.com/viamrobotics/rdk/blob/1b7f3c914c1508dd6571a70bfae1c5bc345c64d3/module/modmanager/module.go#L312-L312) instead of looking for "no such process," which is only a Unix error?

benjirewis (commented):
Nice! Thanks for looking into this more with @jmatth .

pexec/managed_process_unix.go
bkaravan: "no such process" is the error message for the errorno ESRCH, which I _am pretty sure_ is unix-only. The previous behavior was sending SIGTERM to the whole process group, and if we get an `os.ErrProcessDone`, we would not return an error and keep going. However, this case was returning ESRCH instead, which we were special casing in `modmanager/manager.go`. Now, we are returning our new custom error in case we both fail to find the process, or the process is done. We propagate this error on both windows and unix, and in a future PR, will special check this new error rather than a string match.

pexec/managed_process_windows.go
```diff
@@ -46,64 +51,72 @@ func (p *managedProcess) kill() (bool, error) {
 		switch {
 		case strings.Contains(string(out), mustForce):
 			p.logger.Debug("must force terminate process")
+			// if taskkill doesn't find a window to terminate the process, we will attempt to
+			// send a "break" control event, which asks for a graceful shutdown of the whole
+			// process group.
+			if err := windows.GenerateConsoleCtrlEvent(windows.CTRL_BREAK_EVENT, uint32(p.cmd.Process.Pid)); err != nil {
```
bkaravan, line 57: Dan tasked me with some research about the guarantees that we have from calling `kill` on windows, which made me discover that `taskkill` is a very limited binary that would pretty much almost always lead to force killing modules on windows. 
@jmatth helped me look into this more, and here are our findings: 
- `taskkill` without the `/f` (force) flag sends a WM_QUIT signal to the process. This is roughly equivalent to the user hitting the X button in the corner of the application window. Success of this signal relies on the process to be a part of the "window" lifecycle, in other words, having some sort of open window/console for its purposes
- since the vast majority of modules run as a background process, WM_QUIT would not propagate to them, so this graceful shutdown will never happen. An alternative call that we found is `GenerateConsoleCtrlEvent`, which sends a specified signal (in our case, `CTRL_BREAK_EVENT` to the process group that shares the same console as the specified PID. Since viam-server specifies a `CREATE_NEW_PROCESS_GROUP` creation flag when spinning up a module, this approach correctly delivers a stopping signal (according to our testing, it translates to SIGINT) to the modules process group, which leads to its graceful stop. 
- However, since there could potentially be modules that create their own processes, and since those processes could technically have their own consoles/windows depending on creation flags, we thought that the most robust approach would be removing the if `!shouldJustForce` guard when deciding whether to kill the process tree, and always do it, unless the management channel is closed already. 
- The final force kill case is the same, just slightly reformatted for easier reading.

pexec/managed_process_unix.go
```diff
@@ -101,8 +110,15 @@ func (p *managedProcess) kill() (bool, error) {
...
-			return false, errors.Wrapf(err, "error signaling process group %d with signal %s", p.cmd.Process.Pid, p.stopSig)
+		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig); err != nil {
+			var errno syscall.Errno
+			if errors.As(err, &errno) && errno == syscall.ESRCH {
+				return false, &ProcessNotExistsError{err}
+			} else if errors.Is(err, os.ErrProcessDone) {
+				return false, &ProcessNotExistsError{err}
+			}
+			return false, errors.Wrapf(err, "error signaling process group %d with signal %s",
```
cheukt, line 120: is there a behavioral difference compared to before?

pexec/managed_process_windows.go
```diff
@@ -35,75 +37,117 @@ func (p *managedProcess) sysProcAttr() (*syscall.SysProcAttr, error) {
...
+			// shutting down viam-server as well, since they would be in the same process group.
+			if err := windows.GenerateConsoleCtrlEvent(windows.CTRL_BREAK_EVENT, uint32(p.cmd.Process.Pid)); err != nil {
+				p.logger.Debugw("sending a control break event to the process group failed with error", "err", err)
+			}
 			shouldJustForce = true
+			// Note: the following behavior is different on unix. If the first attempt to kill the
+			// process results in os.ErrProcessDone on unix, no error is returned. Only if a process
+			// is not found when trying to kill its tree or force kill it, then
+			// &ProcessNotExistsError is returned.
+			// On windows, we will always return this error, even when the first attempt failed.
```
cheukt, line 86: this difference is pretty significant and should be promoted to the top of the function

cheukt (commented):
looks good! some minor qs/doc suggestions. happy to add the windows tests to ci later if the current tests run on windows locally

cheukt (approved):
LGTM

pexec/managed_process_windows.go
```diff
@@ -35,75 +37,117 @@ func (p *managedProcess) sysProcAttr() (*syscall.SysProcAttr, error) {
...
+			return false, nil
+		default:
+			if err := forceKillCommand.Run(); err != nil {
 				return false, errors.Wrapf(err, "error force killing process tree %d", p.cmd.Process.Pid)
 			}
-			forceKilled = true
-		case <-p.managingCh:
-			timer2.Stop()
+			return true, nil
 		}
```
jmatth, line 133: We were already waiting on the management channel in the previous section so I wouldn't even bother checking it here. You can just skip straight to the code currently in the `default` case.

PR #465, opened 2025-08-14
RSDK-11496 Add metrics for connection establishment failure rate

benjirewis:
RSDK-11496

Adds metrics for total number of connection establishments, total number of connection establishment failures, and separate metrics for tracking whether those failures are from the caller or answerer and whether they are or are not timeouts. To be later used by an alerting policy that will create a GCP alert in the event that the rate of failure over a certain window exceeds a certain threshold.

I tested locally with my own app and MongoDB instances and used logs to see that `Inc()` was being called as I'd expect for both success and failure of connection establishment through the TS SDK.

### DIFF
--- a/rpc/wrtc_call_queue_mongodb.go
+++ b/rpc/wrtc_call_queue_mongodb.go
@@ -63,8 +63,8 @@ var (
 		},
 	})
 
-	connectionEstablishments = statz.NewCounter0(
-		"rpc.webrtc/connection_establishments",
+	connectionEstablishmentAttempts = statz.NewCounter0(
+		"rpc.webrtc/connection_establishment_attempts",
 		statz.MetricConfig{
 			Description: "The total number of connection establishment attempts (offer initializations).",
 			Unit:        units.Dimensionless,
@@ -925,7 +925,7 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
 
 	// An offer initialization (after verifying the host queue size), indicates an attempted
 	// connection establishment attempt.
-	connectionEstablishments.Inc()
+	connectionEstablishmentAttempts.Inc()
 
 	newUUID := uuid.NewString()
 	call := mongodbWebRTCCall{


rpc/wrtc_call_queue_mongodb.go
```diff
@@ -328,7 +344,6 @@ type mongodbCallExchange struct {
 
 type mongodbCallEvent struct {
 	Call    mongodbWebRTCCall
-	Error   error
```
benjirewis, line 331: This field was never set; I removed it.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -876,6 +891,10 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
 		return "", nil, nil, nil, err
 	}
 
+	// An offer initialization (after verifying the host queue size), indicates an attempted
+	// connection establishment attempt.
+	connectionEstablishments.Inc()
```
benjirewis, line 896: Perhaps obviously, this metric will not capture connection establishment attempts _not_ made through app's signaling server (e.g. via direct gRPC or internal signaling). It will also not capture connection establishment attempts that did not reach app's signaling server due to network issues.

Finally, it will not capture any connection establishment attempt that returns early on L891 above. If I understand the logic correctly there, we return early when there are 50 active connection establishments to a single machine. I _could_ increment even in that case, and increment failures when we return early on L891 if that's desired.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1019,17 +1033,29 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferDone(ctx context.Context, host, uu
...
-		return newInactiveOfferErr(uuid)
+	var updatedMDBWebRTCCall mongodbWebRTCCall
+	if err := updateResult.Decode(&updatedMDBWebRTCCall); err != nil {
+		return errors.Wrap(err, "could not decode document where 'caller_error' was set")
 	}
+
+	// Increment connection establishment failure count if we are setting a
+	// `caller_error` and the `answerer_error` has not already been set.
+	if updatedMDBWebRTCCall.AnswererError == "" {
+		connectionEstablishmentFailures.Inc()
```
benjirewis, line 1056: After playing around with app and MongoDB locally, I found that if _either_ of `caller_error` or `answerer_error` is set, a connection establishment has failed. I wrote some logic to increment `connectionEstablishmentFailures` when either is set and the other is _not already_ set. We don't want to double-count failures in the event of both the caller and the answerer setting an error on the document, which I _think_ is possible (although I have not actually seen it in my own testing).

Incrementing in this way will importantly not include any connection establishment errors that occur without the signaling server's knowledge. In other words, as long as neither the caller nor the answerer nor some internal issue within the signaling server sets `caller_error` or `answerer_error`, we treat the establishment as "successful." Of course, signaling could have been successful, but creating the connection with the communicated ICE candidates could have failed. That failure type is tougher to observe/measure, as only `viam-server` and SDK instances really know it has happened (not app.viam.com).

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1208,36 +1230,6 @@ func (queue *mongoDBWebRTCCallQueue) RecvOffer(ctx context.Context, hosts []stri
 		callerDoneCtx:    callerDoneCtx,
 		deadline:         offerDeadline,
 	}
-	setErr := func(errToSet error) {
```
benjirewis, line 1211: I removed this callback and its three usages. I found that `answerer_error` was always getting set on documents in the `calls` DB right before or right after their removal due to the TTL of 10s on the collection. I believe this was because the `recvCtx` in this method was erroring with `context deadline exceeded`. `answerer_error` was getting set even when the connection establishment was totally successful; that was annoying because now I'm using `caller_error` and `answerer_error` to determine if the connection establishment was successful from the signaling server's perspective. I think if `recvCtx` errors in the select and if checks I removed `setErr()` from, we just want to return from the goroutine spawned on L1251. To me, this is reminiscent of some of the bugs we had in the signaling answerer code that reported a context canceled to the signaling server even though nothing actually went wrong.

rpc/wrtc_call_queue_mongodb.go
cheukt: would caller_error and answerer_error be useful to track separately?

PR #473, opened 2025-09-02
RSDK-11644 - Add windows testing for pexec

cheukt:
successful test run here https://github.com/viamrobotics/goutils/actions/runs/17410675825

I think bash may be fine for running tests, but may have to reevaluate if we ever get user reports of something not working. 

for the failures when using cmd/powershell for shell (example https://github.com/viamrobotics/goutils/actions/runs/17409796779/job/49423798245), it seems like the test runner is unable to fully exit the command it started and so hangs for the duration of the test.

the [test](https://github.com/viamrobotics/goutils/blob/a44db38d9b411f0945bea7f5a191b95400b79bdd/pexec/managed_process_test.go#L86) in question starts another process in bash, so that could be part of what's causing issues


go.mod
```diff
@@ -51,7 +51,7 @@ require (
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1
 	google.golang.org/grpc v1.66.0
 	google.golang.org/protobuf v1.35.1
-	gotest.tools/gotestsum v1.10.0
+	gotest.tools/gotestsum v1.10.0 //should be kept in sync with mise.toml and test.yml
```
jmatth, line 54: Can we not just install this through mise on Windows the same as we do on Linux?

mise.toml
```diff
@@ -15,7 +15,7 @@ lockfile = false
 disable_backends = ["asdf"]
 
 [tools]
-buf = "1.55.1"
+"ubi:bufbuild/buf" = "1.55.1"
```
cheukt, line 18: regular buf, installed from the aqua backend, does not install on windows (it fails with `Failed to install aqua:bufbuild/buf@1.55.1: failed copy: C:\Users\Viam\AppData\Local\mise\downloads\buf\1.55.1\buf-Windows-x86_64.exe -> C:\Users\Viam\AppData\Local\mise\installs\buf\1.55.1\buf/bin/buf.exe`). leaving for posterity

jmatth, line 18: Instead of explicitly setting the backend inline it might be nicer to define an alias:

```toml
[alias]
buf = "ubi:bufbuild/buf"
```

This worked for me testing on a local Windows machine and removes the need to modify the workflow files to call `mise tool ubi:bufbuild/buf` instead of `mise tool buf`

cheukt:
successful run https://github.com/viamrobotics/goutils/actions/runs/17768872801/job/50499296123

PR #474, opened 2025-09-05
Extend ClientConn inteface to include GetState and handle webrtc case

danielbotros:


rpc/wrtc_client_channel.go
```diff
@@ -333,6 +334,11 @@ func (ch *webrtcClientChannel) writeReset(stream *webrtcpb.Stream) error {
 	})
 }
 
+// GetState returns the current connectivity state of the channel.
+func (ch *webrtcClientChannel) GetState() connectivity.State {
+	return Unknown
```
danielbotros, line 339: Maybe add a TODO here?

PR #476, opened 2025-09-09
RSDK-11748 Add `sdk_type` and `organization_id` labels to connection establishment attempt metric

benjirewis:
RSDK-11748

Adds two new labels to `signaling/connection_establishment_attempts` so we can aggregate in GCP based off of SDK type and organization ID.

This is _hacky_, although it does work on my local instance of app with the Python, C++, Golang, and TypeScript SDKs (Chrome/Safari) within the limits described by the in-line comments.

I've filed a few tickets for follow-up work to make these labels more robust and add them to more metrics, but this small change will be a nice start.

Follow-up work:
- https://viam.atlassian.net/browse/RSDK-11864
- https://viam.atlassian.net/browse/RSDK-11875
- https://viam.atlassian.net/browse/RSDK-11876

### DIFF
--- a/rpc/wrtc_call_queue_mongodb.go
+++ b/rpc/wrtc_call_queue_mongodb.go
@@ -960,7 +960,7 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


cheukt (approved):
LGTM

PR #478, opened 2025-09-16
Rsdk 11751 11652 expected connection failures to offline robots

danielbotros:


rpc/wrtc_call_queue_test.go
```diff
@@ -73,6 +80,9 @@ func testWebRTCCallQueue(t *testing.T, setupQueues func(t *testing.T) (WebRTCCal
 			close(done)
 		}()
 
+		// Give some time for the queue to recognize an answerer is online so we don't error due to thinking the host is offline.
+		time.Sleep(time.Second)
+
```
danielbotros, line 85: This is an unfortunate consequence of relying on the db to update first in order to determine if a host is offline

rpc/wrtc_call_queue_test.go
```diff
@@ -21,9 +21,16 @@ func testWebRTCCallQueue(t *testing.T, setupQueues func(t *testing.T) (WebRTCCal
...
-		_, _, ansCtx, _, err := callerQueue.SendOfferInit(context.Background(), host, "somesdp", false)
-		test.That(t, err, test.ShouldBeNil)
-		<-ansCtx
+		_, _, _, _, err := callerQueue.SendOfferInit(context.Background(), host, "somesdp", false)
+
+		// NOTE(danielbotros): This is a little bit ugly but in memory queues are only used for internal signaling to localhosts,
+		// so there isn't a concept of attempting to connect to an offline host.
+		if isInMemoryQueue(callerQueue) {
+			test.That(t, err, test.ShouldBeNil)
+		} else {
```
danielbotros, line 30: Don't love this so I'm open to suggestions to handle it otherwise

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -935,6 +1002,18 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
...
 
+	if err := queue.checkHostOnline(ctx, host); err != nil {
+		connectionEstablishmentExpectedFailures.Inc()
+		// TODO(RSDK-11928): Implement proper time-based rate limiting to prevent clients from spamming connection attempts to offline machines so
+		// we can remove sleep and error instantly.
+
+		// Machine is offline but if we return the error instantly, clients can immediately reattempt connection establishment, overwhelming the
+		// signaling server if spammed. Instead, sleep for the default offer deadline duration to slow down reattempts and give robots the chance
+		// to potentially come online.
+		time.Sleep(getDefaultOfferDeadline())
```
danielbotros, line 1013: This sleep might be too aggressive for how early we do it. With this new behavior if a machine is about to come online / JUST came online, we'll probably reject the offer and we'd have to attempt again to make a connection, whereas before we'd return from `SendOffInit()` and hang around a bit and the answers would come online before the context deadline is exceeded.

danielbotros: Discussed on slack

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -922,6 +952,43 @@ func (queue *mongoDBWebRTCCallQueue) checkHostQueueSize(ctx context.Context, for
 	return errTooManyConns
 }
 
+var errOffline = status.Error(codes.Unavailable, "robot is offline or you don't have permissions to access it")
```
benjirewis, line 955: > or you don't have permissions to access it

[q] How is it that the aggregation pipeline you've added for `checkHostOnline` returning no documents indicates that the caller may _not have access_ to machine?

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -935,6 +1002,18 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
...
 
+	if err := queue.checkHostOnline(ctx, host); err != nil {
+		connectionEstablishmentExpectedFailures.Inc()
+		// TODO(RSDK-11928): Implement proper time-based rate limiting to prevent clients from spamming connection attempts to offline machines so
+		// we can remove sleep and error instantly.
+
+		// Machine is offline but if we return the error instantly, clients can immediately reattempt connection establishment, overwhelming the
+		// signaling server if spammed. Instead, sleep for a few seconds to slow down reattempts and give robots the chance to potentially come
+		// online.
+		time.Sleep(3 * time.Second)
```
benjirewis, line 1013: [nit] Make this a constant?

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1441,6 +1520,52 @@ func (queue *mongoDBWebRTCCallQueue) Close() error {
 	return nil
 }
 
+// WaitForAnswererOnline returns a channel that will be closed once there is at least one
```
benjirewis, line 1523: > returns a channel

[q] Is there a reason to not just block in this function until the answerer is online? Rather than spawn a goroutine and return a channel?

rpc/wrtc_call_queue_mongodb_test.go
```diff
@@ -198,3 +207,29 @@ func TestMongoDBWebRTCCallQueueMulti(t *testing.T) {
...
+				"caller_size":   int64(0),
+				"answerer_size": int64(1),
+			}},
+		})
+	test.That(t, err, test.ShouldBeNil)
+}
+
+func removeFakeAnswererForHost(t *testing.T, client *mongo.Client, host string) {
+	t.Helper()
+	_, err := client.Database(mongodbWebRTCCallQueueDBName).Collection(mongodbWebRTCCallQueueOperatorsCollName).UpdateOne(context.Background(),
```
benjirewis, line 228: > UpdateOne

[q] What you have here seems fine, but is it necessary that this be an `UpdateOne` to decrement the answerer size rather than a `RemoveOne` that just removes the `operatorID` document?

rpc/wrtc_call_queue_test.go
```diff
@@ -165,10 +175,20 @@ func testWebRTCCallQueue(t *testing.T, setupQueues func(t *testing.T) (WebRTCCal
...
 		defer teardown()
 
 		host := primitive.NewObjectID().Hex()
+
+		// We need to have an answerer online to handle this host before sending the offer so it doesn't immediately fail, but we don't care what
+		// happens to the offer.
+		answerCtx, answerCancel := context.WithCancel(context.Background())
+		defer answerCancel()
+		go func() {
+			_, _ = answererQueue.RecvOffer(answerCtx, []string{host})
```
benjirewis, line 188: [nit] You _could_ assert that the received offer is in fact "hello" (or whatever caller is sending on L192). Same thing goes for similar line a couple tests below.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1441,6 +1520,52 @@ func (queue *mongoDBWebRTCCallQueue) Close() error {
...
+// answerer online for all the given hosts. Used in testing to synchronize callers and answerers so that
+// call attempts don't immediately fail due to answerers not yet registered as being online.
+func (queue *mongoDBWebRTCCallQueue) WaitForAnswererOnline(ctx context.Context, hosts []string) <-chan struct{} {
```
benjirewis, line 1526: [nit] Can we keep this method private?

rpc/wrtc_signaling_test.go
```diff
@@ -145,6 +146,7 @@ func testWebRTCSignaling(t *testing.T, signalingCallQueue WebRTCCallQueue, logge
 							DisableTrickleICE: tc,
 						}),
 					)
+					waitForAnswererOnline(context.Background(), []string{host}, signalingCallQueue)
 					test.That(t, err, test.ShouldBeNil)
```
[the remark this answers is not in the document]
benjirewis: From your perspective, can the series of events you're describing here happen in production? 

I.e., is it possible that a machine can have an answerer that is blocked on `Answer()`, but, when a caller tries to connect to that machine right after another caller successfully connected, the signaling server still reports that the machine is offline?

rpc/wrtc_call_queue_mongodb.go
danielbotros: No particular reason, I'll just block since it's simpler

rpc/wrtc_call_queue_test.go
danielbotros: Actually in this test we are testing canceling the `SendOfferInit` context closes answerer responses so we never receive an offer to assert, and in the test below we are testing that we explicitly don't receive an offer bc we sent it to the wrong host (though we could test that the fake answerer that listens to that host does receive it)

rpc/wrtc_signaling_test.go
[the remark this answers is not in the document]
benjirewis: > So, it's possible but requires 3 connection attempts to come in within a few seconds of each other (like 2 secs or less) and some bad luck.

I _think_ this situation is going to be possible given some of the other issues we've seen. But, I'm also thinking that we can further consider the consequences of it when you tackle rate-limiting.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -91,6 +91,14 @@ var (
 		},
 	)
 
+	connectionEstablishmentExpectedFailures = statz.NewCounter0(
+		"signaling/connection_establishment_expected_failures",
+		statz.MetricConfig{
+			Description: "The total number of connection attempts that failed because the target robot was offline or does not exist.",
```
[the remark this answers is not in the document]
cheukt: nevermind, don't worry about this

PR #483, opened 2025-09-25
[RSDK-11963] Measure time for signaler call exchange

danielbotros:
### Description
[RSDK-11963](https://viam.atlassian.net/browse/RSDK-11963): Measure time for signaler call exchange
* Adds new metric `callExchangeDuration` which measures the duration of the call exchange attempt, recording which side SDK and org this exchange originated from, whether or not it failed, and from which perspective this time was recorded from (answerer or caller).

[RSDK-11963]: https://viam.atlassian.net/browse/RSDK-11963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rpc/wrtc_call_queue_mongodb.go
+++ b/rpc/wrtc_call_queue_mongodb.go
@@ -1290,20 +1290,7 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferDone(ctx context.Context, host, uu
 		return err
 	}
 
-	var call mongodbWebRTCCall
-	if err := updateResult.Decode(&call); err == nil {
-		if call.CallerDone && call.AnswererDone {
-			duration := float64(time.Since(call.StartedAt).Milliseconds())
-			var finalResult string
-			if call.CallerError != "" || call.AnswererError != "" {
-				finalResult = exchangeFailed
-			} else {
-				finalResult = exchangeFinished
-			}
-			callExchangeDuration.Observe(duration, call.SDKType, call.OrganizationID, finalResult)
-		}
-	}
-
+	recordExchangeDuration(updateResult)
 	return nil
 }
 
@@ -1314,8 +1301,7 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferError(ctx context.Context, host, u
 		{webrtcCallIDField, uuid},
 		{webrtcCallHostField, host},
 		{webrtcCallCallerDoneField, bson.D{{"$ne", true}}},
-	}, bson.D{{"$set", bson.D{{webrtcCallCallerErrorField, err.Error()}}}},
-		options.FindOneAndUpdate().SetReturnDocument(options.After))
+	}, bson.D{{"$set", bson.D{{webrtcCallCallerErrorField, err.Error()}}}})
 	if err := updateResult.Err(); err != nil {
 		// No matching documents is indicative of an inactive offer.
 		if errors.Is(err, mongo.ErrNoDocuments) {
@@ -1626,6 +1612,34 @@ func iceCandidateToMongo(i *webrtc.ICECandidateInit) mongodbICECandidate {
 	return candidate
 }
 
+func recordExchangeDuration(updateResult *mongo.SingleResult) {
+	var call mongodbWebRTCCall
+	if err := updateResult.Decode(&call); err != nil {
+		return
+	}
+
+	// We cannot consider an exchange finished until both sides are done
+	// because a state of done indicates that a side has finished sending
+	// all its candidates (or the client channel is ready). Recording on
+	// the first done would capture an incomplete exchange where one side
+	// has finished exchanging candidates but the other side has not, so
+	// the exchange is clearly not finished. By waiting until both sides
+	// are done, we ensure that we measure the full end-to-end signaling
+	// process from offer creation to both sides confirming completion.
+	if !call.CallerDone || !call.AnswererDone {
+		return
+	}
+
+	duration := float64(time.Since(call.StartedAt).Milliseconds())
+	var finalResult string
+	if call.CallerError != "" || call.AnswererError != "" {
+		finalResult = exchangeFailed
+	} else {
+		finalResult = exchangeFinished
+	}
+	callExchangeDuration.Observe(duration, call.SDKType, call.OrganizationID, finalResult)
+}
+
 // Close cancels all active offers and waits to cleanly close all background workers.
 func (queue *mongoDBWebRTCCallQueue) Close() error {
 	queue.cancelFunc()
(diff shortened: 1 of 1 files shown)


perf/statz/distributions.go
```diff
@@ -16,6 +16,9 @@ type Distribution struct {
 // LatencyDistribution is a basic latency distribution.
 var LatencyDistribution = DistributionFromBounds(0, 5, 25, 50, 75, 100, 200, 400, 600, 800, 1000, 2000, 4000, 6000)
 
+// ConnectionTimeDistribution is a basic distribution for connection times in seconds.
+var ConnectionTimeDistribution = DistributionFromBounds(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
```
benjirewis, line 20: Nice TIL about OC distributions. From what I can tell, this might be the first usage of that metric type we'll have in GCP, as that `LatencyDistribution` above is only used in example code. I'm curious how it'll appear in GCP or what it will be like to set up alerting policies against it.

benjirewis (commented):
Thanks! Some initial questions.

perf/statz/distributions.go
cheukt: seconds is too coarse - can we have more buckets? ideally granularity of 250 or 500ms. If that's too many buckets, I'd be ok removing some of the buckets in the 7/8/9s range.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -180,6 +180,33 @@ var (
...
+					Name:        "sdk_type",
+					Description: "The type of SDK attempting to connect.",
+				},
+				{
+					Name:        "organization_id",
+					Description: "The organization ID of the machine that is being connected to.",
+				},
+				{
+					Name:        "side",
+					Description: "The side of the call (caller or answerer).",
```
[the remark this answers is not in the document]
cheukt: same q

cheukt (changes requested):
looks good, a few questions/requests

rpc/wrtc_call_queue_mongodb.go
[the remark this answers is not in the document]
danielbotros: TBH I think I was mainly just unsure of whether we wanted it for just the caller or both the caller and answerer so I added it to both. I was thinking maybe it could be useful for getting an estimate on how much time a machine is spending trying to connect but since it's pretty much symmetrical w.r.t the caller I don't really think there's a need and I'll remove it.

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1222,6 +1263,7 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
 			}
 
 			if callResp.AnswererDone {
+				finalResult = exchangeSuccess
```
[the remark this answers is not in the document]
danielbotros: Yeah you're right I overlooked this case. I think there's not really a great way to determine if an exchange succeeded or not then, either that it finished or errored / timed out / was cancelled. I'll change it to `exchangeFinished`

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1180,16 +1208,24 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferInit(
 		defer cleanup()
 		defer close(answererResponses)
 
+		var finalResult string
+		defer func() {
+			duration := time.Since(call.StartedAt).Seconds()
+			callExchangeDuration.Observe(duration, sdkType, organizationID, finalResult)
```
benjirewis, line 1214: I _suspect_ you are still measuring only a subset of the call exchange with this value. Before I dive further into theories, what values are you seeing for `duration` in testing? What about same-subnet vs. different-subnet connections?

perf/statz/distributions.go
```diff
@@ -16,6 +16,10 @@ type Distribution struct {
 // LatencyDistribution is a basic latency distribution.
 var LatencyDistribution = DistributionFromBounds(0, 5, 25, 50, 75, 100, 200, 400, 600, 800, 1000, 2000, 4000, 6000)
 
+// ConnectionTimeDistribution is a basic distribution for connection times in seconds.
```
benjirewis, line 19: > in seconds.

I'm assuming you wanted this to be in milliseconds, but it looks like you're still calculating `duration` with a `.Seconds()` method invocation instead of `.Milliseconds()`.

danielbotros: Thanks good catch

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1249,16 +1277,34 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferUpdate(ctx context.Context, host,
...
+	if err := updateResult.Decode(&call); err == nil {
+		if call.CallerDone && call.AnswererDone {
+			duration := float64(time.Since(call.StartedAt).Milliseconds())
+			var finalResult string
+			if call.CallerError != "" || call.AnswererError != "" {
+				finalResult = exchangeFailed
+			} else {
+				finalResult = exchangeFinished
+			}
+			queue.logger.Infow("call exchange completed", "duration_secs", duration)
```
cheukt, line 1303: probably should remove before merge?

rpc/wrtc_call_queue_mongodb.go
cheukt: is that different from when the caller gets a response and is "done"? basically I'm asking why the places we're adding the metric don't match 1:1 to where we're incrementing for the connection success/failure result

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1727,22 +1776,41 @@ func (resp *mongoDBWebRTCCallOfferExchange) AnswererRespond(ctx context.Context,
...
+		if call.CallerDone && call.AnswererDone {
+			duration := float64(time.Since(call.StartedAt).Milliseconds())
+			var finalResult string
+			if call.CallerError != "" || call.AnswererError != "" {
+				finalResult = exchangeFailed
+			} else {
+				finalResult = exchangeFinished
+			}
+			callExchangeDuration.Observe(duration, call.SDKType, call.OrganizationID, finalResult)
+		}
```
cheukt, line 1812: this block seems repeated from above, can we refactor it out?

cheukt (approved):
LGTM!

rpc/wrtc_call_queue_mongodb.go
```diff
@@ -1269,7 +1314,8 @@ func (queue *mongoDBWebRTCCallQueue) SendOfferError(ctx context.Context, host, u
 		{webrtcCallIDField, uuid},
 		{webrtcCallHostField, host},
 		{webrtcCallCallerDoneField, bson.D{{"$ne", true}}},
-	}, bson.D{{"$set", bson.D{{webrtcCallCallerErrorField, err.Error()}}}})
+	}, bson.D{{"$set", bson.D{{webrtcCallCallerErrorField, err.Error()}}}},
+		options.FindOneAndUpdate().SetReturnDocument(options.After))
```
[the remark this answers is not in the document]
danielbotros: I think I just threw it onto every `FindOneAndUpdate` I touched for this PR tbh lol but we don't need it for this one or in `AnswererRespond` so I'm removing

PR #484, opened 2025-09-25
Reorder attempt and fail metrics

danielbotros:
### Description

We want expected connection failures to be a subset of all connection attempts so I re-ordered when each metric gets ordered so that this is the case


cheukt:
can we track sdk type and org id now too?

cheukt (approved):
thanks, LGTM!

PR #555, opened 2026-04-18
Add non-blocking govulncheck CI workflow

NickPPC:
## Summary

Add a govulncheck job to CI that runs on push to `main` and on every pull request. Marked `continue-on-error: true` so findings are surfaced in CI without blocking merges while we triage and upgrade dependencies.

## Changes

- **.github/workflows/govulncheck.yml**: New workflow — checks out code, installs the Go version from `go.mod`, runs `golang/govulncheck-action`. Pinned to commit SHAs, read-only workflow permissions, uses `pull_request` (not `pull_request_target`) so no secrets are exposed to forked PRs.

## Testing

- Workflow will execute on this PR itself — verify the job runs, reports vulnerabilities (expected: ~21 symbol-level findings on current tree), and does not fail the required status checks.
- Confirm branch protection on `main` is NOT configured to require the `govulncheck` check, otherwise it would still effectively block merges.

## Follow-ups

- Separate PRs will clear the findings: Go 1.25.9 + `otel/sdk` v1.40.0 (biggest win), then `lestrrat-go/jwx` v1.2.29, then an investigation of `pion/dtls` and `aws-sdk-go`.

## Claude Code Prompts Used

- "Add govulncheck to CI. It should be not blocking for now; assess the dependencies we can upgrade in this repo, focus on security risks"
- "Proceed with the 3 PRs and the investigation"

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -14,8 +14,6 @@ jobs:
   govulncheck:
     name: govulncheck
     runs-on: ubuntu-latest
-    # Non-blocking: surface vulnerabilities without failing required status checks.
-    continue-on-error: true
     timeout-minutes: 15
     steps:
       # actions/checkout@v6.0.2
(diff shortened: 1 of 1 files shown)


NickPPC:
We are centralizing this in https://github.com/viamrobotics/overwatch

PR #559, opened 2026-05-06
RSDK-13933 - Ratelimiter docs are sometimes persisted with unevaluated expressions

cheukt:
The bug: In MongoDB, regular update operators like $setOnInsert don't evaluate aggregation expressions — they store the BSON they receive verbatim. So this:

```
bson.M{"$setOnInsert": bson.M{
    "expires_at": bson.M{
        "$dateAdd": bson.M{"startDate": "$$NOW", "unit": "second", "amount": 60.0},
    },
}}
```
doesn't compute a future timestamp. It writes expires_at as the literal sub-document `{"$dateAdd": {"startDate": "$$NOW", "unit": "second", "amount": {"$numberDouble": "60.0"}}}`. Aggregation operators like `$dateAdd` and system variables like `$$NOW` are only evaluated inside aggregation contexts (pipelines, $expr, etc.), not inside regular update operators.

Issue:

A sub-document is invisible to TTL, so any document that ended up with the literal would never get cleaned up.

tested that the test failed without the fix and succeeds with

### DIFF
--- a/rpc/wrtc_rate_limiter.go
+++ b/rpc/wrtc_rate_limiter.go
@@ -94,20 +94,27 @@ func NewMongoDBRateLimiter(
 // The update creates a new array that adds the current timestamp and removes old timestamps outside the window.
 // This prevents race conditions and keeps the requests array bounded.
 func (rl *MongoDBRateLimiter) Allow(ctx context.Context, key string) error {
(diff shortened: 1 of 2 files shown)


danielbotros (approved):
LGTM! this was silly lol

PR #560, opened 2026-05-13
APP-16300: Implement rate limit sampling on a per root span name basis

jmatth:
This adds a new sampler implementation for OpenCensus tracing. It is based on [rate limiting sampling][1] with the added complexity of performing the sampling on a per root span name basis. For example, consider a server that

- Receives 8 requests to `/a` per second
- Receives 2 requests to `/b` per second
- Is configured to sample one span per second

With simple rate limit sampling, only 1 of the 10 total requests per second would be sampled. With this implementation it will sample 2, one from each route. This will ensure we have some minimum level of coverage for each code path at the cost of increasing the total number of sampled spans as a function of the number of total root span names across the application.

On web servers like app most root span names will be based on the route a request was sent to.

This behavior is off by default but can be configured with a new environment variable.

[1]: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#ratelimiting-sampler-implementation-details

### DIFF
--- a/perf/root_name_ratelimit_sampler.go
+++ b/perf/root_name_ratelimit_sampler.go
@@ -9,10 +9,10 @@ import (
 	"go.opencensus.io/trace"
 )
 
-// NewRouteRateLimitingSampler creates a [trace.Sampler] that samples x spans
+// NewRootNameRateLimitingSampler creates a [trace.Sampler] that samples x spans
 // per second per root span name. The first encountered root span of each name
 // is always sampled.
-func NewRouteRateLimitingSampler(perSec float64) trace.Sampler {
+func NewRootNameRateLimitingSampler(perSec float64) trace.Sampler {
 	if perSec <= 0 {
 		return trace.NeverSample()
 	}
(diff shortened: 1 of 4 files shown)


perf/root_name_ratelimit_sampler.go
```diff
@@ -0,0 +1,94 @@
...
+		lastSampleAtomic, present := lastSampledMicrosByName.Load(sp.Name)
+		nowNanos := time.Now().UnixNano()
+		var sample bool
+		if present {
+			// Compute the sampling probability based on the seconds since the last
+			// time we sampled as detailed here:
+			// https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#when-does-opencensus-sample-traces
+			lastSample := lastSampleAtomic.Load()
+			elapsedSec := max(float64(nowNanos-lastSample)/1e9, 0)
+			samplingProb := min(elapsedSec*perSec, 1)
```
dgottlieb, line 45: This is the most interesting part of the change. I'm not sure I exactly buy the analysis offered by that link (but to be fair, the claims are imprecise). But certainly nothing that's going to be grossly over or under counting. But I did have to convince myself this was approximately the same as just doing a plain old "elapsedSec > configuredPeriodSec".

I suppose the author was very motivated to minimize the amount of lost races with a CompareAndSwap (read: wasted "expensive" atomic operation)?

dgottlieb (approved):

perf/root_name_ratelimit_sampler.go
jmatth: That struck me as odd too and I had about the same conclusion: the author is trying to avoid threads regulary racing for the atomic every x seconds. I admittedly didn't dig too far into it though since for this use case "mostly correct" is good enough

PR #561, opened 2026-05-15
RSDK-13822 Add stun:turn.viam.com:443 to default ICE servers

benjirewis:
RSDK-13822

## What

Adds `stun:turn.viam.com:443` to the list of default ICE servers.

## Why

Adding that extra default ICE server will let `viam-server` and Golang SDK clients (e.g. `viam-server` connections to remote) gather STUN candidates from `turn.viam.com:443`. This is especially useful when STUN candidates _cannot_ be gathered from the other default ICE server: `stun:global.stun.twilio.com:3478`. We have found firewalls in the wild that disallow STUN traffic to the Twilio URL but allow it to our TURN server.

**There is a notable side-effect of this change:**

`viam-server` and Golang SDK clients will gather an "extra" STUN candidate in the event that both `global.stun.twilio.com` and `turn.viam.com` are reachable (this is most of the time). That extra STUN candidate _will_ be transmitted through MongoDB for signaling and _will_ be used in connectivity checks by the remote. The two candidates may look something like:

```
{"FoundAt":"2026-05-15T18:50:23.360999936Z","CandType":"server-reflexive","IP":"73.47.51.100","Port":53533},
{"FoundAt":"2026-05-15T18:50:23.359000064Z","CandType":"server-reflexive","IP":"73.47.51.100","Port":63389}
```

You'll see that the IPs are identical but the ports are _not_. IIUC this is because pion creates two distinct `net.Conn`s: one to Twilio and one to our Viam TURN server. Since these are distinct outgoing connections, NAT will assign them two different ports. I was tempted to add some logic to only submit one of two STUN candidates with identical IPs ("dedupe"), but I don't know what effect that will have in "weird" network topologies. Maybe there _is_ a case where a STUN candidate with address 73.47.51.100:53533 is unreachable but a STUN candidate with 73.47.51.100:63389 is reachable? Since my change does not seem to slow connection establishment times in my testing, I'm tempted to do the "simple" thing here, and add deduping if we find connection times to be slow later on.

## Testing

I ensured that if I drop gathered host candidates and STUN candidates gathered from Twilio, `viam-server` and a Golang SDK instance can connect solely based on the STUN candidates gathered from turn.viam.com. `viam-server` and the Golang SDK instance could _not_ connect in that case without this change. Furthermore, I connected 100 times with this change and without it, and found an almost identical average connection establishment time (~1.2 seconds in my testing).

### DIFF
--- a/rpc/wrtc_peer.go
+++ b/rpc/wrtc_peer.go
@@ -38,6 +38,9 @@ var DefaultICEServers = []webrtc.ICEServer{
 	{
 		URLs: []string{"stun:global.stun.twilio.com:3478"},
 	},
+	{
+		URLs: []string{"stun:turn.viam.com:443"},
+	},
 }
 
 // DefaultWebRTCConfiguration is the standard configuration used for WebRTC peers.


danielbotros (approved):
LGTM! If my understanding of the WebRTC spec is correct (and pion implements it correctly) I think the extra STUN candidates _shouldn't_ be too big of a deal because we run the checks concurrently and nominate the first successful pair, so it doesn't seem like extra candidates should have a noticeable impact on connection establishment time in the case we can connect using `srflx` (or lower priority ones e.g. `host`) candidates since we may find a viable candidate pair in the same amount of time as before; maybe we end up firing off a few more checks in the mean time before one is marked as nominated? The "queue"  of candidates is longer but the index in the queue we find our nominated candidate doesn't change (unless maybe we need a relay candidate; see below). Your experiment seems to back this up as well which is encouraging. 

Another experiment I think could be worth trying is measuring connection establishment time where the only candidate pair that we can connect with are ones with a relay candidate(s). Since relay candidates are lowest priority I wonder if there would be a measurable latency in connection establishment time because the candidate check scheduler will start firing off the relay checks a bit later than before due to firing the additional STUN candidate checks first. It still probably doesn't warrant the deduping complexity though. I don't think using the new `force_relay` dial option would be useful here since I think it just strips the STUN candidates BTW.

rpc/wrtc_peer.go
```diff
@@ -38,6 +38,9 @@ var DefaultICEServers = []webrtc.ICEServer{
 	{
 		URLs: []string{"stun:global.stun.twilio.com:3478"},
 	},
+	{
+		URLs: []string{"stun:turn.viam.com:443"},
```
jmatth, line 42: Since `URLs` is a slice I'm curious what would happen if you put both URLs in a single struct instead of separate ones:

```
URLs: []string{"stun:global.stun.twilio.com:3478", "stun:turn.viam.com:443"}
```

Edit: It's hard to track down since the field(s) get passed through multiple layers of sanitation and validation but it looks like the struct thing is just a convenient way to reuse shared auth and the URLs ultimately just end up in a flat list passed to the ICE gatherer

jmatth (approved):

benjirewis:
> LGTM! If my understanding of the WebRTC spec is correct (and pion implements it correctly) I think the extra STUN candidates shouldn't be too big of a deal because we run the checks concurrently and nominate the first successful pair, so it doesn't seem like extra candidates should have a noticeable impact on connection establishment time in the case we can connect using srflx (or lower priority ones e.g. host) candidates since we may find a viable candidate pair in the same amount of time as before; maybe we end up firing off a few more checks in the mean time before one is marked as nominated? The "queue" of candidates is longer but the index in the queue we find our nominated candidate doesn't change (unless maybe we need a relay candidate; see below). Your experiment seems to back this up as well which is encouraging.

Thanks for thinking through it with me! This was my reasoning, too. WebRTC as-described-by-the-spec feels built to handle lots of ICE candidates. I tested with a Typescript client, too, and found that the extra STUN candidate only  _lowered_ the connection establishment time.

```
Without my change:

Results over 100 iterations (ms):
  avg: 838.46
  median: 748.57
  p90: 888.76
  min: 616.93
  max: 5157.80

With my change:

Results over 100 iterations (ms):
  avg: 776.39
  median: 761.60
  p90: 887.76
  min: 613.51
  max: 1444.23
```

I'm assuming that diff is due to the one 5s outlier in the OG data, but I'll take it to mean that there's not an obvious negative impact in most connection scenarios.

> I wonder if there would be a measurable latency in connection establishment time because the candidate check scheduler will start firing off the relay checks a bit later than before due to firing the additional STUN candidate checks first.

Good idea; will test.

rpc/wrtc_peer.go
benjirewis: > end up in a flat list passed to the ICE gatherer

That's what I saw, too.

```golang
	var validatedServers []*stun.URI
	if len(opts.ICEServers) > 0 {
		for _, server := range opts.ICEServers {
			url, err := server.urls()
			if err != nil {
				return nil, err
			}
			validatedServers = append(validatedServers, url...)
		}
	}
```

Semantically, it also made sense to me to mark `stun:turn.viam.com` as a separate ICE server from `stun:global.stun.twilio:3478` and not just a different URL for the _same_ ICE server 🤷🏻  (that's how I interpreted the `URLs` field).

benjirewis:
> Good idea; will test.

Tested with TS (Safari) and Golang clients, and, although connection establishment time is 500ms-1s higher due to only relay candidates being allowed, I don't see any tangible timing difference to connecting through relay to a `viam-server` with my change and without it.

I'm going to merge this change into main and watch the connection establishment time metric in GCP carefully for spikes. We _could_ do a change [like this](https://github.com/benjirewis/goutils/commit/f48f44f485d278623963f415219be60e1ed5cc33), where we use only `stun:turn.viam.com` if the Twilio URL reportedly _doesn't work_. I don't like that solution because it's more complex than the one in this PR and there's a window (5 minutes in that impl) where we wouldn't know that the Twilio URL wasn't working.

PR #563, opened 2026-05-26
Fix rate limit sampler

jmatth:
This fixes the per root name rate limit sampling introduced in #560. That implementation relied on a formula for rate limit sampling in the uprstream OpenCensus documentation that required knowing the requests per second that the service would receive. Since we don't know that it did not behave as expected in production. This rewrite uses a new implementation that just tracks when the next sampling should occur. Some specifics:

- I'm trying to avoid a situation where all sampling occurs on a regular interval and causes bursts of CPU and network usage. Suppose we set the sampling rate at 1 / 10s. I would expect that most of the routes get hit for the first time and therefore produce a trace at about the same time when the instance comes online and the load balancer starts routing traffic, and therefore we would see burts of CPU usage and tracing exports every 10s. Ideally this usage would be evenly spread out across the sampling interval. To make this even resourse usage more likely this implementation randomizes the time between first and second samples to be between 0 and the configured interval. An alternate solution would be to randomize the time between every sample to between 0 and 2 * the configured interval, which should average out to the correct number of samples.
- The `x/time` package provides an implementation of a token bucket rate limiter that could be used in place of the custom timestamps. You can see a version of the code that uses it [here][1]. I went with this timestamp implementation because it has better performance when multiple goroutines are racing for the root span name, but benchmarking showed that using `x/time` still isn't terribly slow overall even with 8 goroutines all racing for the same lock. I could be worth sacrificing performance for code readability.

```
%> go test -bench=. 
goos: linux
goarch: amd64
pkg: go.viam.com/utils/perf
cpu: AMD Ryzen AI 9 HX 370 w/ Radeon 890M           
Benchmark/Atomics-24            15785199                70.17 ns/op
Benchmark/Locks-24               1348744               882.9 ns/op
PASS
ok      go.viam.com/utils/perf  2.309s
```

[1]: https://github.com/viamrobotics/goutils/commit/cbdf9b2f61ed2741567ed07fb5f3178732c03657


perf/root_name_ratelimit_sampler_test.go
```diff
@@ -59,10 +59,7 @@ func TestRouteRateLimitingSampler(t *testing.T) {
 
 	t.Run("first root span for each unique name is always sampled", func(t *testing.T) {
 		synctest.Test(t, func(t *testing.T) {
```
benjirewis, line 61: 🎉

benjirewis (approved):
Readability-wise I think what you have is well-documented and simple enough that you don't have to use `x/time`.

The initial jitter between first and second sample seems fine to me, too (as compared to random([0, 2*requestedInterval]) between each sample).

PR #568, opened 2026-06-15
rpc: refresh time-limited TURN credentials on long-lived relayed connections

danielbotros:
## Problem

When a WebRTC connection is relayed through a TURN server, pion keeps the TURN allocation alive for the life of the `PeerConnection`, refreshing permissions with the credentials baked in at dial time. Signaling servers typically mint **time-limited** TURN credentials (the username is a unix expiry timestamp), so on a long-lived relayed connection those credentials eventually expire and the TURN server rejects the refresh with `401 Unauthorized` — and if the relay is the selected path, the connection drops. Nothing currently re-fetches credentials for an established connection.

## Change

Adds a per-connection worker (`maintainTURNCredentials`, `wrtc_turn_refresh.go`) spawned by `dialWebRTC` that:

1. Reads the soonest credential expiry off the live `PeerConnection` (parsed from the ICE-server username).
2. Before it lapses (default: halfway to expiry), re-queries the signaling server's `OptionalWebRTCConfig` for fresh ICE servers.
3. Applies them with `PeerConnection.SetConfiguration` and triggers an **ICE restart** so a new TURN allocation is established with the fresh credentials — without tearing down the connection.

It is a no-op for connections that aren't relayed through a credential-expiring TURN server (no timestamp username → nothing scheduled), so callers don't need to know in advance whether they'll connect over a relay.

To support a client-initiated ICE restart, `ConfigureForRenegotiation` now also returns a `renegotiate` function (either peer can initiate), and renegotiation SDP sends wait for ICE gathering to complete so an ICE-restart's re-gathered candidates are all included (there is no trickle path over the negotiation channel). This wait is a no-op for non-ICE renegotiations (e.g. adding a track), where gathering is already complete.

## Testing

- `go build ./...`, `go vet ./rpc/`, `gofmt` — clean
- `go test ./rpc/` — full suite passes (incl. renegotiation / WebRTC dial)

## Open items (draft)

- Needs validation against a live relay (cred expiry → ICE restart → uninterrupted traffic).
- Review renegotiation concurrency: `OnNegotiationNeeded` was previously server-only; both peers can now initiate, so simultaneous (glare) renegotiations want a hardening pass.

### DIFF
--- a/rpc/wrtc_peer.go
+++ b/rpc/wrtc_peer.go
@@ -219,15 +219,15 @@ func newPeerConnectionForClient(
 	config webrtc.Configuration,
 	disableTrickle bool,
 	logger utils.ZapCompatibleLogger,
-) (*webrtc.PeerConnection, *webrtc.DataChannel, error) {
+) (*webrtc.PeerConnection, *webrtc.DataChannel, renegotiateFunc, error) {
 	webAPI, err := newWebRTCAPI(logger)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	peerConn, err := webAPI.NewPeerConnection(config)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 	var successful bool
 	defer func() {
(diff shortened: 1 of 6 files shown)


rpc/wrtc_peer.go
```diff
@@ -453,15 +491,9 @@ func ConfigureForRenegotiation(
...
-			encodedSDP, err := EncodeSDP(peerConn.LocalDescription())
-			if err != nil {
-				logger.Errorw("renegotiation: error encoding SDP", "error", err)
-				return
-			}
-			if err := negotiationChannel.SendText(encodedSDP); err != nil {
-				logger.Errorw("renegotiation: error sending SDP", "error", err)
-				return
-			}
+			sendCtx, cancel := context.WithTimeout(context.Background(), getDefaultOfferDeadline())
```
danielbotros, line 494: confirm context deadline

danielbotros:
Closing: the dialer-side in-band refresh here has a glare gap that can't be fixed via perfect negotiation on our webrtc fork (no local-offer rollback). Replacing with a single-initiator (server-driven) design in a new draft.

PR #571, opened 2026-06-17
Log noisy offline no connection errors at most once if conn state is TRANSIENT_FAILURE or CONNECTING

aldenh-viam:
I moved `ConnectivityState` from RDK to here, and will update the other PRs that use it.

This helps reduce some log spam when we go offline. Previously, every minute we'd get:
```
INFO    rdk.networking.signaler.external        rpc/wrtc_signaling_answerer.go:208      Message logged 120 times in past 1m0s: failed to communicate with signaling server      {"error":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup ...
INFO    rdk.networking.signaler.external        rpc/wrtc_signaling_answerer.go:208      failed to communicate with signaling server     {"error":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup...
INFO    rdk.networking.signaler.external        rpc/wrtc_signaling_answerer.go:208      failed to communicate with signaling server     {"error":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup...
INFO    rdk.networking.signaler.external        rpc/wrtc_signaling_answerer.go:208      failed to communicate with signaling server     {"error":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup...
```

### DIFF
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -174,14 +174,6 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 		ans.connMu.Lock()
 		conn := ans.conn
 		ans.connMu.Unlock()
-		// do not attempt to answer if the connection is not in a ready state.
-		// we could exclude the connectivity.Connecting check and move this block closer to the Answer call for lower latency,
-		// but we retry often enough (2x/sec) that it shouldn't make a noticeable difference.
-		if cs, ok := grpchelpers.ConnConnectivityState(conn); ok {
-			if cs == connectivity.TransientFailure || cs == connectivity.Connecting {
-				return nil, grpchelpers.ErrConnNotReady
-			}
-		}
 		client := webrtcpb.NewSignalingServiceClient(conn)
 		md := metadata.New(nil)
 		md.Append(RPCHostMetadataField, ans.hosts...)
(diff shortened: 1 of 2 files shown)


rpc/wrtc_signaling_answerer.go
```diff
@@ -172,6 +174,14 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 		ans.connMu.Lock()
 		conn := ans.conn
 		ans.connMu.Unlock()
+		// do not attempt to answer if the connection is not in a ready state.
+		// we could exclude the connectivity.Connecting check and move this block closer to the Answer call for lower latency,
+		// but we retry often enough (2x/sec) that it shouldn't make a noticeable difference.
+		if cs, ok := grpchelpers.ConnConnectivityState(conn); ok {
```
benjirewis: So what stops us from using the recommended `WaitForReady` pattern here? We'd be using a context whose deadline is close enough that we'd just hit `context deadline exceeded` at some still-noisy cadence? Can we just use `ans.bgWorkers.Context()` to wait for as long as the answerer is alive for the connection to be ready?

benjirewis (commented):
Thanks for looking into this; the "anti-pattern" piece is interesting.

cheukt (approved):

rpc/wrtc_signaling_answerer.go
aldenh-viam: With `WaitForReady` it only waits until the ctx times out, so it would still spam the logs, just with a different message (`context deadline exceeded`) and less often (the ctx timeout period). After the `client.Answer` we'd need to check if it's that err, and then check the connection to suppress the error only if offline, so it wouldn't look too different from this.

IIUC checking the connection before making the RPC call is considered an "antipattern" primarily because it's not totally reliable (check passes->lose internet->RPC fails, or vice-versa), but that shouldn't really matter for us.

benjirewis: Sounds good.

> IIUC checking the connection before making the RPC call is considered an "antipattern" primarily because it's not totally reliable (check passes->lose internet->RPC fails, or vice-versa), but that shouldn't really matter for us.

Surely `WaitForReady` must sort of do the same thing under the hood? Basically a `for` loop checking the same thing as us? By the time the RPC request is sent, one could always have lost internet, I suppose 🤷🏻 .

benjirewis (approved):

aldenh-viam:
New design: https://github.com/viamrobotics/goutils/pull/571/commits/ef4d66a801997c6e7bd0d73f8b5264035e55e71e, based on discussion here: https://github.com/viamrobotics/rdk/pull/6127#discussion_r3430532779

Now logs all new unique errors from `answerClient, err := client.Answer(answerCtx)` at most once for each offline period, instead of checking the conn state and skipping the RPC call if maybe offline. 

Let me know if you prefer this design or the previous?

cheukt:
it'd be nice to have the loop/err map as a helper but probably not worth the squeeze. design lgtm

PR #579, opened 2026-06-29
Add granular additive WebRTC dial options

danielbotros:
## Description

`WithWebRTCOptions(...)` does a replacement of the whole `DialWebRTCOptions` struct. So setting a single flag like `ForceP2P` clobbers everything else, e.g. `SignalingServerAddress`, `SignalingCreds`, `Config`, the TURN knobs etc...

Callers work around this today by re-passing the whole struct and manually constructing `DialWebRTCOptions` with all the desired fields. This PR adds granular dial options that mutate only their own field on `o.webrtcOpts`, leaving every other WebRTC option in place:

- `WithForceP2P()`
- `WithForceRelay()`
- `WithTurnURI(uri)`
- `WithTurnScheme(scheme)`
- `WithTurnTransport(transport)`
- `WithTurnPort(port)`

Semantics: **additive** when applied after `WithWebRTCOptions(...)`, **last-writer-wins** if two touch the same field.

Ex.
```go
rpc.Dial(ctx, address, logger,
    rpc.WithEntityCredentials(entity, creds),
    rpc.WithWebRTCOptions(rpc.DialWebRTCOptions{SignalingServerAddress: signalingAddr}),
    rpc.WithForceRelay(),
    rpc.WithTurnURI("turn:turn.viam.com:443"),
)
```

### DIFF
--- a/rpc/dial_options.go
+++ b/rpc/dial_options.go
@@ -191,6 +191,13 @@ func WithTLSConfig(config *tls.Config) DialOption {
 
 // WithWebRTCOptions returns a DialOption which sets the WebRTC options
 // to use if the dialer tries to establish a WebRTC connection.
+//
+// Note: this replaces the entire DialWebRTCOptions struct in one shot, so it
+// clashes with the granular dial options (WithForceP2P, WithForceRelay,
+// WithTurnURI, etc.). Combining it with any of those is order-dependent —
+// whichever DialOption is applied last wins — which can produce unexpected
+// behavior. For additive tweaks that leave other fields intact, prefer the
+// granular options.
 func WithWebRTCOptions(webrtcOpts DialWebRTCOptions) DialOption {
 	return newFuncDialOption(func(o *dialOptions) {
 		o.webrtcOpts = webrtcOpts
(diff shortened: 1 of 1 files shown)


cheukt (approved):
should we deprecate WithWebRTCOptions? it'd probably be way less confusing that way.

we should also just note somewhere that WithWebRTCOptions clashes with each of these and unexpected behavior may happen

danielbotros:
> should we deprecate WithWebRTCOptions? it'd probably be way less confusing that way.
> 
> we should also just note somewhere that WithWebRTCOptions clashes with each of these and unexpected behavior may happen

I noted the collision between the dial options and documented that they need to be set after `WithWebRTCOptions` if both are used, and also decided against deprecating it. `WithWebRTCOptions` is used in a few places in rdk to bulk-set fields that go together e.g. `SignalingServerAddress` and `SignalingCreds`.  A granular knob per field doesn't make sense there, and we wouldn't tweak those individually after construction anyway. If the signaling config changes we'd just build a new `webrtcOpts`, so there isn't an obvious way to move away from it to me.  The `WithForceRelay/P2P/TURN*` options however, we would want to add on top of the dialer-provided WebRTC options without rebuilding the whole struct. It's not ideal but an improvement over having to reconstruct the `webrtcOpts` to use them and mainly for internal use so I think this is acceptable.

PR #583, opened 2026-07-08
APP-14871: Report WebRTC client connection metadata

danielbotros:
### Description

[APP-14871](https://viam.atlassian.net/browse/APP-14871) Track coturn usage / WebRTC client connection telemetry

**Proto** — adds to `SignalingService`:
- `ReportConnectionMetadata` unary RPC, plus `ICECandidateType` (host / stun / relay), `DialStage`, and `ConnectionSignalingPath` enums and a `ConnectionCandidate` message.
- `failure_code` is an `int32` gRPC status code (app translates it to a human-readable name)

**Client (`dialWebRTC`)** — after a dial succeeds or fails, reports information about the attempt best-effort to app over a connection the client opens to app itself:
- One report per logical dial. The default dialer races mDNS and cloud signaling in parallel; a per-dial collector carried on the dial context gathers the connection attempt reports from every racing path and flushes the furthest-progressed.  On a successful dial the losing/cancelled paths' failure reports are suppressed (no spurious `Canceled` failures), and on a failed dial the furthest failure is reported. 
- Inspects the WebRTC stats report to find the selected candidate pair and classifies each`local` (the dialing SDK) and `remote` (the peer) as host, stun, or relay.
  - For relay candidates the relay server address is included so the signaling server can attribute the relay to a provider (Viam coturn vs Twilio).
- `reached_stage`: the furthest dial checkpoint, tracked exactly as the dial advances (`SIGNALING_CONNECTED → CONFIG_FETCHED → OFFER_SENT → ANSWER_RECEIVED → ICE_CONNECTED → DTLS_CONNECTED → READY`). `READY` means success; any earlier value is where a failed dial stopped, so failures are attributable (answerer never replied vs ICE/NAT vs DTLS vs data channel).
- `failure_code`: the gRPC status code of the dial failure; set only on failure.
- `signaling_path`: how the dial was signaled, derived from the signaling address, reported on every outcome. Known app signaling hosts are the `CLOUD_SIGNALED`; mDNS discovery is `MDNS_LOCAL` and everything else is `LOCAL`
- `duration_ms` (dial start → ready/failure). 
- The dialing SDK's type and version ride in the `viam_client` gRPC metadata header that every SDK's client interceptor sets, so the report connection carries them automatically and app reads them there.
- The reporting code lives in its own `wrtc_client_report.go`. The report is flushed in the background with a 5s timeout detached from the dial context so reporting never adds latency to the dial's return, and cancelled/timed-out dials still report.

**Tests** — added coverage for the classification and report-selection logic: signaling-path classification, candidate classification, app-target derivation, and the outcome-aware report selection (success sends `READY`; success-with-only-failure-reports is suppressed; failure sends the furthest), plus an integration test asserting exactly one `ReportConnectionMetadata` per dial.

App side changes found [here](https://github.com/viamrobotics/app/pull/11316); rdk wiring [here](https://github.com/viamrobotics/rdk/pull/5853).

### Testing
Brought up the local app + dedicated signaling server on this branch with `--output-telemetry` (statz metrics dump to the container logs), pointed at the local goutils. Registered a `viam-server` built from [rdk #5853](https://github.com/viamrobotics/rdk/pull/5853) as a machine, then drove a range of dial outcomes with a goutils WebRTC client and confirmed each produced the expected metrics in the app.

| Scenario | how | `signaling_path` | `outcome` / `reached_stage` | result |
|---|---|---|---|---|
| Cloud-signaled success | dial the app signaling server, mDNS off | `local`¹ | `success` / `ready` | `candidate_type` host↔host, `duration_ms` recorded |
| mDNS success (forwarded by the machine) | default dial; machine advertises mDNS | `mdns_local` | `success` / `ready` | `candidate_type` host↔host, `duration_ms` recorded, exactly one report |
| Offline answerer | dial with no machine online | `local` | `failure` / `offer_sent` | `connection_failure {failure_code=NotFound}` |
| Relay-only, no reachable TURN | force ICE relay-only | `local` | `failure` / `config_fetched` | `connection_failure {failure_code=Unknown}` |
| Failed mDNS dial (**forwarded**) | mDNS + relay-only | `mdns_local` | `failure` / `config_fetched` | `connection_failure {failure_code=Unknown}` |

¹ `cloud_signaled` requires the signaling address to be an app host (`app.viam.com` / `app.viam.dev`); dialing `localhost` classifies as `local`. The rest of the classification path is exercised.

Tested end-to-end:
- One report per dial: the parallel mDNS+cloud race emits a single report for the dial (the furthest-progressed path); the losing path's `Canceled` report is suppressed on success. 
- Forwarding: locally-signaled reports, both successes **and** failures, reach the app with `signaling_path=mdns_local/local`, exercising the machine-side forwarder.
- Success reports carry the selected candidate pair (local/remote → host/stun/relay) and `duration_ms`
- Failure reports carry` failure_code` and `reached_stage`
- sdk_version is the `go.viam.com/rdk` module version; sdk_type is the SDK language (go).

Example app metrics for the mDNS success:
```
connection_outcome        {org, signaling_path: "mdns_local", outcome: "success"} = 1
connection_candidate_type {org, side: "local",  candidate_type: "host"}           = 1
connection_candidate_type {org, side: "remote", candidate_type: "host"}           = 1
connection_sdk_type       {org, sdk_type: "go"}                                    = 1
connection_duration_ms    {sdk_type: "go", signaling_path: "mdns_local"} count=1 sum=11ms
```

[APP-14871]: https://viam.atlassian.net/browse/APP-14871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rpc/dial.go
```diff
@@ -44,6 +44,14 @@ func dialInner(
...
 	}
 
+	// One logical dial can fan out into several WebRTC attempts through different signaling paths
+	// (mDNS to local signaling	, direct to local signaling, and through cloud signaling). A collector
+	// shared via the context lets them report exactly once: on completion we emit a single report
+	// for the furthest-progressed attempt and close the held signaling connections.
+	collector := &dialReportCollector{ctx: ctx, host: address, logger: logger}
+	defer collector.flush()
+	ctx = contextWithReportCollector(ctx, collector)
+
```
danielbotros: I don't like this extra layer of complexity, but without it we always send two reports per "logical" dial because we race mDNS / dialing the signaling server and each attempt reports (one failure report for every successful dial since one side fails). If we re-use a cached connection, I think we may also send two failed reports despite a successful connection. 

I also don't love stuffing it into the context, but the alternatives were putting it in the dial options (it's not really a user configurable option) or threading an extra param through all the dial helpers.

rpc/server.go
```diff
@@ -718,6 +721,12 @@ func NewServer(logger utils.ZapCompatibleLogger, opts ...ServerOption) (Server,
 			server.signalingCallQueue = signalingCallQueue
 			server.signalingServer = NewWebRTCSignalingServer(signalingCallQueue, nil, logger,
 				defaultHeartbeatInterval, internalSignalingHosts...)
+			if externalAnswerer != nil && len(externalSignalingHosts) > 0 {
+				// The external answerer connects to the cloud signaling server, so re-use its connection to forward
+				// metadata about connections through the internal signaling server to the cloud signaling server.
+				server.signalingServer.SetConnectionMetadataHandler(
+					connectionMetadataForwarder(externalAnswerer.currentConn, externalSignalingHosts[0], logger))
+			}
```
danielbotros: We hijack the external signaler's app conn exposed through a new method to be able to forward connection metadata reports from local signaling to cloud signaling so the only dial attempt reports we miss out on are truly offline ones. If we care about offline connection attempts that may eventually make it to app we could write to disk and push later.

The external signaler's dial opts hold the shared app conn created in `rdk` (or will create it's own if one is not passed in) but there is not a good way to retrieve it from the dial option since we have to `apply` it to access it from it's internal `options` struct. This way allows us to re-use the same conn but is kind of a gross hack. We could add a separate dial option for it but it would be technically redundant. Open to suggestions.

danielbotros:
Q's from old pr:

>[q] What are the downsides of just sending up the entire peer stats object? I guess that's a pretty big object with a bunch of info we might not care about right now... My worry is that we will care about more peer connection over time and the proto schema you've designed here will need to be updated.

I think it's a pretty huge object with a lot of information that we don't care about, like every ICE pair and stats about that connection check. A lot of useful information is also not recorded in it, e.g. signaling path and time to connect.

> [nit] I feel like we have find-nominated-and-successful-candidate-pair code elsewhere. Wonder if this could be a helper we invoke in a few places.

The last place I'm aware of we did this was in the coturn dial options test suite, and this is the same pattern, but agreed it's probably better as a unified helper so we have a consistent definition. I will make that change.

proto/rpc/webrtc/v1/signaling.proto
```diff
@@ -211,4 +219,87 @@ message OptionalWebRTCConfigResponse {
...
+	// CONNECTION_SIGNALING_PATH_MDNS_LOCAL: signaled over an mDNS-discovered local-network path.
+	CONNECTION_SIGNALING_PATH_MDNS_LOCAL = 2;
+	// CONNECTION_SIGNALING_PATH_LOCAL: signaled through a loopback/private-address signaling server
+	// (e.g. a machine's own signaling server) without mDNS discovery.
+	CONNECTION_SIGNALING_PATH_LOCAL = 3;
+}
+
+// ReportConnectionMetadataRequest reports metadata about a WebRTC dial, per side: local is the
+// dialing SDK and remote is the robot.
+message ReportConnectionMetadataRequest {
```
danielbotros, line 278: Nothing would read or do anything with it yet but each connection has a UUID that we could add as a field here to correlate the connection between the different sides later.

rpc/server.go
cheukt: directionally the stats seem fine - is there a way to just make a separate connection to app (spin off a worker or something) after the connection has been made to upload the stats? we'll have to thread the connection report up, but right now the calls/flow is hard to reason about. we'll have to add plumbing in a few places, but i think that's fine and is more maintainable in the long run

took me a while to understand that for the local mdns case, we're doing client -> rdk -> cloud signaler to report stats

proto/rpc/webrtc/v1/signaling.proto
benjirewis: Is the caller or answerer even aware of that UUID? I thought that would've only been known by the signaling server and would not need to be communicated from caller -> signaling server or answerer -> signaling server.

rpc/wrtc_client.go
```diff
@@ -124,9 +125,22 @@ func dialWebRTC(
...
 	dialStart := time.Now()
 
+	// reachedStage tracks the furthest dial checkpoint reached, so a failed dial can report where it
+	// stopped. It is advanced from both this goroutine and the candidate-exchange / ICE callbacks, so
+	// it is an atomic; advance only ever moves it forward.
+	var reachedStage atomic.Int32
+	advance := func(s webrtcpb.DialStage) {
+		for {
+			cur := reachedStage.Load()
+			if int32(s) <= cur || reachedStage.CompareAndSwap(cur, int32(s)) {
```
benjirewis, line 138: This conditional in English is "if the state we've requested to advance to is not any more advanced than the current state OR (that previous predicate was false and) we have won the race to swap our more advanced state in"?

rpc/dial.go
```diff
@@ -191,6 +200,22 @@ func dial(
 				"host", originalAddress,
 			)
 
+			// When this attempt signals through the app (not an mDNS/local machine), record how to reach
```
benjirewis: Will sending up an mDNS/locally-signaled connection report not work until the client makes at least one connection through cloud signaling?

rpc/wrtc_client.go
```diff
@@ -227,17 +272,27 @@ func dialWebRTC(
...
-	peerConn, dataChannel, err := newPeerConnectionForClient(ctx, extendedConfig, dOpts.webrtcOpts.DisableTrickleICE, logger)
+	var dataChannel *webrtc.DataChannel
+	peerConn, dataChannel, err = newPeerConnectionForClient(ctx, extendedConfig, dOpts.webrtcOpts.DisableTrickleICE, logger)
 	if err != nil {
 		return nil, err
 	}
 
+	// Advance to DTLS_CONNECTED when the peer connection reaches Connected (ICE + DTLS complete), so a
+	// failure between ICE connectivity and data-channel-open can be attributed to DTLS vs the data channel.
+	peerConn.OnConnectionStateChange(func(state webrtc.PeerConnectionState) {
```
benjirewis, line 282: [q] We didn't have a callback for this event yet? I guess `OnICEConnectionStateChange` is a different thing?

What's the difference between `webrtc.PeerConnectionStateConnected` and `webrtc.ICEConnectionStateConnected`? My understanding was that the latter is when we output logs like "successful connection establishment."

rpc/wrtc_client_report.go
```diff
@@ -0,0 +1,211 @@
...
+// dialDurationMS returns milliseconds elapsed since start.
+func dialDurationMS(start time.Time) uint32 {
+	return uint32(time.Since(start).Milliseconds())
+}
+
+// viamSDKModule is the module path of the Viam RDK, which houses the Go SDK.
+const viamSDKModule = "go.viam.com/rdk"
+
+// sdkVersion returns the version of the Viam SDK module in the running binary.
+func sdkVersion() string {
```
benjirewis: Can you use https://github.com/viamrobotics/rdk/blob/4e9ec5f68f6599c3c2b8c9046aba34aec63a7f9c/robot/robot.go#L34-L46? I think we have logic that puts a `viam_client` value on the outgoing metadata of all gRPC requests already.

rpc/wrtc_client_report.go
```diff
@@ -0,0 +1,211 @@
...
+// caller wins; the mDNS/local attempts don't set it, so it ends up pointing at the app.
+func (c *dialReportCollector) setReportTarget(fn reportDialer) {
+	c.mu.Lock()
+	if c.reportTo == nil {
+		c.reportTo = fn
+	}
+	c.mu.Unlock()
+}
+
+// flush delivers a single connection report for a logical dial. dialErr is the logical dial's outcome:
```
benjirewis: > logical dial

I like this term and I at least abstractly understand this `flush` / `furthestReport` logic. The tests you've added give me some confidence that what you have works, but I'd be curious to see the distribution of your new metrics in staging once we merge to there, too.

rpc/wrtc_client_report_test.go
benjirewis, line 1: Thanks for adding all of these tests 🙏🏻 .

rpc/server.go
danielbotros: I changed the way we send local and mDNS reports by always opening a connection to prod app and sending it there. Any cloud registered robot dials with credentials app can authenticate, even for local dials, so we can still recieve those reports no problem.. The one quirk is that we may not always be able to attribute these local/mDNS dials to an org (since we do that according to `rpc-host` which is just a raw IP in the local case) but we decided that's acceptable for now.

rpc/wrtc_client.go
danielbotros: Exactly

rpc/dial.go
danielbotros: So the version you reviewed was one Claude pushed before I could get a chance to review lol, so that might've been the case. 

Now, we will always try to hit `ReportConnectionMetadata` on `app.viam.com`. We re-use the non-mDNS attempt's dial options, which are already for app if it is a cloud signaled attempt, or if it is a locally signaled attempt still uses credentials which are valid to authenticate to app with if it is a prod registered robot (so, we won't capture local dials to staging robots, for example). We re-point the address to app in this case.

rpc/wrtc_client_report.go
danielbotros: Yes, this should work. I changed the app side to read `viam_client` on the context and remove the sdk fields from the proto.

rpc/wrtc_client_report.go
danielbotros: Yup me too, excited to see what this reveals!

rpc/wrtc_client.go
danielbotros: Yeah they are different. `OnICEConnectionStateChange` fires on ICE state changes while `OnConnectionStateChange` fires on ICE + DTLS state change. `webrtc.PeerConnectionStateConnected` represents ICE connectivity being established and the DTLS handshake completed while `webrtc.ICEConnectionStateConnected` represents just that we found a usable candidate pair. We do log that msg on `ICEConnectionStateConnected` but technically we could log that and still fail to open the data channel due to DTLS reasons.

proto/rpc/webrtc/v1/signaling.proto
danielbotros: It looks like we actually already pass the UUID to the caller and answerer from the signaling server in `CallResponse` and `AnswerRequest`

rpc/dial.go
```diff
@@ -39,11 +39,20 @@ func dialInner(
...
 	}
 
+	// One logical dial can fan out into several WebRTC attempts (racing mDNS and cloud paths) or re-use a
+	// cached connection. A collector shared via the context gathers their reports and, on completion, opens
+	// its own connection to app (only app implements the report RPC) to deliver exactly one — the
+	// furthest-progressed attempt. Flushed in the background so reporting never adds latency to the dial's
+	// return.
+	collector := &dialReportCollector{ctx: ctx, host: address, logger: logger}
+	defer func() { go collector.flush(err) }()
+	ctx = contextWithReportCollector(ctx, collector)
```
cheukt: i'd prefer the private dial functions to return a report instead of adding a reporter to the context - feels like this is a misuse of the context. would that be a huge pain?

cheukt (commented):
just a question around how we surface the reports - otherwise things look good

rpc/dial.go
danielbotros: It wasn't too bad, just a bit of re-organizing. I agree it was context abuse, but thought to keep the function signatures simpler. I made the change to stop using the context to store the collector and have the dial functions return their own reports, aggregate them and in the defer of the top-level logical send we send the furthest progressed one. This made the collector struct unnecessary so I also just removed that.

cheukt (approved):
LGTM!

benjirewis (approved):
LGTM!

PR #584, opened 2026-07-09
RSDK-13986: Fix flaky stream reset test

EvanDorsky:
Fix for `TestWebRTCServerChannelResetStream/reset_stream_after_message_before_server_response` flaky test

The issue here was a race between a server handling a reset and handling a request.

The test was sending a request that kicked off an RPC, and then immediately sending reset request. The intent of the test was for the server to handle the reset request before the normal response could be sent (simulating a server that received a reset request while a long-running RPC was in flight). We should have expected to see two messages come back from the server: first the headers, then the trailers indicating that the stream was closed (reset request was handled). We shouldn't see the normal response in this case - the server would later try to send it and just get a closed pipe error which we throw out.

However the test was written around the other way that this situation could go, where the normal RPC response came back before the server actually handled the reset request. `expectedMessages` has the normal RPC response. And most of the time, this is the way the test actually went, since the RPC in the test was just getting held until the reset request was _sent_ and then immediately being allowed to respond. So the normal response usually won the race and we rarely saw the "correct" response with the trailers, and without the normal response.

The fix is to update the expected messages and then just wait for all the expected messages _before_ unblocking the server. This is basically just a swap of the two lines at the end of the test, but then also changing one of the lines to a `defer` so that we don't leak a goroutine if the test fails for some other reason.

I also added a bunch of comments in this test explaining what was actually going on, will hopefully help maintainers in the future.

### DIFF
--- a/rpc/wrtc_server_channel_test.go
+++ b/rpc/wrtc_server_channel_test.go
@@ -451,11 +451,6 @@ func TestWebRTCServerChannelResetStream(t *testing.T) {
 	})
 
 	t.Run("reset stream after message before server response", func(t *testing.T) {
-		respMd, err := proto.Marshal(&echopb.EchoResponse{
-			Message: "hello",
-		})
-		test.That(t, err, test.ShouldBeNil)
-
 		// We need to reset after the service has received the request but before
 		// its response has been written to the wire. To do this we inject a
 		// callback into the echoserver and use channels to coordinate the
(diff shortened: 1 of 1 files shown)


NickPPC (approved):

benjirewis (approved):
Looks great! It took me a while to parse through the test setup, what the subtest was doing and what your fix was. Thanks for adding more in-line comments. 

I asked Claude if any other subtests within `TestWebRTCServerChannelResetStream` suffer from similar races and it said "no." 🤷🏻 This test bug does just seem like a poorly placed `close(blockResp)` line that resulted in mostly incorrect timing; I'm assuming when we wrote the test originally we just adjusted the `expectedMessages` to account for the wrong order.

EvanDorsky:
Yeah, that was my exact guess too (re: adjusting the messages to account for the wrong order).

I went through the other sub-tests in order to understand this one so I also have a feeling that they're pretty solid.

PR #586, opened 2026-07-15
RSDK-14271 Expose `WaitForStateChange`

benjirewis:
RSDK-14271

## What

Adds `WaitForStateChange` to the `grpchelpers.ConnectivityState` interface, exposing gRPC's blocking state-transition notification API (`(*grpc.ClientConn).WaitForStateChange`) alongside the existing `GetState`. Also adds a forwarding implementation on `clientConnRPCAuthenticator`, mirroring its existing `GetState` forwarding, so that credentialed connections to app satisfy the updated interface.

## Why

Callers holding a connection to app can currently only poll GetState to observe connectivity. WaitForStateChange blocks until the connectivity state actually transitions, letting callers subscribe to lost/regained-connection events without polling. The immediate consumer is `viam-server`'s AppConn in `rdk` ([PR here](https://github.com/viamrobotics/rdk/pull/6234/changes)), which uses it to log Info-level "Lost connection to app" / "Regained connection to app" messages so offline periods are visible in robot logs.

[Description generated by Claude 🤖]

### DIFF
--- a/grpchelpers/grpchelpers.go
+++ b/grpchelpers/grpchelpers.go
@@ -2,6 +2,7 @@
 package grpchelpers
 
 import (
+	"context"
 	"errors"
 
 	"google.golang.org/grpc"
@@ -9,10 +10,13 @@ import (
 )
 
 // ConnectivityState allows callers to check the connectivity state of
-// the connection
+// the connection and to be notified of state transition events
 // see https://github.com/grpc/grpc-go/blob/master/clientconn.go#L648
 type ConnectivityState interface {
 	GetState() connectivity.State
+	// WaitForStateChange blocks until the connectivity state of the connection changes from
+	// sourceState or ctx expires, returning true in the former case and false in the latter.
+	WaitForStateChange(ctx context.Context, sourceState connectivity.State) bool
 }
 
 // ConnConnectivityState returns the connectivity.State of the current connection, if available.
--- a/grpchelpers/grpchelpers.go
+++ b/grpchelpers/grpchelpers.go
@@ -16,7 +16,7 @@ type ConnectivityState interface {
 	GetState() connectivity.State
 	// WaitForStateChange blocks until the connectivity state of the connection changes from
 	// sourceState or ctx expires, returning true in the former case and false in the latter.
-	WaitForStateChange(ctx context.Context, sourceState connectivity.State) bool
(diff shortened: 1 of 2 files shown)


cheukt (approved):

grpchelpers/grpchelpers.go
```diff
@@ -2,17 +2,21 @@
...
 	"errors"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
 )
 
 // ConnectivityState allows callers to check the connectivity state of
-// the connection
+// the connection and to be notified of state transition events
 // see https://github.com/grpc/grpc-go/blob/master/clientconn.go#L648
```
aldenh-viam: This link was a holdover from when I moved this interface from rdk to this repo (https://github.com/viamrobotics/rdk/pull/6127), but should be changed to a permalink: likely https://github.com/grpc/grpc-go/blob/v1.79.3/clientconn.go#L724

rpc/dialer.go
```diff
@@ -685,6 +685,21 @@ func (cc clientConnRPCAuthenticator) GetState() connectivity.State {
 	return checker.GetState()
 }
 
+// WaitForStateChange blocks until the connectivity state of the underlying connection changes from
+// sourceState or ctx expires, returning true in the former case and false in the latter. It is
+// exposed if it is available on the underlying type; otherwise it blocks until ctx expires and
```
aldenh-viam: [nit] I wonder if you could do better than waiting until the ctx expires if the type assertion fails?

aldenh-viam (approved):

rpc/dialer.go
danielbotros: Agreed, if someone was expecting this to work with the wrong type it'll silently not work

grpchelpers/grpchelpers.go
benjirewis: Thanks! Updated.

rpc/dialer.go
benjirewis: I changed `WaitForStateChange` to return an error in the event the type assertion fails. I agree that the old behavior could cause something unexpected. I did not muck with the analogous type assertion in `CheckForState`.

danielbotros (approved):
LGTM thanks!

PR #588, opened 2026-07-22
RSDK-14304 - Offline machines can't do local WebRTC

cheukt:
Cause: The internal signaling answerer dials the machine's own TLS listener over loopback. Its client TLS config has no RootCAs, so it verifies the Viam cert against system roots and fails ("unknown authority"). Under grpc.WithBlock the failing handshake retries until the 10s timeout → answerer never registers → local calls go unanswered.

Only reproduces offline (online uses external signaling) and with a cert present (with a cert the server binds TLS; without one it binds insecure loopback and works).

Fix: Set InsecureSkipVerify = true on the internal answerer's dial TLS config in rpc/server.go. Safe because it's a loopback self-dial — trust is guaranteed by loopback, not the cert. TLS itself must stay (shared listener serves external clients).

### DIFF
--- a/rpc/wrtc_signaling_answerer.go
+++ b/rpc/wrtc_signaling_answerer.go
@@ -228,8 +228,8 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 						}
 					}
 					ans.logger.Infow("failed to communicate with signaling server", "error", err)
-					utils.SelectContextOrWait(ctx, answererReconnectWait)
 				}
+				utils.SelectContextOrWait(ctx, answererReconnectWait)
 				continue
 			}
 			clearOfflineErrMap = true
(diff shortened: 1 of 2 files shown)


rpc/wrtc_signaling_answerer.go
```diff
@@ -228,8 +228,8 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 						}
 					}
 					ans.logger.Infow("failed to communicate with signaling server", "error", err)
-					utils.SelectContextOrWait(ctx, answererReconnectWait)
 				}
+				utils.SelectContextOrWait(ctx, answererReconnectWait)
```
cheukt, line 232: noticed that we were spamming reconnects, so added a way for all error paths

danielbotros (approved):
LGTM!

PR #590, opened 2026-07-28
rpc: deliver the dial connection report synchronously

danielbotros:
## Problem

v0.10.0 (#583) delivers each dial's connection report from a fire-and-forget goroutine:

```go
defer func() { go sendDialReport(ctx, address, logger, report, err) }()
```

`sendDialReport` detaches from the dial's context and opens its own connection to app. With no owner and nothing waiting on it, that goroutine outlives the dial, which surfaces problems in two ways:

1. Goroutine-leak checks fail — the report goroutine is still mid grpc-dial to app when a leak check snapshots goroutines.
2. Log in report goroutine after test has completed panic

Both are the same root cause: a background goroutine with no lifecycle owner.

## Fix

Deliver the report synchronously at the end of the dial so no goroutine outlives the dial. It still detaches from the dial's context (a cancelled/timed-out dial still reports), now bounded by `dialReportTimeout` (2s).

I chose this for simplicity over lifecycle tracking machinery for the report goroutine but a dial now waits on the report, which can add up to 2s of latency (but typically far less).

### DIFF
--- a/rpc/dial.go
+++ b/rpc/dial.go
@@ -47,8 +47,8 @@ func dialInner(
 
 	// One logical dial can fan out into several WebRTC attempts (racing mDNS and cloud paths) or re-use a
 	// cached connection. Each attempt reports its outcome which is threaded back up the call stack and, on
-	// completion, delivers over a separate connection to app (only app implements the report RPC) the
-	// furthest-progressed attempt. Flushed in the background so reporting never adds latency to the dial.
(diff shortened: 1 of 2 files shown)


danielbotros:
Closing in favor of #591, which takes a different approach: rather than making the dial wait on the report synchronously (adding up to 2s of dial latency in prod), #591 keeps reporting in the background and simply skips it in test binaries via `testing.Testing()`. Fixes the leak/panic with no production latency cost.

PR #593, opened 2026-07-31
Retry backing MongoDB test client connection

ale7714:
## What

`backingMongoDBClientWithOptions` in `testutils/mongodb.go` gave the initial connection one attempt with a 5s budget shared across Connect, Ping, and `replSetGetStatus`, and cached any failure for the life of the test process. This change gives it 3 attempts of 10s each, 1s apart, and caches the error only after all attempts fail — so a hard outage still fails fast for the rest of the process.

## Why

On a loaded CI runner, one missed 5s ping poisons every test in the package. In viamrobotics/app, [a single transient blip](https://github.com/viamrobotics/app/actions/runs/30258491158/job/91035279434) failed all 56 tests in the `auth` package: the first test burned the 5s budget, the remaining 55 failed instantly on the cached error. The 56-failure count then exceeded gotestsum's `--rerun-fails-max-failures` cap, so its retry never ran. This has hit app CI three times; each rerun passed with no code change. (viamrobotics/app#13040 raises the gotestsum cap as the in-repo guard; this PR removes the amplification at the source.)

## Notes

- Only a successful client is cached during the retry window; the failure-caching behavior for a genuinely down MongoDB is unchanged, it just happens after 3 attempts (~32s) instead of one (5s).
- No API change; `BackingMongoDBClient` / `BackingMongoDBClientWithOptions` behave the same on success.

### DIFF
--- a/testutils/mongodb.go
+++ b/testutils/mongodb.go
@@ -99,6 +99,12 @@ func backingMongoDBClient() (*mongo.Client, error) {
 	return backingMongoDBClientWithOptions(options.Client())
 }
 
+const (
+	backingMongoDBConnectAttempts       = 3
+	backingMongoDBConnectAttemptTimeout = 10 * time.Second
+	backingMongoDBConnectRetryBackoff   = time.Second
+)
+
 func backingMongoDBClientWithOptions(baseOptions *options.ClientOptions) (*mongo.Client, error) {
 	cacheMu.Lock()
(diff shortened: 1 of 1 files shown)


ale7714:
`test go` has been flaky in `app` and getting your pr kicked out of the queue because it failed to talk to mongo is sad

abe-winter (approved):

PR #595, opened 2026-08-12
Add OpenCensus collector exporter

jmatth:
This adds the `ocagent` go library, which provides support for exporting trace data to an OpenCensus agent. This will be used to send trace data to an OpenTelemetry collector configured to receive OpenCensus agent messages

### DIFF
--- a/perf/ocagent_exporter.go
+++ b/perf/ocagent_exporter.go
@@ -1,0 +1,108 @@
+package perf
+
+import (
+	"errors"
+	"os"
+	"time"
+
+	"contrib.go.opencensus.io/exporter/ocagent"
+	"github.com/caarlos0/env/v11"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/trace"
+
+	"go.viam.com/utils"
+)
+
+// OCAgentExporter exports traces and metrics over gRPC to an OpenCensus
+// agent/collector. The OpenTelemetry Collector accepts the same protocol via
+// its "opencensus" receiver.
+type OCAgentExporter struct {
+	e       *ocagent.Exporter
+	sampler trace.Sampler
+}
+
+var _ Exporter = &OCAgentExporter{}
+
+// OCAgentOptions is used to configure [OCAgentExporter].
+type OCAgentOptions struct {
+	Logger utils.ZapCompatibleLogger
+	// Address is the "host:port" of the collector. If empty, OC_AGENT_ADDRESS
+	// is used instead.
+	Address string
+}
+
+// NewOCAgentExporter creates a new OpenCensus agent/collector [Exporter],
+// configured from the environment:
+//
+//	OC_AGENT_ADDRESS             collector "host:port" (used when OCAgentOptions.Address is empty)
+//	OC_AGENT_INSECURE            dial without TLS (default true)
+//	OC_AGENT_RECONNECTION_PERIOD retry interval for reconnecting to the collector
+//	OC_SAMPLING_BY_NAME_PER_SEC  same sampling behavior as the cloud exporter
+//	OC_SAMPLING_PROB             same sampling behavior as the cloud exporter
+//	SERVICE_NAME                 service name reported with spans
+func NewOCAgentExporter(opts OCAgentOptions) (*OCAgentExporter, error) {
+	envOpts, err := env.ParseAs[struct {
+		Address              string        `env:"OC_AGENT_ADDRESS"`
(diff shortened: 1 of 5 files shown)


perf/ocagent_exporter.go
```diff
@@ -0,0 +1,108 @@
...
+
+// OCAgentExporter exports traces and metrics over gRPC to an OpenCensus
+// agent/collector. The OpenTelemetry Collector accepts the same protocol via
+// its "opencensus" receiver.
+type OCAgentExporter struct {
+	e       *ocagent.Exporter
+	sampler trace.Sampler
+}
+
+var _ Exporter = &OCAgentExporter{}
```
abe-winter, line 23: is this a global or a type check?

perf/ocagent_exporter.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return &OCAgentExporter{
+		e:       exp,
+		sampler: samplerFromEnvOpts(envOpts.SamplingByNamePerSec, envOpts.SamplingProbability, opts.Logger),
+	}, nil
+}
+
+// Start implements [Exporter]. Registers views and starts trace/metric
+// exporting to the collector.
+func (o *OCAgentExporter) Start() error {
```
abe-winter, line 93: what do the 'implements [Exporter]` comments mean in this file?

perf/ocagent_exporter.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	ocOpts := []ocagent.ExporterOption{ocagent.WithAddress(address)}
+	if envOpts.Insecure {
+		ocOpts = append(ocOpts, ocagent.WithInsecure())
+	}
+	if envOpts.ReconnectionPeriod > 0 {
+		ocOpts = append(ocOpts, ocagent.WithReconnectionPeriod(envOpts.ReconnectionPeriod))
+	}
+	if serviceName := os.Getenv("SERVICE_NAME"); serviceName != "" {
```
abe-winter: why read this directly rather than in the envOpts block above?

perf/ocagent_exporter.go
```diff
@@ -0,0 +1,108 @@
...
+// Start implements [Exporter]. Registers views and starts trace/metric
+// exporting to the collector.
+func (o *OCAgentExporter) Start() error {
+	if err := registerApplicationViews(); err != nil {
+		return err
+	}
+	if err := o.e.Start(); err != nil {
+		return err
+	}
+	view.RegisterExporter(o.e)
```
abe-winter, line 100: does registration / deregistration order matter here?

perf/ocagent_exporter.go
```diff
@@ -0,0 +1,108 @@
...
+	trace.ApplyConfig(trace.Config{DefaultSampler: o.sampler})
+	return nil
+}
+
+// Stop implements [Exporter]. Flushes any pending spans/metrics and closes the
+// connection to the collector.
+func (o *OCAgentExporter) Stop() {
+	trace.UnregisterExporter(o.e)
+	view.UnregisterExporter(o.e)
+	o.e.Flush()
```
abe-winter, line 111: does this time out? should it?

abe-winter (approved):
left notes but seems good

perf/ocagent_exporter.go
jmatth: Compile time type check

perf/ocagent_exporter.go
jmatth: I think these specific ones are claude generated but my IDE fills in the same comment when generating method stubs. Doc comments have to start with the method name to keep the linter happy and the `[Exporter]` syntax renders as a link to that symbol in godoc and is a valid goto target for stuff like command click, at least in my IDE

perf/ocagent_exporter.go
jmatth: As far as I know no, the trace and stats exporters don't depend on each other

perf/ocagent_exporter.go
jmatth: Not in a way that we can easily control. Looks like it fires off some requests in the background and then waits on a waitgroup. I'm guessing those requests have their own timeouts that were maybe configured when the exporter was set up. I don't think it's worth wrapping this in our own goroutine + timeout logic here, if the app is shutting down then eventually it should hit its own timeout and finish exiting even if this function hasn't returned

perf/ocagent_exporter.go
jmatth: Probably claude generated and I didn't catch that. I'll move it into the struct

PR #596, opened 2026-08-13
APP-17548 - rpc: identify the client SDK from one shared parser

cheukt:
## Problem

`SendOfferInit` guessed the client SDK with an `if / else if` chain over `viam_client`, `x-grpc-web`, and `user-agent`. The comment above it stated the premise: *"The first only exists for Golang, the second only exists for TypeScrpt"*.

That stopped being true in Jan 2026, when the TypeScript SDK began sending `viam_client` ([viam-typescript-sdk#710](https://github.com/viamrobotics/viam-typescript-sdk/pull/710), RSDK-12734). TS now sends **both** headers, so it enters the first branch, fails `strings.Contains(v, "go;")`, and falls out with `sdkType` still at its initialized `"unknown"` — the `x-grpc-web` fallback meant to catch it is unreachable.

Every TypeScript client has recorded as `unknown` since. That value feeds `signaling/connection_establishment_attempts`, `connection_establishment_expected_failures`, `connection_establishment_failures`, `call_exchange_duration`, the `sdk_type` span attribute, and the `sdk_type` field persisted on `rpc.calls` docs (which is re-read as a metric label later in the exchange).

## Change

New `rpc/sdk_info.go`, resolving the `TODO(RSDK-11864)` that sat on the old code:

- `SDKInfo{Type, Source, Version, Raw}`, `SDKInfoFromCtx`, and `Label()`.
- **Sequential fallbacks** instead of exclusive branches, so an unrecognized `viam_client` no longer shadows `x-grpc-web` / `user-agent`.
- **Allowlists** for both the SDK type and the embedder source. Both are client-supplied and reach metric labels, so anything unrecognized is dropped rather than labeled.
- A source qualifier (`typescript(viam-app)`) is only read when the SDK type itself was recognized — otherwise a fallback could stamp its inferred type onto someone else's source, turning `rust(viam-app)` over gRPC-Web into `typescript(viam-app)`.
- `Label()` returns `""` when unknown so each caller keeps its own sentinel (`unknown` here, `unspecified` in app).
- `Raw` retains the reported value, truncated to 128B, **only** when it was not fully understood — an unrecognized SDK or an unlisted source. It is logged and attached to the span, never used as a label.

This is how a new SDK or embedder gets noticed instead of being silently absorbed by a fallback.

## Notes

- The `user-agent: tonic/` branch is **not** legacy cruft — it is the only signal for Python and C++, whose WebRTC signaling is performed by viam-rust-utils rather than by the SDK. `python/c++` stays one value because the two are not separable at that vantage point.
- `Raw` is logged rather than persisted on the call doc: the collection's TTL is short enough that storing it buys little over the log line.
- app consumes this next (APP-17548, [viamrobotics/app#13176](https://github.com/viamrobotics/app/pull/13176)), deleting its own copy of the same parsing.
- rdk holds a third copy (`robot/client/client.go`) plus a stale `"maybe-typescript"` fallback in `robot/web/request_counter.go`, wrong for the same reason. Out of scope here.

## Testing

`rpc/sdk_info_test.go` — 16 cases, including the regression (`viam_client` + `x-grpc-web` together), the source-leak case above, unrecognized type and source both retaining `Raw`, and truncation. `golangci-lint run ./rpc/...` reports 0 issues.

### DIFF
--- a/rpc/sdk_info.go
+++ b/rpc/sdk_info.go
@@ -84,12 +84,16 @@ func SDKInfoFromCtx(ctx context.Context) SDKInfo {
 		}
 	}
 
+	// A version reported alongside an unrecognized SDK belongs to that SDK, not to the one the
+	// fallback infers, so it is dropped. Raw still carries it.
 	if vals := md.Get(XGRPCWebMetadataField); len(vals) > 0 && vals[0] == "1" {
 		info.Type = SDKTypeTypeScript
+		info.Version = ""
 		return info
 	}
 	if vals := md.Get(UserAgentMetadataField); len(vals) > 0 && strings.Contains(vals[0], "tonic/") {
 		info.Type = SDKTypePythonOrCPP
+		info.Version = ""
 	}
 	return info
 }
--- a/rpc/sdk_info.go
+++ b/rpc/sdk_info.go
@@ -1,0 +1,125 @@
+package rpc
+
+import (
+	"context"
+	"strings"
+
+	"google.golang.org/grpc/metadata"
+)
+
+// SDK types reported by Viam SDKs in the viam_client metadata field.
+const (
+	SDKTypeGo         = "go"
+	SDKTypeTypeScript = "typescript"
+	SDKTypePython     = "python"
+
+	// SDKTypePythonOrCPP is reported when a tonic user agent is the only signal available.
+	// Python and C++ both signal through viam-rust-utils, so at that point they cannot be
+	// told apart.
+	SDKTypePythonOrCPP = "python/c++"
+)
+
+// maxSDKRawLen bounds the client-supplied viam_client value before it is logged or stored.
+const maxSDKRawLen = 128
+
+// Recognized values. Anything else is dropped rather than labeled, so that a client cannot
+// inflate the cardinality of metrics derived from SDKInfo.Label.
+var (
+	knownSDKTypes = map[string]bool{
+		SDKTypeGo:         true,
+		SDKTypeTypeScript: true,
+		SDKTypePython:     true,
+	}
+	knownSDKSources = map[string]bool{
+		"viam-app": true,
+	}
+)
+
+// SDKInfo identifies the client SDK behind a request.
+type SDKInfo struct {
+	// Type is the bounded SDK identifier, empty when the SDK could not be determined.
+	Type string
+
+	// Source is the bounded embedder wrapping the SDK, e.g. "viam-app". Empty when the client
+	// reported none or reported one that is not recognized.
+	Source string
+
+	Version string
+
+	// Raw is the reported viam_client value, truncated to maxSDKRawLen and set only when it was
+	// not fully understood — an unrecognized SDK or an unlisted source. It is how a new SDK or
+	// embedder gets noticed: safe to log or store, never to use as a metric label.
+	Raw string
+}
+
+// Label renders the SDK as a bounded metric label, qualified by source when one was
+// recognized, e.g. "typescript(viam-app)". It returns "" when the SDK is unknown so that
+// callers can apply their own sentinel.
+func (i SDKInfo) Label() string {
+	if i.Type == "" {
+		return ""
+	}
+	if i.Source == "" {
+		return i.Type
+	}
+	return i.Type + "(" + i.Source + ")"
+}
+
+// SDKInfoFromCtx determines which SDK issued the request described by ctx.
+//
+// A recognized viam_client value wins. The x-grpc-web and user-agent fallbacks cover clients
+// that do not send one: SDK releases predating viam_client, and the Python and C++ SDKs, whose
+// signaling is performed by viam-rust-utils rather than by the SDK itself.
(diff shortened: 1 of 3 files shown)


rpc/sdk_info_test.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+		{
+			// The fallback classifies the client, but the unrecognized value is still surfaced:
+			// a new TypeScript-based SDK would otherwise be labeled and never noticed.
+			name: "falls back to x-grpc-web when viam_client is unrecognized",
+			md: map[string]string{
+				ViamClientMetadataField: "rust;v1.2.3;v0.1.0",
+				XGRPCWebMetadataField:   "1",
+			},
+			expected: SDKInfo{Type: SDKTypeTypeScript, Version: "v1.2.3", Raw: "rust;v1.2.3;v0.1.0"},
```
danielbotros: nit: Maybe in this case we should clear the `Version` field when hit either the grpc-web or tonic fallback and `viam_client` is unrecognized since `v1.2.3` is probably not actually the TS SDK version so we probably don't wanna record that for metrics.

danielbotros (approved):
LGTM!

rpc/sdk_info.go
```diff
@@ -0,0 +1,129 @@
...
+	// Raw is the reported viam_client value, truncated to maxSDKRawLen and set only when it was
+	// not fully understood — an unrecognized SDK or an unlisted source. It is how a new SDK or
+	// embedder gets noticed: safe to log or store, never to use as a metric label.
+	Raw string
+}
+
+// Label renders the SDK as a bounded metric label, qualified by source when one was
+// recognized, e.g. "typescript(viam-app)". It returns "" when the SDK is unknown so that
+// callers can apply their own sentinel.
+func (i SDKInfo) Label() string {
```
benjirewis, line 58: [nit] `String()` would let you `%s` it with formatting directives and is maybe more idiomatic.

rpc/sdk_info.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"context"
+	"strings"
+
+	"google.golang.org/grpc/metadata"
+)
+
+// SDK types reported by Viam SDKs in the viam_client metadata field.
+const (
+	SDKTypeGo         = "go"
```
benjirewis, line 12: [nit] Could use an `sdkType` type here instead of just `string`.

rpc/sdk_info_test.go
cheukt: done

rpc/sdk_info.go
```diff
@@ -0,0 +1,129 @@
...
+	SDKTypeTypeScript = "typescript"
+	SDKTypePython     = "python"
+
+	// SDKTypePythonOrCPP is reported when a tonic user agent is the only signal available.
+	// Python and C++ both signal through viam-rust-utils, so at that point they cannot be
+	// told apart.
+	SDKTypePythonOrCPP = "python/c++"
+)
+
+// maxSDKRawLen bounds the client-supplied viam_client value before it is logged or stored.
```
benjirewis, line 22: Are we trying to guard against users shoving super long `viam_client` values in their metadata for some reason? Trying to understand if this is defensive or this actually happened somewhere.

rpc/sdk_info.go
```diff
@@ -0,0 +1,129 @@
...
+	var info SDKInfo
+	sdkField, rest, _ := strings.Cut(raw, ";")
+	info.Version, _, _ = strings.Cut(rest, ";")
+
+	base, source, hasSource := strings.Cut(strings.ToLower(strings.TrimSpace(sdkField)), "(")
+	if knownSDKTypes[base] {
+		info.Type = base
+
+		// Only meaningful against a recognized SDK. Keeping it otherwise would let a fallback
+		// stamp its own Type onto someone else's source, e.g. "rust(viam-app)" over gRPC-Web
```
benjirewis: > onto someone else's source

I understand the code here, but this comment is confusing to me. What is "someone else's source"? Who else would be reporting sources besides Viam SDKs and apps?

rpc/sdk_info.go
cheukt: String() is probably not the right method here, since Label() will not include version

rpc/sdk_info.go
cheukt: we could, but we do comparisons in a few places - would rather just keep these types as a string to make that easier

rpc/sdk_info.go
cheukt: ya, this is defensive. users can overwrite their metadata so some sanitization such that we won't get overwhelmed randomly

rpc/sdk_info.go
cheukt: updated the comment, hopefully its clearer now. but any users, maybe even malicious ones could report sources - relates to other comments here where we try to be defensive around user defined inputs.

benjirewis (approved):
All makes sense; thanks!

PR #598, opened 2026-08-18
RSDK-14416 Forward caller credentials from signaling server to answerer

benjirewis:
RSDK-14416

## What

Forwards callers' entity and auth metadata from the signaling server to the answerer for authorization (used [here](https://github.com/viamrobotics/rdk/pull/6269)). 

Also puts the auth metadata (JWT claims) on `EntityInfo` so `viam-server` can read metadata like "app_user_id" without reparsing the token.

## Why

`viam-server` had no way to tell which user was responsible for which gRPC request for connections made over WebRTC (direct gRPC connections do have the JWT in every request). We need to forward the caller's identity down to the answerer.

Also, the existing auth interceptor injected only the entity into the context and did not pass along metadata like "app_user_id" despite having already parsed it (in `claims.Metadata()`).


rpc/wrtc_server_channel.go
```diff
@@ -142,3 +160,17 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
...
+// claimsFromAuthToken parses the (already signaler-verified) bearer token into its
+// claims. Parsing is unverified because we trust the signaler that forwarded the token;
+// it returns nil if the token is empty or unparseable.
+func claimsFromAuthToken(token string) *JWTClaims {
+	if token == "" {
+		return nil
+	}
+	var claims JWTClaims
+	if _, _, err := jwt.NewParser().ParseUnverified(token, &claims); err != nil {
+		return nil
```
jmatth: I wouldn't mind returning `nil, err` here and logging a warning at the callsite. Probably good to know if we break something and start sending unparsable garbage

rpc/wrtc_server_channel.go
```diff
@@ -126,8 +133,19 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
...
+		// coarse audience approximation when no token was forwarded (e.g. an unauthenticated
+		// caller). We trust the signaler's authentication, so we do not re-verify the token.
+		entityInfo := EntityInfo{Entity: ch.authAudience}
+		if claims := claimsFromAuthToken(ch.callerAuthToken); claims != nil {
+			if entity := claims.Entity(); entity != "" {
+				entityInfo.Entity = entity
+			}
+			entityInfo.AuthMetadata = claims.Metadata()
+		}
+		handlerCtx = ContextWithAuthEntity(handlerCtx, entityInfo)
```
danielbotros: nit: It looks like `authAudience` and `callerAuthToken` don't change for a channel and nothing else reads it besides this block. Maybe we can create `entityInfo` in the constructor and just read it here instead of doing it per message.

rpc/wrtc_server_channel.go
```diff
@@ -126,8 +133,19 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
...
-		handlerCtx = ContextWithAuthEntity(handlerCtx, EntityInfo{Entity: ch.authAudience})
+		// Reconstruct the auth entity the auth interceptor would have set (that interceptor
+		// is not in the WebRTC chain). The signaler-forwarded token gives us both the
+		// caller's identity (JWT subject) and its auth metadata claim; fall back to the
+		// coarse audience approximation when no token was forwarded (e.g. an unauthenticated
+		// caller). We trust the signaler's authentication, so we do not re-verify the token.
+		entityInfo := EntityInfo{Entity: ch.authAudience}
+		if claims := claimsFromAuthToken(ch.callerAuthToken); claims != nil {
+			if entity := claims.Entity(); entity != "" {
+				entityInfo.Entity = entity
```
danielbotros: IIRC `entityInfo.Entity` is used in some way to manage session ownership in the rdk, do these changes have any impact on that?

danielbotros (approved):
Overall LGTM mainly just a question about downstream effects on sessions

jmatth (approved):

rpc/wrtc_server_channel.go
benjirewis: Good idea; done.

rpc/wrtc_server_channel.go
benjirewis: > IIRC entityInfo.Entity is used in some way to manage session ownership in the rdk

Yep.

> do these changes have any impact on that?

Also yes. 

The entity used to always be `ch.authAudience` and now it will _almost_ always be `ch.callerAuthToken`. `ch.authAudience` is basically just the name of the answerer (`"c-3po-main.ydbecmp2zs.viam.cloud"` for example) and is the same for every caller. `ch.callerAuthToken` is the JWT token of the connecting client and is obviously different for every caller.

We read the `entityInfo.Entity` value [here](https://github.com/viamrobotics/rdk/blob/b0d2f62e935afd9270047057be1fe60c010a1a93/robot/session_manager.go#L180). Particularly the `!sess.CheckOwnerID()` bit. We use the "owner ID" (read: `entityInfo.Entity`) to do a basic check of "does this session belong to the person that's connecting?" But, that value was identical for every connecting client. Now, that value will in fact be different for every  connecting client, so it is possible for us to reject a client trying to use a _different_ client's session because they know the session ID value or somehow collided.

Effectively, it's an improvement to our session identification logic that _should_ only benefit users. But, yes, it affects session ownership/management in `rdk` and could result in `SESSION_EXPIRED` errors that weren't occurring before if someone is relying on using one session across multiple clients.

rpc/wrtc_server_channel.go
benjirewis: Yep; good idea.

proto/rpc/webrtc/v1/signaling.proto
```diff
@@ -131,6 +131,11 @@ message AnswerRequestInitStage {
 	string sdp = 1;
 	WebRTCConfig optional_config = 2;
 	optional google.protobuf.Timestamp deadline = 3;
+	// caller_auth_token is the signed bearer token the caller authenticated to the
+	// signaling server with, forwarded from the signaler so the answerer can identify the
+	// caller for authorization. Empty when the caller was unauthenticated (can only happen
+	// with local signaling).
+	optional string caller_auth_token = 4;
```
benjirewis: From an offline meeting: we can't forward this entire token because it would let a bad answerer mimic the client and gain all of its privileges. We should only forward the pieces of metadata important for authorization (`app_user_id`, API key ID, email).

rpc/wrtc_server_channel.go
danielbotros: Got it, thanks for the explanation and digging into this! I think this makes more sense than what we had before. 

> reject a client trying to use a different client's session because they know the session ID value or somehow collided

It doesn't sound like a user can easily run into this so this all SGTM!

proto/rpc/webrtc/v1/signaling.proto
benjirewis: Changed to only forward
```go
// caller_auth_entity is the caller's authenticated entity (its JWT subject), extracted
// from the caller's auth token by the signaling server. Only the caller's identity is
// forwarded to the answerer, not its bearer token. Empty when the caller was
// unauthenticated.
string caller_auth_entity = 4;
// caller_auth_metadata is the caller's auth metadata (its rpc_auth_md claim), extracted
// from the caller's auth token by the signaling server.
map<string, string> caller_auth_metadata = 5;
```
instead of `caller_auth_token`.

jmatth (approved):

PR #599, opened 2026-08-21
Replace stale operator document on call queue startup instead of failing

cecilyViam:
## What changed

`NewMongoDBWebRTCCallQueue` now registers its operator document with a guarded upsert (`FindOneAndReplace` + upsert) instead of `InsertOne`:

- No prior doc → same behavior as before.
- Prior **expired** doc (TTL sweep hasn't fired yet) → replaced silently.
- Prior **unexpired** doc → replaced with a warning log, since that means a same-identity predecessor died within the heartbeat window — or operator IDs are not unique as required, which the warning acts as a tripwire for.

Replacing also clears the predecessor's stale host counts immediately instead of letting them pollute queue-size maxima and host-online checks until the TTL sweep (which runs only ~every 60s).

## Why

A process that dies without cleanup (`Close` does not delete the doc) leaves an unexpired operator document behind. A successor with the same operator ID — e.g. a restarted container — previously crashed on startup with `E11000 duplicate key error collection: rpc.operators` until the TTL sweep removed the doc. This is the crashloop seen on nearly every app GKE deploy in staging usw4 (amplified by pods sharing the node's instance ID — fixed separately in viamrobotics/app#13274).

Jira: [APP-17636](https://viam.atlassian.net/browse/APP-17636)

## Testing

- New `TestMongoDBWebRTCCallQueueOperatorReRegistration` reproduces the exact E11000 against the old code and passes with this change.
- Full `TestMongoDBWebRTCCallQueue*` suite passes against a single-node replica-set Mongo; `make lint-go` clean.

[APP-17636]: https://viam.atlassian.net/browse/APP-17636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rpc/wrtc_call_queue_mongodb.go
+++ b/rpc/wrtc_call_queue_mongodb.go
@@ -308,17 +308,44 @@ func NewMongoDBWebRTCCallQueue(
 		return nil, err
 	}
 
-	result, err := operatorsColl.InsertOne(ctx, bson.D{
-		{webrtcOperatorIDField, operatorID},
-		{webrtcOperatorExpireAtField, time.Now().Add(operatorHeartbeatWindow)},
-	})
-	if err != nil {
-		return nil, err
-	}
-	logger.Infow(
-		"successfully added operator document to operators collection",
-		"operator_id", operatorID, "document_id", result.InsertedID,
+	// A document with our ID may remain from a previous process with the same identity
+	// (e.g. a restarted container) whose TTL has not yet fired; replace it rather than
+	// failing startup. Replacing also clears the predecessor's stale host counts.
+	priorOperator := operatorsColl.FindOneAndReplace(
+		ctx,
+		bson.D{{webrtcOperatorIDField, operatorID}},
+		bson.D{
+			{webrtcOperatorIDField, operatorID},
+			{webrtcOperatorExpireAtField, time.Now().Add(operatorHeartbeatWindow)},
+		},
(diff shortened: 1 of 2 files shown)


abe-winter:
removing myself as this is a netcode question. add me back if needed

jmatth (commented):
I'm guessing Cloud Run and GKE are pretty consistent about assigning unique ids to their instances so this is probably fine, but the failure mode where signaling is just broken with a warning log until we intervene is scary. As an alternative could we do one of:

- Have [the app code][1] that figures out the operator id append a uuid ever time, even if it has an instance id or hostname? Then we can be guaranteed that each operator is unique while still having a way to map operator documents back to instances
- Enter a retry loop and only clobber the document after the expiry time has passed? The TTL is only 15 seconds
- Drain in flight signalling connections and delete the document as part of a clean shutdown? (This one would probably make sense to send over to netcode to deal with)

[1]: https://github.com/viamrobotics/app/blob/3e3e2a776acaa88099dfdb398ebc1b2b962508e3/server/config/config.go#L414-L438

cecilyViam:
I think I've decided we don't need this part of the fix. https://github.com/viamrobotics/app/pull/13274 scopes the instanceID to pods, and the application already successfully restarts into a viable launch with the current design (where duplicates were being caused by new pods deployed to the same node). This introduces unnecessary risk to optimize only the scenario where a pod restarts in-place quicker than the expiry ttl.

PR #600, opened 2026-08-24
pexec: treat EPERM from a process-group signal as already gone

NickPPC:
## Summary

`kill` sweeps the process group a few seconds after signaling the process itself, to catch orphaned children. When the process has already exited there is nothing to catch — but that reports two different ways. Reaped, the group is gone and `kill(2)` answers `ESRCH`, which is already handled. Exited-but-unreaped, the group still holds a zombie: Linux signals it without complaint, Darwin answers `EPERM`.

Reading that `EPERM` as a permission failure turned a clean shutdown into a failure to stop, on macOS alone. It also bypassed `ProcessNotExistsError`: rdk's `modManager` ignores that error when stopping a module, and instead saw a hard failure whenever a module process exited promptly.

Accept `EPERM` alongside `ESRCH`, folding the classification duplicated at both group-signal sites into `killGroup`/`processGroupGone`. Drive-by: the SIGKILL site named `p.stopSig` in its error rather than the signal it actually sent.

Related: viamrobotics/rdk#6379 makes rdk's own tests robust to this. The two are complementary and neither depends on the other — see that PR for how they fit together.

## Testing

- New table test on the classifier; the full `./pexec/` suite passes on darwin/arm64.
- Darwin behavior confirmed with a probe (own process group: zombie → `EPERM`, reaped → `ESRCH`). The timing window that triggers it is load-dependent, so it is not reproduced in a test.

<details>
<summary>Claude Code prompts used</summary>

- "You said that "the real fix belongs upstream in goutils", Can you create a worktree and create the fix in goutils directly? goutils is available at @../../../../goutils"
- "open a draft PR for goutils
Since the change in rdk is complementary you probably need to update the description, it states that the real fix is in goutils which turns out to not be true. You should link the goutils PR in the rdk PR description and explain how they are related"

</details>

### DIFF
--- a/pexec/managed_process_unix.go
+++ b/pexec/managed_process_unix.go
@@ -110,15 +110,8 @@ func (p *managedProcess) kill() (bool, error) {
 	select {
 	case <-timer.C:
 		p.logger.Infof("stopping entire process group %d with signal %s", p.cmd.Process.Pid, p.stopSig)
-		if err := syscall.Kill(-p.cmd.Process.Pid, p.stopSig); err != nil {
-			var errno syscall.Errno
-			if errors.As(err, &errno) && errno == syscall.ESRCH {
-				return false, &ProcessNotExistsError{err}
-			} else if errors.Is(err, os.ErrProcessDone) {
-				return false, &ProcessNotExistsError{err}
-			}
-			return false, errors.Wrapf(err, "error signaling process group %d with signal %s",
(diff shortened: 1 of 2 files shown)


jmatth (approved):
LGTM. I manually tested and confirmed macOS and Linux have different behavior when signaling a process group containing only zombie processes just as described.

PR #601, opened 2026-08-28
jwks: retry the initial OIDC discovery and JWKS fetch

ale7714:
NewCachingOIDCJWKKeyProvider makes two network calls (the discovery GET and the initial JWKS refresh) with no retry, so a single reset TCP connection is fatal to every caller at process startup; viamrobotics/app nightly E2E run 33142747007 lost a shard exactly this way, since all four app server binaries call this at init. Transient failures retry with exponential backoff (500ms doubling to 5s, 30s cap), bounded by the caller's context; oidc.ErrIssuerInvalid returns immediately since retrying a misconfigured issuer cannot help. The loop is hand-rolled to avoid adding a backoff dependency to the library. ServeFakeOIDCEndpointWithFailures injects 500s into the existing fake OIDC server so the new test exercises one failed round trip before success.

Replaces viamrobotics/app#13353, which put the retry app-side.

### DIFF
--- a/jwks/jwks.go
+++ b/jwks/jwks.go
@@ -76,10 +76,43 @@ func (cp *cachingKeyProvider) Fetch(ctx context.Context) (KeySet, error) {
 // ensure interface is met.
 var _ KeyProvider = &cachingKeyProvider{}
 
+// Backoff schedule for the initial discovery and JWKS fetches; a transient network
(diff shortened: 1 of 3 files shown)


ale7714:
Addressing https://github.com/viamrobotics/app/actions/runs/33142747007

PR #603, opened 2026-09-02
RSDK-14415 Pull metadata off auth.State / internalForeverJWT

benjirewis:
RSDK-14415

## What

If `entity.AuthMetadata` is empty (`rpc_auth_md` was missing in the JWT token headers), try to use `CreateMetadata` instead. See [PR for app-side implementation](https://github.com/viamrobotics/app/pull/13406) of `CreateMetadata`.

## Why

App currently uses the `internalForeverJWT` for all machine page connections in the `Authorization` header. There is no metadata on that token, so the `app_user_id` and `email` are unclear. We can pull those pieces of information from `State.AppUser` and `State.User` where `State` is `entity.Data` in the signaling server code. That should let us insert metadata with `internalForeverJWT` connections for now.

A future app-side change may stop using `internalForeverJWT` entirely and actually invoke `Call` with proper a `Authorization` header. This change is a stopgap until that point.

### DIFF
--- a/rpc/wrtc_signaling_server.go
+++ b/rpc/wrtc_signaling_server.go
@@ -203,7 +203,7 @@ func (srv *WebRTCSignalingServer) Call(req *webrtcpb.CallRequest, server webrtcp
 		callerAuthEntity, callerAuthMetadata = entity.Entity, entity.AuthMetadata
 
 		// Auth metadata (app user ID, email) should be in the JWT token, but the
-		// internalForeverJWT used by the machine page has no metadata. Hackily pull it off
+		// internalForeverJWT used by the machine page has no metadata. Pull it off
 		// `entity.Data`.
 		if callerAuthMetadata == nil {
 			if provider, ok := entity.Data.(interface{ CreateMetadata() map[string]string }); ok {
--- a/rpc/wrtc_signaling_server.go
+++ b/rpc/wrtc_signaling_server.go
@@ -201,15 +201,6 @@ func (srv *WebRTCSignalingServer) Call(req *webrtcpb.CallRequest, server webrtcp
 	var callerAuthMetadata map[string]string
 	if entity, ok := ContextAuthEntity(ctx); ok {
 		callerAuthEntity, callerAuthMetadata = entity.Entity, entity.AuthMetadata
-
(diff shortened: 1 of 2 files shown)


rpc/wrtc_signaling_server.go
```diff
@@ -201,6 +201,15 @@ func (srv *WebRTCSignalingServer) Call(req *webrtcpb.CallRequest, server webrtcp
 	var callerAuthMetadata map[string]string
 	if entity, ok := ContextAuthEntity(ctx); ok {
 		callerAuthEntity, callerAuthMetadata = entity.Entity, entity.AuthMetadata
+
+		// Auth metadata (app user ID, email) should be in the JWT token, but the
+		// internalForeverJWT used by the machine page has no metadata. Hackily pull it off
+		// `entity.Data`.
+		if callerAuthMetadata == nil {
+			if provider, ok := entity.Data.(interface{ CreateMetadata() map[string]string }); ok {
```
danielbotros: goutils is already filled with a bunch of auth types/paths (JWTs, fusionauth, api keys, secrets, etc...) and I honestly don't fully understand what all of them do or how they flow so I can't tell if adding another explicit interface that implements `CreateMetadata` for this cast to be less gross makes that better or not. I'll leave it up to you.

danielbotros (commented):
IIUC (I'm not confident) `ensureAuthed` populates `entity.AuthMetadata`. Why don't we add this fallback in `ensuredAuth` instead of in `Call` when `claims.Metadata` comes back empty?  This `Call` site goes back to a plain `entity.AuthMetadata` read and any future consumer of the metadata gets the fallback for free.

danielbotros (approved):
Also just approving a head of time since I'm going to be OOO and nothing is blocking

benjirewis:
> Why don't we add this fallback in ensuredAuth instead of in Call when claims.Metadata comes back empty?

Good call! Moved.

rpc/wrtc_signaling_server.go
benjirewis: I originally had it as a separate interface but felt creating another interface introduced yet another auth `type` that was really only useful for this single invocation. Will keep as-is if that SGTY.

PR #605, opened 2026-09-14
New per-rpc credential seed, handler, test

EvanDorsky:
This branch adds two new rpc dial options to support https://github.com/viamrobotics/rdk/pull/6384 - caching app-provided JWTs on the robot.

The two dial options are
- `WithInitialAccessToken` - takes a previously-obtained token that is used as the initial token in the existing per-RPC auth flow
- `WithAccessTokenHandler` - takes a handler that runs every time a new token is acquired in the per-RPC auth flow

In https://github.com/viamrobotics/rdk/pull/6384, `WithInitialAccessToken` is used to provide the cached JWT to the signaling server on an initial auth failure, and `WithAccessTokenHandler` is used to cache the signaling server JWT every time rdk gets a new JWT (for any reason) so that the cached token can update when needed.


dgottlieb (approved):
The `WithInitialAccessToken` is the API I would expect. Leaving the other half question of "is this the same code path that regular `Authenticate` takes" to Cheuk/Benji.

cheukt (approved):
lgtm - the Authenticate function is probably getting a bit complicated/hard to reason about, but don't have a clear idea of how to necessarily clean it up so happy to punt on that for now.

benjirewis (approved):
Seems reasonable to me.

PR #606, opened 2026-09-14
Support more pointer types in protoutils

jmatth:
We were seeing an issue where `protoutils.StructToStructPb` would error when converting a struct containing a field of type `*float64`. The underlying `strcutpb` library only deals with the subset of types necessary to represent valid JSON and rejects pointer types. For some reason we were already dereferencing several other pointer types in our code but not floats. I updated the code to 

1) handle floats and bools.
2) handle `nil` pointers to supported types. I suspect this is another bug that previously lacked testing.

~I also bumped our go version to `1.26` so I could use the new `new(<value>)` syntax in the test~


go.mod
```diff
@@ -1,6 +1,6 @@
 module go.viam.com/utils
 
-go 1.25.9
+go 1.26
```
npmenard: would that means rdk needs to be bump 1.26 for the fix to be deployed?

npmenard (approved):
LGTM, small nit not sure if its worth it to bump to 1.26 if it means rdk needs to be bumped too

go.mod
jmatth: Sadly yes. Reverted.

PR #607, opened 2026-09-15
APP-17935 - Increase onAnswerliveness interval to 2 seconds

cheukt:
Observations suggest that calling it too often is what's causing app alerts to fire


danielbotros (approved):
LGTM!

PR #608, opened 2026-09-15
protoutils: dereference *float64 and *bool in toInterface

JosephBorodach:
toInterface already dereferences pointer-to-string, pointer-to-uint, and pointer-to-int correctly, but pointer-to-float and pointer-to-bool fall into the fallthrough that returns the raw pointer

## The issue
Any struct carrying *float64 or *bool fields fails StructToStructPb with `proto: invalid type: *float64`

Screenshot 2026-09-15 at 1 15 39 PM

## How I noticed this bug
rdk added MaxVelocity *float64 / MaxAcceleration *float64 to JointConfig - https://github.com/viamrobotics/rdk/pull/6361

Every consumer of FrameSystemPart.ToProtobuf on a model with those set now errors out - frame system, motion service, module test cards

## The fix 
Mirrors the existing Uint/Int/String cases


jmatth:
Resolved by #606

PR #610, opened 2026-09-16
APP-17877: Resilient signaling protos

EvanDorsky:
Scope: https://docs.google.com/document/d/1YIJv0Ted5WqyxM33LRGjopmNEfpOL3NlCXnvY4hxMsI/edit?tab=t.0#heading=h.tcicyojyqi6c

This branch contains the protos for the resilient signaling project.

Today, when an SDK caller tries to dial a robot answerer via the cloud signaling server, the cloud signaling server authenticates both the caller and the answerer itself. The answerer will trust any caller that the cloud signaling server connects to it, without authenticating that caller itself.

This authentication flow requires the signaling server's encyrption key provider (Google KMS) to be available for signaling to work, because the signaling server must decrypt the API keys and robot secrets to authenticate the caller and answerer, respectively.

This project allows the robot to authenticate the caller itself in an event where the signaling server cannot perform its authentication function. The protos changes here:

1. Enable the signaling server to warn callers that they must authenticate against the answerer
2. Enable the signaling server to warn answerers that they must authenticate the caller
3. Enable answerers to inform the signaling server whether or not they can authenticate callers
4. Enable the caller to send its API key to the answerer for authentication

---


proto/rpc/webrtc/v1/signaling.proto
```diff
@@ -79,6 +79,9 @@ message CallRequest {
 // the initial response to starting a call.
 message CallResponseInitStage {
 	string sdp = 1;
+	// must_auth_against_answerer tells the caller that the signaling server could not
+	// authenticate it, so it must authenticate with the answerer over the PeerConnection.
+	bool must_auth_against_answerer = 2;
```
benjirewis, line 84: Perhaps a dumb question: the initial caller _request_ does _not_ need `can_auth_self` or something? In other words, clients do not need to let the signaling server know they are new enough to authenticate themselves against answerers?

EvanDorsky: Correct. Discussed offline, for but context here, the reasoning is that if callers do not know how to pass their API key to the answerer for the answerer to auth, then auth will simply fail. This is the same behavior whether or not the caller has some kind of `can_auth_self` for the signaling server to check.

benjirewis (approved):

PR #611, opened 2026-09-16
APP-9064: Authenticate unauthenticated callers

EvanDorsky:

This branch enables answerers to authenticate callers, and to inform the signaling server that they can authenticate callers.

This change should be done and merged before any other changes can be merged, because otherwise the signaling server could send unauthenticated callers to rdk versions that do not know they need to authenticate them, however there are other safeguards:
- cloud signaling will only send unauthenticated callers to servers that advertise their ability to authenticate callers (via the `can_auth_callers`) flag
- `can_auth_callers` can be manually disabled until the caller authentication feature is fully tested and trusted, even after deployment

---


rpc/wrtc_server_channel.go
```diff
@@ -24,8 +27,13 @@ type webrtcServerChannel struct {
 	// identity (or the coarse audience approximation when none was forwarded), since both
 	// inputs are fixed for the channel's lifetime.
 	entityInfo EntityInfo
-	server     *webrtcServer
-	streams    map[uint64]*webrtcServerStream
+	// mustAuthCaller is set when the signaler did not authenticate the caller.
+	mustAuthCaller bool
+	// callerAuthed is set once the caller presents a valid api_token. Only the data channel
+	// read loop (onChannelMessage) touches it, so it needs no lock.
+	callerAuthed bool
```
benjirewis, line 34: Can `onChannelMessage` fire async? I was a little surprised this wasn't `atomic.Bool`. You also do seem to be writing to it behind `ch.mu`.

rpc/wrtc_server_channel.go
```diff
@@ -96,6 +106,33 @@ func (ch *webrtcServerChannel) writeTrailers(stream *webrtcpb.Stream, trailers *
...
+// apiTokenVerifyTimeout bounds one api_token verification so a slow auth handler cannot hold the
+// data channel read loop.
+const apiTokenVerifyTimeout = 10 * time.Second
+
+// verifyAPIToken checks a caller's api_token (`<api-key-id>:<api-key>`) against the server's API
+// key auth handler and returns the caller's identity on success.
+func (ch *webrtcServerChannel) verifyAPIToken(apiToken string) (EntityInfo, bool) {
+	handler := ch.server.apiKeyAuthHandler
+	if handler == nil {
+		ch.webrtcBaseChannel.logger.Debug("no API key auth handler; cannot authenticate caller")
```
benjirewis, line 118: I wonder if these should be higher than `Debug` level and also print the API Key ID that's attempting to connect?

rpc/wrtc_server_channel.go
```diff
@@ -117,6 +154,31 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
...
+	if ch.mustAuthCaller && !ch.callerAuthed {
+		tok, isToken := req.GetType().(*webrtcpb.Request_ApiToken)
+		var callerEntity EntityInfo
+		verified := false
+		if isToken {
+			callerEntity, verified = ch.verifyAPIToken(tok.ApiToken)
+		}
+		if !verified {
+			// Fail the request on its own stream so the caller's RPC errors at once.
+			st := status.New(codes.Unauthenticated, "answerer requires an api_token from an unauthenticated caller")
```
benjirewis, line 167: Just a note that I like `codes.Unauthenticated` as opposed to `codes.PermissionDenied` to indicate authn vs. authz failure respectively.

rpc/wrtc_server_channel.go
```diff
@@ -117,6 +154,31 @@ func (ch *webrtcServerChannel) onChannelMessage(msg webrtc.DataChannelMessage) {
...
+			}
+			return
+		}
+		ch.mu.Lock()
+		ch.entityInfo = callerEntity
+		ch.callerAuthed = true
+		ch.mu.Unlock()
+		return
+	}
+	if _, isToken := req.GetType().(*webrtcpb.Request_ApiToken); isToken {
```
benjirewis, line 179: What situation is this? Signaling server told answerer "no need to auth this guy" and told caller "auth against the answerer please"?

rpc/wrtc_signaling_answerer.go
```diff
@@ -287,6 +287,10 @@ func (ans *webrtcSignalingAnswerer) startAnswerer() {
 			aa.offerSDP = initStage.Init.GetSdp()
 			aa.callerAuthEntity = initStage.Init.GetCallerAuthEntity()
 			aa.callerAuthMetadata = initStage.Init.GetCallerAuthMetadata()
+			aa.mustAuthCaller = initStage.Init.GetMustAuthCaller()
+			if aa.mustAuthCaller {
+				ans.logger.Debugw("caller is unauthenticated. robot must auth caller", "uuid", aa.uuid)
```
benjirewis, line 292: Also might raise above `Debug` and log the API Key ID of the unauthed client. I believe all these logs should be diagnostic anyway?

benjirewis (commented):
Nice! A few comments/questions.

rpc/wrtc_server_channel.go
EvanDorsky: Good call, yeah, this should be `atomic.Bool`.

rpc/wrtc_server_channel.go
EvanDorsky: The idea here is that after the answerer has authenticated the caller with the API key (the block right above this line), we want the answerer to throw out any other API keys it receives.

I think my logic was that after the caller sends an API token the first time and gets a success response, it shouldn't be sending more tokens, so this is kind of just a way to defend against API tokens making it further down the `onChannelMessage` response than they should go (and potentially getting logged by accident, for example).

That said, maybe it would be fine or even preferable to just re-auth against any new API token that the caller sends us, without caring about whether or not the caller is already authenticated.

benjirewis: Gotcha. I would say your current design is good (just drop any unexpected API key messages), but maybe a log here would be good to say "hey the client sent an out of order API key".

PR #612, opened 2026-09-16
APP-9061: Warn unauthenticated callers

EvanDorsky:

This branch enabled the signaling server to warn callers that the signaling server could not authenticate them, and that they must authenticate against the answerer directly by sending their API key.

In order to make this behavior safe, robots cache their signaling server-provided JWTs, allowing the signaling server to authenticate them even when it cannot directly check that their robot secrets are valid. That functionality is handled here: https://github.com/viamrobotics/rdk/pull/6384

---


rpc/server_auth.go
```diff
@@ -280,7 +280,7 @@ func (ss *simpleServer) tryAuth(ctx context.Context) (context.Context, error) {
 		if status, _ := status.FromError(err); status.Code() != codes.Unauthenticated {
 			return nil, err
 		}
-		return ctx, nil
+		return contextWithUnauthenticatedCaller(ctx), nil
```
benjirewis, line 283: Is this line also reached when the signaling server correctly and fully determines that the caller is unauthenticated to communicate with the answerer? Just want to make sure we're only asking the answerer to authenticate the caller when the signaling server is incapable of doing so (not when it deems the caller unauthenticated).

rpc/wrtc_signaling_server.go
```diff
@@ -194,16 +194,22 @@ func (srv *WebRTCSignalingServer) Call(req *webrtcpb.CallRequest, server webrtcp
...
 	var callerAuthEntity string
 	var callerAuthMetadata map[string]string
-	if entity, ok := ContextAuthEntity(ctx); ok {
+	entity, callerAuthed := ContextAuthEntity(ctx)
+	if callerAuthed {
 		callerAuthEntity, callerAuthMetadata = entity.Entity, entity.AuthMetadata
 	}
+	// APP-9061 if the auth interceptor let the caller through a public Call without
+	// authenticating it, continue and warn the answerer that it must authenticate the caller.
+	// A server that does no authentication at all never sets this.
```
benjirewis, line 209: > never sets this

This is saying that a signaling server setup without authentication _always_ reports `mustAuthCaller == true`? Or the other way around? Is the internal signaling server setup without authentication?

PR #613, opened 2026-09-16
APP-9065: Allow caller authentication errors to pass through

EvanDorsky:

For the degraded signaling path to work, authentication errors must not be fatal.

---


rpc/wrtc_client.go
```diff
@@ -196,12 +201,27 @@ func dialWebRTC(
...
-		if s, ok := status.FromError(err); ok && (s.Code() == codes.Unimplemented ||
-			(s.Code() == codes.InvalidArgument && s.Message() == hostNotAllowedMsg)) {
+		s, ok := status.FromError(err)
+		switch {
+		case ok && (s.Code() == codes.Unimplemented ||
+			(s.Code() == codes.InvalidArgument && s.Message() == hostNotAllowedMsg)):
+			// this would be where we would hit an unimplemented signaler error first.
 			return nil, nil, ErrNoWebRTCSignaler
+		case ok && dOpts.webrtcOpts.AllowUnauthenticatedSignaling &&
+			(s.Code() == codes.PermissionDenied || s.Code() == codes.Unauthenticated):
```
benjirewis: In the event that the signaling server is incapable of authenticating the caller, shouldn't the signaling server respond with "hey you can continue connecting but you have to authenticate against the answerer yourself" as opposed to `codes.PermissionDenied` or `codes.Unauthenticated`?

In other words, I had only expected a need for https://github.com/viamrobotics/goutils/pull/614 and not for this PR.

EvanDorsky: So you're raising a good point, which is that we could modify the signaling server to have it return more helpful error messages when it can't authenticate the caller but it knows that the degraded path is available.

I think the way we're trying to handle this (telling the caller it needs to use the degraded path) is by passing the `must_auth_against_answerer` flag back in the `CallResponseInitStage` (that's implemented in https://github.com/viamrobotics/goutils/pull/612) rather than giving a better authentication failure message.

Regardless though, my understanding is that we need changes in this part of caller code so that the caller can continue after authentication fails - since authentication is a separate call to the signaling server that the caller does before actually trying to call.

I think what's going on here is that when line 202 runs:
```
configResp, err := signalingClient.OptionalWebRTCConfig(signalCtx, &webrtcpb.OptionalWebRTCConfigRequest{})
```

under the hood the caller is authenticating against the signaling server. And immediately after we look for authentication errors.

Before now, an authentication error here would be fatal and the caller would give up on trying to dial. So that's why we need to change this code.

But I might be misunderstanding this part or might not be accounting for something properly so please let me know if this makes sense.

benjirewis: Gotcha; thanks for more info here. Let's sync when you get back.

PR #618, opened 2026-09-19
rpc: cut mDNS traffic from ICE candidate resolution and duplicate zeroconf responders

erh:
## Problem

Robots were reported sending ~10k mDNS packets/minute. viam-server is silent at idle; the volume comes from three amplifiers in this package.

1. **ICE re-queried browser `.local` candidates forever.** `SetICEMulticastDNSMode(QueryOnly)` makes pion/ice resolve every remote `<uuid>.local` host candidate (browsers always send these) via `mDNSConn.Query(cand.context(), ...)`. For a remote candidate that context never completes, so an unresolvable name (a browser on another network) is queried once a second, on every interface, for the life of the peer connection. Measured: one candidate on three interfaces produced ~190 packets/min, IPv4 alone.
2. **Eight independent zeroconf responders per cloud robot.** Each instance name was registered as name + dashed name, each as both a `RegisterDynamic` and a `RegisterProxy` (the second added in #422). Every `zeroconf.Server` answers every matching query on its own, per question, so one instance lookup or one `_services._dns-sd._udp.local` enumeration from any device on the LAN drew 4 responses per name variant per IP family.
3. **Dial-side lookups queried over IPv6** even though only IPv4 records are accepted.

## Changes

- ICE mode is now `MulticastDNSModeDisabled`. New `rpc/wrtc_mdns.go` resolves `.local` candidates itself with a bounded timeout (5s, shorter if the dial/answer context ends) using a short-lived pion/mdns querier, then adds the rewritten candidate; unresolvable ones are dropped. Applied to trickle candidates on both client and answerer, and to candidates embedded in remote SDPs (stripped before `SetRemoteDescription`, then resolved asynchronously).
- Each name variant gets one `RegisterProxy` registration advertising the name as hostname (so `<name>.local` still resolves, RSDK-10402) with no static address list, which makes zeroconf answer with the receiving interface's addresses as `RegisterDynamic` did. Names whose dashed form equals another registered name are not registered twice.
- Dial resolver uses `SelectIPTraffic(IPv4)`.

Behavior change to note: viam-server no longer answers mDNS A queries for the OS hostname (that came from the `RegisterDynamic` SRV target). avahi/mDNSResponder own that name on hosts that have them; a machine relying on viam-server to resolve `<os-hostname>.local` would need the name in its instance names.

## Testing

- New `TestAddRemoteICECandidate` (resolvable name via in-test responder, unresolvable name dropped within the timeout, context cancellation), `TestStripMDNSCandidatesFromSDP`, `TestMDNSCandidateAddress`, `TestServerMulticastDNSRegistrations`.
- Existing mDNS dial/server tests pass; full `go test ./rpc` passes; golangci-lint 2.6.2 with `etc/.golangci.yaml` reports 0 issues.


cheukt (commented):
Was able to repro the issue, confirm that the fix helps, and that nothing seems to have regressed on linux and darwin. On windows, there is a regression where <name>.local stop returning address records. will fix and merge when ready

cheukt (approved):
check that windows fix did allow for mdns lookups on same lan

shannonbradshaw:
@jmatth and @dgottlieb would you please prioritize review here?

cheukt:
@shannonbradshaw this pr has been merged and pulled into rdk main, release later today

